PR #13, opened 2026-05-16
update todo

perminder-klair:
we made it look nicer


perminder-klair:
also add about learning working with claude agents

PR #37, opened 2026-05-18
Pause DJ LLM calls when no one is listening

perminder-klair:
Add an opt-in mode that stops all autonomous DJ LLM work when Icecast
reports zero listeners, letting the stream coast on the auto playlist,
and resumes it as soon as someone tunes in.

- New broadcast/listeners.js polls Icecast status-json.xsl every 15s and
  caches the live listener count; fail-open if Icecast is unreachable.
- llm.pauseWhenEmpty setting (off by default) with an admin UI toggle.
- Gates the autonomous LLM triggers: scheduler segments, the between-
  track picker/link, and listener requests. Operator overrides stay live.
- Feeds the listener count into the DJ prompt context so dialogue can be
  aware of the audience size.

https://claude.ai/code/session_01P3yHBxfYwc4cZ22Dw9QMYY

### DIFF
--- a/controller/src/broadcast/listeners.js
+++ b/controller/src/broadcast/listeners.js
@@ -1,0 +1,55 @@
+// Icecast listener-count monitor.
+//
+// Polls Icecast's status-json.xsl on an interval and caches the live listener
+// count for the /stream.mp3 mount, so the DJ gates can ask "is anyone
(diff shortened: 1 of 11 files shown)


perminder-klair:
hold on this one, needs testing, I will check later

PR #63, opened 2026-05-22
feat(cli): optional managed Navidrome in the setup wizard

perminder-klair:
## What

Adds a **managed Navidrome** path to the `subwave setup` wizard. Today SUB/WAVE assumes the operator already runs Navidrome somewhere — neither compose file defines it, and `collectNavidrome()` only collects a URL + credentials. A first-time user with no music server has nothing to point the wizard at.

Now the wizard branches: bring your own Navidrome (unchanged), or let SUB/WAVE provision one.

## How

**Compose files** — new `navidrome` service in both compose files, guarded by `profiles: ["navidrome"]`, so the default stack stays lean. Dev publishes `:4533` for the Navidrome UI; prod keeps it internal (preserves the "only Caddy binds a port" invariant). Uses `${NAVIDROME_PASS:-changeme}` rather than `:?` because compose interpolates the whole file regardless of which profiles are active — a required-var would break operators who run their own Navidrome.

**Setup wizard** (`cli/src/commands/setup.ts`) — step 4 now asks where music comes from:
- *I already run Navidrome* → today's URL/credentials + live probe flow, byte-identical.
- *Run one for me* → prompts for a music folder (default `<STATE_DIR>/music`), creates the music + data dirs, auto-generates an admin password, wires `NAVIDROME_URL=http://navidrome:4533` / `user=admin`.

**Activation** is entirely through `docker/.env`: managed mode writes `COMPOSE_PROFILES=navidrome` + music/data paths + `NAVIDROME_PASS`; external mode writes an empty `COMPOSE_PROFILES` so a managed→external re-run cleanly disables the profile. Because Compose reads `COMPOSE_PROFILES` from `docker/.env`, every lifecycle command (`start`/`stop`/`restart`/`ps`) picks up the profile with no per-command flags and no changes to those commands. The wizard is re-run-idempotent — managed branch pre-selected, admin password retained.

`doctor.ts` needs no change — it iterates compose services dynamically, so managed Navidrome surfaces automatically.

## Verification

- CLI typechecks clean (`tsc --noEmit`).
- Both compose files validate via `docker compose config`; `navidrome` appears only when the profile is active.
- Confirmed empirically that `docker compose` reads `COMPOSE_PROFILES` from `docker/.env` — the load-bearing assumption behind zero lifecycle-command changes.
- **Not run:** a full interactive wizard pass (clack prompts block on stdin).

## Follow-up

Confirm `ND_DEVAUTOCREATEADMINPASSWORD` is still the current Navidrome auto-admin env var against their docs before relying on it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,9 +130,12 @@ npm run dev:web        # web UI on :7700 — a separate, hot-reloading process
 ```
 
 In dev the wizard runs Icecast + Liquidsoap + Controller in Docker; only the
-Next.js web UI runs as a separate host process. You'll need a reachable
-**Navidrome** instance and an **LLM provider** — the homelab default is a local
-**Ollama** box (no API key needed); the wizard collects and probes both.
+Next.js web UI runs as a separate host process. You'll need an **LLM
+provider** — the homelab default is a local **Ollama** box (no API key
+needed) — and a music source. For music the wizard either points SUB/WAVE at
+an existing **Navidrome** instance, or, if you don't run one, provisions a
+managed Navidrome container for you (just give it the folder your music lives
+in); both paths are collected and probed by the wizard.
 
 Running the steps by hand instead — see *Common commands* in
 [`CLAUDE.md`](CLAUDE.md).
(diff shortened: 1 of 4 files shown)


perminder-klair:
FYI also check https://github.com/perminder-klair/subwave/tree/claude/navidrome-subsonic-comparison-fQZRW

PR #115, opened 2026-05-25
fix(controller/picker): stop deterministic track/artist clustering

perminder-klair:
Diagnosed via 24h event log: top tracks aired 6-8x ("2 AM" by Karan Aujla x8, "Welcome To Heartbreak" x4, "Eyes on Me" x7) and the picker's tool calls were highly repetitive (`tracksByMood("night")` 46x; `topSongsByArtist("Karan Aujla")` 7x). Two compounding causes:

1. Navidrome / `library.songsByMood` return results in deterministic order, so every repeat call returned the same first 8 candidates.
2. The picker's repeat-block window was count-based (last 25 plays, ~90 min - 2h) and lived only in `queue.history` (capped at 50, in-memory only) — so any track aired more than ~2h apart was free to re-pick.

Fix:

- `llm/tools.ts` collect() shuffles each tool's results before slicing to cap=8, breaking the deterministic Navidrome ordering. Dominant fix.
- New rolling 24h sidecar at `state/recent-plays.json` (cap 300), written debounced from `onTrackStarted`. `queue.recentlyPlayed(hours)` returns `{ids, keys}`; new `queue.recentArtistsSince(hours)` returns a Set. Both walk the sidecar newest-first to the cutoff.
- On boot, `recover()` backfills the sidecar from today's + yesterday's `events-*.jsonl` `track.play` events so a controller restart never resets the block window to zero (observed cause of a "2 AM" repeat 14 min after restart).
- Picker callsites switched to hour-based windows: auto-DJ uses 12h tracks + 2h artists; `runRequest` and `routes/request.ts` use 2h tracks (requests stay near-unfiltered so listeners can re-request a song from earlier in the day); scheduler and pool fallback match the agent's 12h.
- `music/picker.pickViaPool` (the fallback used when the LLM agent fails its done-tool ~1 in 4-5 calls) now also reads `recentArtistsSince(2)` and filters candidates by it — without this, a Prabh Deep cluster of 4 tracks in 1h 24min slipped through.
- `PICKER_CRITERIA` adds a no-title-as-clock-cue rule (caught the agent verbatim picking "2 AM" with rationale "Punjabi night thread, 2AM timing").
- `MAX_PER_ARTIST` 3 -> 2 in tools.ts.
- `session.buildHandoff` no longer prepends "recently aired ..." track list to the next session prompt — that was leaking favourites across session rolls.

Verified live: 5h post-deploy, 73 plays, 0 track repeats, 68 unique artist strings (vs 481 plays / "2 AM" x8 / heavy Kanye/Diljit concentration in the prior 24h).

### DIFF
--- a/controller/src/broadcast/dj-agent.ts
+++ b/controller/src/broadcast/dj-agent.ts
@@ -72,8 +72,8 @@ export const pickerAgent = defineAgent({
   maxSteps: 4,
   timeoutMs: 22000,
   buildSystem: () => pickSystem(),
-  buildTools: ({ recentIds, recentArtists }) => {
-    const { tools, seen } = buildPickerTools({ recentIds, recentArtists });
+  buildTools: ({ recentIds, recentKeys, recentArtists }) => {
+    const { tools, seen } = buildPickerTools({ recentIds, recentKeys, recentArtists });
     return { tools, extras: { seen } };
   },
 });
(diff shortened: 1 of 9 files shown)


perminder-klair:
@WhatKateDoes repeating songs issue fixed, hopefully results will be better now

WhatKateDoes:
Updated, testing! :)

PR #286, opened 2026-06-03
ci: stop baking tts-heavy models at build time

perminder-klair:
The subwave-tts-heavy image build pre-warmed Chatterbox + PocketTTS weights via from_pretrained()/load_model() during the Docker build, pulling several GB from the Hugging Face Hub inside CI over an unauthenticated, rate-limited connection. That made the build slow (20+ min) and flaky (HF 429s/timeouts, the 'set a HF_TOKEN' warning, retried runs).

Drop the build-time model download and keep an import-only smoke test instead. Weights now download lazily into the HF cache (HF_HOME, set per-engine by server.py) on the first /speak after container start. ARG HF_TOKEN stays declared so the compose build.args pass-through doesn't warn.

### DIFF
--- a/docker-compose.byo.yml
+++ b/docker-compose.byo.yml
@@ -150,3 +150,13 @@ services:
       - HF_TOKEN=${HF_TOKEN:-}
     volumes:
       - *state-mount
+      # Persist the per-engine Hugging Face caches across container recreates.
+      # Weights download at boot (the workers load the model before reporting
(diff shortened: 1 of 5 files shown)


perminder-klair:
Superseded by #294, which targets `develop` (per the repo's develop-only convention) and adds the HF model-cache persistence on top of the same CI fix.

PR #299, opened 2026-06-04
fix(controller): relax picker recency for small libraries

### LINKED ISSUE #285: Make the picker's recency-exclusion window configurable (hardcoded 12h/2h starves small-artist libraries)
### What I expected vs. what happened

With `pickerAgent` enabled, I expected the DJ agent to pick a real track from my library. On a small-*artist* library (~10 artists, 210 tracks) running continuously, almost every pick instead fails with `agent returned unknown id … — falling back to pool`. The agent is handed `candidates=0`, so it returns a fabricated id, which then gets rejected.

### Why it happens

`collect()` in `controller/src/llm/tools.ts` excludes every candidate whose artist was heard in the last 2h (`recentArtistsSince(2)`) and every track played in the last 12h (`recentlyPlayed(12)`). Both are called with hardcoded literals in `controller/src/broadcast/dj-agent.ts` (around lines 202 and 205). With only a handful of artists in continuous rotation, every artist is always "recent," so `collect()` returns empty for *every* discovery tool, including the no-argument `randomSongs` and `tracksByMood`. The agent then has nothing real to choose from. (`queue.ts` already notes that this window "narrows the pool fast on a small library.")

The core gap: these are sensible defaults for a large, artist-diverse library, but there is no supported way to tune them for a smaller catalog. They are not exposed in settings or env, and `pickerAgent` is the only picker-related setting.

### Steps to reproduce

1. Library with ~10 distinct artists, `pickerAgent` enabled, any non-Ollama provider.
2. Let it run for more than ~2h of continuous playback.
3. Watch `state/logs/events-*.jsonl`: `pick.rejected` events show `candidates=0`, and discovery tool calls (including `randomSongs`) come back empty.

### Relevant logs

```
pick.rejected candidates=0 steps=2 tools=['tracksByMood'] resultsizes=[0]
pick.rejected candidates=0 steps=4 tools=['similarSongs','tracksByEnergy','randomSongs'] resultsizes=[0,0,0]
[error] DJ agent pick failed: agent returned unknown id 60sY61q2l7x1Xb5m3F4QJb — falling back to pool
```

### What helped, locally

As an experiment I lowered the windows (`recentlyPlayed(12)` to `1`, `recentArtistsSince(2)` to `0.25`) and pick success went from 0/5 to 3/5 over consecutive cycles; growing the library improves it further. So the picker logic itself looks fine, it just needs room on a smaller catalog.

### A suggestion (offered, not prescriptive)

The underlying need is just a supported way to tune that exclusion window for different library sizes. A couple of directions, whichever fits your design best:

- Expose the two windows as settings and/or env (e.g. alongside the existing picker settings), with the current values as defaults so nothing changes for existing users.
- Or auto-scale the windows to the library's distinct-artist count, so they relax on small catalogs without the operator needing to know the knob exists.

The second degrades gracefully with zero configuration; the first is simpler and more explicit. Happy to send a PR for whichever direction you'd prefer.

Thanks for SUB/WAVE, it has been a genuinely fun project to run.

uhhuhyeah:
## Summary

- scale the picker recency windows from the tagged library's distinct-artist count
- add a shared picker candidate filter that relaxes recency exclusions instead of returning an empty candidate set
- apply the same recency windows and empty-pool guard to both the conversational picker agent and the stateless pool fallback
- expose `distinctArtists` in library stats so the picker can make the scaling decision without an operator-facing knob

## Why

Fixes #285.

Small-artist libraries can cycle through every artist inside the fixed 2h artist window. When that happens, every discovery tool can return no visible candidates after filtering, including `randomSongs`. The picker agent then has no real ids to choose from and can fabricate an id, which gets rejected and falls through to a weak fallback.

The new helper keeps the existing 12h track / 2h artist windows for diverse libraries, but scales them down for smaller artist counts. For example, a 10-artist library gets shorter windows, while libraries with 48+ artists keep the old values.

## Details

The candidate filter now tries strict filtering first, then relaxes recent artists, then relaxes recent tracks/keys only if the stricter passes produce nothing. This preserves the existing anti-repeat behavior when candidates are available, while guaranteeing the picker is not handed an empty pool unless the source list itself is genuinely empty.

`music/picker.ts` also samples each source with a recent-track fallback before the final shared filter runs. That is intentionally broader than just changing the final filter: without keeping some source candidates around, the final guard has nothing to relax when a small library has all sampled tracks inside the recency window.

## Validation

- `./node_modules/.bin/tsx scripts/picker-recency-regression.ts`
- `npm --prefix controller run lint`

`npm --prefix controller run lint` exits 0. The repo still reports existing `@typescript-eslint/no-explicit-any` warnings.


perminder-klair:
Thanks for this, really clean refactor, and nice that it fixes the root cause for *both* the agent picker and the pool fallback rather than just one. The auto-scaling approach + empty-pool guard is exactly what we were after on #285, and the ≥48-artist path preserving the old 12h/2h defaults keeps existing setups unchanged.

One process note: could you retarget this to **`develop`** rather than `main`? We run feature → `develop` → `main` (release-please cuts the releases on `main`), so changes need to land on `develop` first. The files you touched are identical on both branches, so switching the base is a clean no-conflict diff, no rebase needed. (I've gone ahead and flipped the base to `develop` for you.)

Two tiny things, non-blocking: `distinctArtists` counts only tagged tracks (`SQL_HAS_MOODS`), so on an untagged library the scaling falls back to the 12h/2h defaults, probably fine, just worth a comment. And the `1/12` scale floor + the `MIN_*` constants end up clamping to the same values, so one of them is redundant.

PR #332, opened 2026-06-09
feat(picker): add OpenSubsonic sonicSimilarity as an optional track source

perminder-klair:
## What

Navidrome **v0.62** ships the OpenSubsonic **`sonicSimilarity`** extension (`getSonicSimilarTracks`) — *audio-based* neighbours computed from the actual audio via Navidrome's plugin system. This wires it into the pool picker as an additive track-selection source, complementing the two similarity signals already in place:

1. `getSimilarSongs2` — Last.fm collaborative-filtering graph
2. `library.tracksLikeThis()` — the controller's own embedding-KNN
3. **`getSonicSimilarTracks` — acoustically-grounded, new in this PR**

## How

- **`subsonic.ts`**
  - `supportsSonicSimilarity()` — probes `getOpenSubsonicExtensions`, caches the result 30 min, never throws (old Navidrome / network failure → `false`). The TTL means a just-upgraded server is picked up without a controller restart.
  - `getSonicSimilarTracks(id, {count})` — calls the endpoint, archive-filtered like every other source.
  - The coverage-logging extractor learns the `sonicMatch` wrapper shape (defensive — handles top-level vs nested, and `entry`/`song`/inlined Child), so `/debug` analytics still count these as real song results.
- **`picker.ts`** — new source **1c. `sonic-similar`**, slotted after the embedding-KNN source, capped at 4 and capability-gated. It only spends a call when the extension is present; otherwise it's a silent no-op, so an un-upgraded Navidrome behaves exactly as before.
- **`README.md`** — recommends Navidrome ≥0.62 (security hardening + this extension), noting the new transcode-concurrency limits never throttle the radio because SUB/WAVE streams with `format=raw`.

## Why

Purely additive: a third, acoustically-grounded similarity signal — especially valuable for regional / non-Western catalogues where Last.fm coverage is thin — with **zero behaviour change** when the operator hasn't enabled the plugin.

## Verification

- `tsc --noEmit` → exit 0
- `eslint .` → 0 errors (only the codebase's pre-existing `any` warnings)
- Commit scoped to the three intended files (reverted incidental `npm install` artifacts)

> Note on response shape: the OpenSubsonic docs are slightly ambiguous on whether `sonicMatch` sits at the top level or under a `sonicSimilarTracks` wrapper, so extraction handles both. On a real 0.62 box with the plugin enabled, the quickest check is watching `/debug` / the booth log for a `sonic-similar=N` entry in the pool summary.

https://claude.ai/code/session_01V7Qa6nAW2unAzzfmjPRS4d

---
_Generated by [Claude Code](https://claude.ai/code/session_01V7Qa6nAW2unAzzfmjPRS4d)_


diablo581:
OK I got audiomuse-ai setup and running (sloooooowly) and it plumbed into navidrome, but I don't see any settings for it in the subwave UI?

perminder-klair:
> OK I got audiomuse-ai setup and running (sloooooowly) and it plumbed into navidrome, but I don't see any settings for it in the subwave UI?

audiomuse ai has no integration with subwave or other way around. what exactly you looking for?

diablo581:
Sorry to be confused and confusing.  I was looking at this release and attempting to get the similarity stuff working, and I think I misunderstood that it needed to talk directly to audiomuse ai, but rereading it looks like it is polling navidrome for the info, which navidrome will get the info from audiomuse ai.

So I guess getting the info from navidrome is just part of the sync of the library?

PR #371, opened 2026-06-13
feat(analysis): richer acoustic analysis (Music Understanding blueprint, phases 1–6)

perminder-klair:
## What

Implements all six features from `docs/music-understanding-plan.md` — richer, time-ranged acoustic analysis modelled on Apple's WWDC26 Music Understanding framework. Each feature is computed in the existing heavy-DSP analyze path, stored as a nullable column, and **degrades cleanly to today's behaviour when NULL** (un-analysed library, or a backend without the relevant dependency).

Built as one commit per phase so it can be reviewed incrementally. **Work in progress — to be continued.** Two consumer payoffs are intentionally deferred (see below).

## Phases

| Phase | Feature | Highlights |
|---|---|---|
| 1 | **LUFS gain** | `pyloudnorm` integrated loudness → per-track `liq_amplify` gain toward −14 LUFS (±6 dB clamp), via Liquidsoap's ReplayGain mechanism. Retires the deliberately-removed bus normaliser by measuring, not reshaping. |
| 2 | **Song structure** | librosa agglomerative sections; intro-aware crossfade cap so the incoming fade-in finishes before the vocals. |
| 3 | **Vocal ranges** | Demucs vocal-presence ranges (opt-in: `WITH_DEMUCS` build + `ANALYZE_VOCAL_ACTIVITY` / `settings.audio.vocalActivity`); vocal-onset intro detection; dedicated backfill scope. |
| 4 | **Pace curve** | onset-strength perceptual energy curve (decoupled from BPM); scalar mean pace surfaced to the LLM picker. |
| 5 | **Beat/bar grid** | per-beat + 4/4 downbeat timestamps (previously discarded); bar-aligned crossfade snap. |
| 6 | **Structured key** | per-region tonic + mode over time; scalar `musical_key` kept for back-compat. |

## Design invariants held throughout

- **Uniform write path** every phase: worker emits (omit-on-failure) → `analyzer.ts` coerces both backends → `analyze.ts` forwards → `library-db.ts` migration + `upsertTrackAnalysis` + `rowToTrack` → consumer.
- **NULL-degrades everywhere.** No new feature can turn the analysis pass into a hard failure; the broadcast bus stays the brick-wall limiter only — masters untouched.
- **Schema discipline.** Migrations v4→v9; `ANALYSIS_VERSION` 1→6 (bumped for main-pass outputs). Vocal activity is **not** a version bump — it's opt-in via a backfill scope + a `COALESCE` write so a normal bpm/key pass never wipes existing Demucs data.
- **Heavy deps gated** like CLAP: `pyloudnorm` is base (cheap); Demucs behind a build arg + runtime flag.

## Deferred (good follow-ups, noted in commits)

- Content-aware ducking talk-window scheduling (Phase 3 payoff a).
- Per-section harmonic mixing in the picker (Phase 6 payoff).
- Boundary-aligned **outro** crossfade — the 60s analysis window only reliably yields intro-side structure; the outro side needs a fuller decode.

## Verification

- `tsc --noEmit` clean, eslint 0 errors, both Python workers `py_compile` clean.
- Pure helpers spot-checked via `tsx`: `gainForLoudness` clamp, `crossSecondsFor` intro-cap + bar-snap, vocal null/`[]` semantics.
- **Not yet run on real audio / Liquidsoap** (containerized deps): the worker against a library and a `liquidsoap --check`, plus the live even-loudness / boundary-crossfade checks per the plan's verification section.

### DIFF
--- a/controller/scripts/analyze_worker.py
+++ b/controller/scripts/analyze_worker.py
@@ -50,6 +50,17 @@ EMBED_ENABLED = os.environ.get("ANALYZE_AUDIO_EMBEDDING", "").strip().lower() in
 CLAP_SR = 48000
 CLAP_EMBED_DIM = 512
 
+# --- Vocal-activity ranges (optional, opt-in) ------------------------------
+# Off unless ANALYZE_VOCAL_ACTIVITY is truthy. Runs Demucs source separation to
+# isolate the vocal stem, then thresholds its energy envelope into present/
+# absent ranges. Heavy (a real torch model) — gated like CLAP. Demucs wants
+# 44.1 kHz stereo.
+VOCAL_ENABLED = os.environ.get("ANALYZE_VOCAL_ACTIVITY", "").strip().lower() in (
+    "1", "true", "yes",
+)
+DEMUCS_SR = 44100
+DEMUCS_MODEL = os.environ.get("DEMUCS_MODEL", "htdemucs")
+
 # Krumhansl-Kessler key profiles (major/minor), indexed from the tonic.
 MAJOR_PROFILE = [6.35, 2.23, 3.48, 2.33, 4.38, 4.09, 2.52, 5.19, 2.39, 3.66, 2.29, 2.88]
 MINOR_PROFILE = [6.33, 2.68, 3.52, 5.38, 2.60, 3.53, 2.54, 4.75, 3.98, 2.69, 3.34, 3.17]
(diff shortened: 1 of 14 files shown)


perminder-klair:
Superseded by #372 (same commits, cleaner branch name `feat/music-understanding-analysis`).

PR #372, opened 2026-06-13
feat(analysis): richer acoustic analysis (phases 1–6)

perminder-klair:
## What

Implements all six features from `docs/music-understanding-plan.md` — richer, time-ranged acoustic analysis (key, pace, structure, loudness, beat grid, and instrument activity that vary *across* a song, stored alongside today's scalars). Each feature is computed in the existing heavy-DSP analyze path, stored as a nullable column, and **degrades cleanly to today's behaviour when NULL** (un-analysed library, or a backend without the relevant dependency).

Built as one commit per phase so it can be reviewed incrementally. **Work in progress — to be continued.** Two consumer payoffs are intentionally deferred (see below).

## Phases

| Phase | Feature | Highlights |
|---|---|---|
| 1 | **LUFS gain** | `pyloudnorm` integrated loudness → per-track `liq_amplify` gain toward −14 LUFS (±6 dB clamp), via Liquidsoap's ReplayGain mechanism. Retires the deliberately-removed bus normaliser by measuring, not reshaping. |
| 2 | **Song structure** | librosa agglomerative sections; intro-aware crossfade cap so the incoming fade-in finishes before the vocals. |
| 3 | **Vocal ranges** | Demucs vocal-presence ranges (opt-in: `WITH_DEMUCS` build + `ANALYZE_VOCAL_ACTIVITY` / `settings.audio.vocalActivity`); vocal-onset intro detection; dedicated backfill scope. |
| 4 | **Pace curve** | onset-strength perceptual energy curve (decoupled from BPM); scalar mean pace surfaced to the LLM picker. |
| 5 | **Beat/bar grid** | per-beat + 4/4 downbeat timestamps (previously discarded); bar-aligned crossfade snap. |
| 6 | **Structured key** | per-region tonic + mode over time; scalar `musical_key` kept for back-compat. |

## Design invariants held throughout

- **Uniform write path** every phase: worker emits (omit-on-failure) → `analyzer.ts` coerces both backends → `analyze.ts` forwards → `library-db.ts` migration + `upsertTrackAnalysis` + `rowToTrack` → consumer.
- **NULL-degrades everywhere.** No new feature can turn the analysis pass into a hard failure; the broadcast bus stays the brick-wall limiter only — masters untouched.
- **Schema discipline.** Migrations v4→v9; `ANALYSIS_VERSION` 1→6 (bumped for main-pass outputs). Vocal activity is **not** a version bump — it's opt-in via a backfill scope + a `COALESCE` write so a normal bpm/key pass never wipes existing Demucs data.
- **Heavy deps gated** like CLAP: `pyloudnorm` is base (cheap); Demucs behind a build arg + runtime flag.

## Update — picker wiring + robustness

Follow-on commits on top of the phase work:

- **Pace + structure surfaced to the DJ picker.** `pace` and `sections` now reach `djAgentPick` (and the pool picker), explained in the shared `PICKER_CRITERIA`, so the DJ reasons about energy arcs / arrangement, not just tempo/key. Previously the analysis data was computed but never reached the picker LLM.
- **Missing-file library entries handled.** Navidrome serves a Subsonic JSON error (HTTP 200) for tracks whose file is gone from disk; `downloadCapped` now detects the non-audio response and fails fast with the real reason (`no such file or directory`) instead of writing JSON-as-audio and failing opaquely as "analyze failed". Also fixes a one-ahead-prefetch unhandled-rejection that crashed the whole pass once those entries started rejecting frequently.

## Deferred (good follow-ups, noted in commits)

- Content-aware ducking talk-window scheduling (Phase 3 payoff a).
- Per-section harmonic mixing in the picker (Phase 6 payoff).
- Boundary-aligned **outro** crossfade — the 60s analysis window only reliably yields intro-side structure; the outro side needs a fuller decode.

## Verification

- `tsc --noEmit` clean, eslint 0 errors, both Python workers `py_compile` clean.
- Pure helpers spot-checked via `tsx`: `gainForLoudness` clamp, `crossSecondsFor` intro-cap + bar-snap, vocal null/`[]` semantics.
- Picker wiring + prefetch hardening verified in-container against the running stack (real v6 track surfaces `pace`/`sections`; a known missing-file entry now throws the typed `NonAudioResponseError`).


perminder-klair:
## Analysis — live-tested on the production stack

Ran this branch on the prod single-host stack and kicked a full re-analysis of the library (~7.1k tracks). Findings below; the picker-wiring + robustness commits in this PR came out of it.

### What's live on-air (verified)
- **LUFS gain normalisation** — working. Per-track `liq_amplify` toward −14 LUFS; mixed-loudness library plays at even volume, master bus still just the limiter.
- **Adaptive crossfade** — working, from `bpm`/`musical_key`/`intro_ms` (tempo/key compatibility + bar-snap + intro cap).
- **CLAP audio-similarity** — working (`tracksThatSoundLikeThis`, audio-similar pool source).
- **Pace + structure now reach the picker** — before the follow-on commit, the analysis data was computed but `djAgentPick` only saw `bpm`/`key`/`intro`. `pace` (mean) and `sections` (count) now surface to both the agent and pool pickers, explained in the shared `PICKER_CRITERIA`.

### Computed but not yet consumed (suggested follow-ups)
A few v6 columns are populated by the pass but have no reader yet — flagging so the DSP cost is tracked against a consumer. All NULL-degrade correctly, so this is *measurement ahead of consumption*, not a correctness issue:
- **`beats_json` / `bars_json`** — the bar-aligned crossfade derives bar length from the scalar `bpm`, not this grid; the per-beat / downbeat grid itself is unread. True downbeat-phase beatmatching is the Phase 5 payoff.
- **`key_ranges_json`** — the picker/crossfade use the whole-track `musical_key`; per-region key is unread (Phase 6 harmonic-mixing payoff).
- **`peak_db`** — stored, no consumer.
- **`vocal_ranges_json`** — NULL across the library unless tts-heavy is built `WITH_DEMUCS=1` (Phase 3, deliberately opt-in).

### Robustness finding (addressed here)
Re-analysing surfaced a pre-existing library-integrity issue: a meaningful chunk of the Navidrome library (~8–13% in this run, clustered by missing album) are **stale DB entries whose audio files are gone from disk**. Navidrome serves these as an HTTP-200 Subsonic JSON error, not audio. Two problems it exposed, both fixed in this PR:
1. `downloadCapped` only checked `res.ok` + non-empty, so it wrote the error envelope to disk as `.audio` → the decoder failed opaquely (`analyze failed`), masking the real `no such file or directory`. → now detects the non-audio response (content-type + a tiny-file `{`/`<` magic backstop) and fails fast with a typed `NonAudioResponseError` carrying the real reason.
2. Once those entries started rejecting frequently, the **one-ahead prefetch's floating rejection crashed the whole pass** under Node's default `--unhandled-rejections=throw` (exit 1 at track ~100). → prefetch reduced to a settled result so a rejection can never float.

> Note: these stale tracks are also **unplayable on-air** (the picker can select one and it fails to stream). The data-side cure is a Navidrome rescan to purge missing-file entries — out of scope for this PR.

### Coverage
Re-analysis pass **~48% complete and climbing**, running cleanly well past the earlier crash point; stream stays on-air throughout. Each re-analysed track goes live incrementally (loudness + structure-aware) with no restart.

Issue #405, opened 2026-06-16
Embedding options not shown


diablo581:
So if you are using locca (llama.cpp) as your LLM and you try to run a tagging session it gives you the following error:

`[tag] embedding preflight failed (not_embedding_model):
The embedding endpoint is reachable but "text-embedding-3-small" can't produce embeddings —
it's a chat/generative model, not an embedding model.
By default settings.embedding inherits settings.llm.baseUrl, so embeddings
point at your CHAT server. Run a DEDICATED embedding model on its own
llama.cpp server (note --embeddings --pooling mean):
llama-server -m nomic-embed-text-v1.5.Q8_0.gguf \
--embeddings --pooling mean --host 0.0.0.0 --port 8090
then in /admin/settings → Embedding set:
baseUrl = http://<host>:8090/v1
model   = nomic-embed-text
(server said: Failed after 3 attempts. Last error: This server does not support embeddings. Start it with `--embeddings`)
[exit 1]`

But when I go to the library settings and to the embedding section, no matter what option I select there are no fields to fill out other than the model name:

<img width="904" height="289" alt="Image" src="https://github.com/user-attachments/assets/65fa97be-142f-40f4-b3c4-0d9efb94eace" />

So I can't change the port number or anything to make it work.

perminder-klair:
Thanks for the detailed report, with the screenshot. This is fixed in #429 (on the `feat/locca-provider` branch), landing in the next release.

Two things were going on:

**1. The missing fields.** The embedding section now shows a dedicated **Embedding server base URL** field whenever the embedding provider is `locca` or `openai-compatible`. That field was hidden before, which is exactly why you only saw the model name. There's also a **Detect locca server** button and a **Test embeddings** button so you can confirm it works before kicking off a tagging run.

**2. The actual requirement.** A single llama.cpp / locca process can't serve both chat and embeddings, so embeddings need their own server. The error you hit was the tagger talking to your *chat* server. The fix is to run a dedicated embedding server. With locca that's one command:

```
locca embed nomic        # starts an embedding server on :8090
```

Then in **Settings → Library tagger → Embedding server**: with `locca` selected you can leave the base URL blank (it now defaults to `http://host.docker.internal:8090/v1`), or set it explicitly. Hit **Test embeddings**, then **Start tagging**.

If you'd rather not use locca, any dedicated llama.cpp embedding server works the same way:

```
[...]

PR #429, opened 2026-06-17
feat(llm): first-class locca provider + constrained pool picks

perminder-klair:
## What

Makes **locca** (local llama.cpp via the [locca](https://github.com/perminder-klair/locca) CLI) a first-class LLM provider in SUB/WAVE — a dedicated provider with auto-discovery, rather than a hand-typed `openai-compatible` URL — and hardens local-model track picking.

Validated end-to-end against a real locca server (Qwen3-30B-A3B on an iGPU): the DJ picker, track selection, and between-track links all run on the local model.

## Changes

**Provider**
- `'locca'` added to `LLM_PROVIDERS`; shares the openai-compatible capability descriptor (native objects, no `repeat_penalty`, no-think handled in transport via `noThinkFetch`).
- `registry`: `locca` reuses the openai-compatible builder (factored into `openAICompatibleModel`) with a default base URL of `host.docker.internal:8080/v1`, overridable via `llm.baseUrl`. baseUrl is optional for locca.

**Discovery**
- `GET /settings/llm/discover` probes a locca / openai-compatible server's `/v1/models` → `{ reachable, models }`.
- Onboarding step 2 + admin Settings get a `locca` option; onboarding adds a **Detect models** button that auto-fills the model id.
- `onboarding/test-llm` gains a `locca` case and now applies `noThinkFetch` to local servers, so a thinking model (Qwen3 etc.) returns visible content in the test instead of an empty reasoning-only reply.

**Reliability**
- The stateless pool picker constrains the chosen `id` to `z.enum(candidateIds)` built from the pool, so a local model can't hallucinate an id outside the candidate set (closes the "agent returned unknown id → fall back to pool" hole on that path). Empty pool falls back to `z.string()`.

## Out of scope (deferred by design)

- Managed lifecycle (SUB/WAVE starting/stopping locca) — not in this PR.
- Agent-path retry-with-enum — kept today's straight-to-pool fallback.
- A discovery button in admin Settings (onboarding has it); admin shows the locca base-URL field + hints.

## Depends on

- locca headless serve (perminder-klair/locca#11) for the `locca serve <model> --yes` startup the docs reference — runtime-independent (discovery degrades gracefully when no locca server is up).

## Test

- `npm run test` (controller): llm-pure + picker regression pass, incl. new `locca` cases (providerOptions, capability flags, agentPlan routing).
- `npm run lint` clean in `controller/` (eslint + tsc) and `web/`.

### DIFF
--- a/controller/scripts/llm-pure.test.ts
+++ b/controller/scripts/llm-pure.test.ts
@@ -12,6 +12,8 @@ import { withDeadline } from '../src/llm/internal/core/retry.js';
 import { providerOptions, needsToolCallObject, repeatPenaltyApplies, appliedNumCtx } from '../src/llm/internal/provider/capabilities.js';
 import { agentPlan } from '../src/llm/internal/strategy/plan.js';
 import { introBudgetPhrase, enforceIntroBudget } from '../src/llm/internal/prompts/intro-budget.js';
+import { embeddingBaseUrl } from '../src/llm/internal/provider/embedding.js';
+import { DEFAULT_LOCCA_EMBED_BASE_URL } from '../src/llm/internal/provider/registry.js';
 
 let failures = 0;
 function test(name: string, fn: () => void | Promise<void>) {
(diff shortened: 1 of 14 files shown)


perminder-klair:
Folded in a fix for #405 (embedding endpoint not configurable) — it surfaced directly from the locca work, since a locca/llama.cpp chat server can't also serve embeddings.

**Bug:** `settings.embedding` could never point at its own server. `embeddingCfg()` already reads `s.baseUrl`/`s.ollamaUrl`, but `DEFAULTS.embedding`, the load-normalizer, and the `update()` validator all omitted them, and the admin UI showed only the model field — so embeddings always inherited `settings.llm.baseUrl` (the chat server). The preflight error even told users to set a baseUrl field that didn't exist.

**Fix:**
- `settings.ts` — `baseUrl` + `ollamaUrl` added to `DEFAULTS.embedding`, the load-normalizer, and the `update()` validator (same URL rules as the LLM leg).
- `SettingsPanel` — fields added to the Embedding section (base-URL for openai-compatible/locca, Ollama-URL for ollama).
- `embeddings.ts` — `not_embedding_model` guidance now covers `locca` and points at `locca embed`.

Verified live on the dev stack: `POST settings.embedding.baseUrl` now persists and survives the load/normalize cycle (previously dropped).

Closes #405.

PR #455, opened 2026-06-19
fix(requests): honest intro/ack when requested artist isn't in the library

perminder-klair:
## What

When a listener requests an artist that isn't in the Navidrome library, the DJ now **owns the substitution** instead of pretending the filler track is by the requested artist.

## Why

Real example from the request log:

| field | value |
|---|---|
| request | "Do you have something from Kate Perry?" |
| artist (parsed) | Katy Perry |
| `artistMiss` | Katy Perry |
| pickSource | `library-mood:evening(context)` |
| aired track | **"The Game of Love" — Daft Punk** |
| intro read on air | *"…Katy Perry, stepping into the golden hour."* |

The matcher parsed the request correctly, but Katy Perry isn't in the library, so the stateless cascade fell through to the dominant-mood (`evening`) filler pick — **by design, the station never refuses**. The bug was that both:

- the up-front `ack` (written by `matchRequest` *before* the cascade knew it would miss), and
- the generated `introScript` (`generateIntro` was never told about the miss)

confidently announced Katy Perry over a Daft Punk track — even though the cascade had already detected and recorded `entry.artistMiss`.

## How

- **`routes/request.ts`** — on a miss, replace the lying `ack` with an honest stand-in line, and pass `artistMiss` into `generateIntro`.
- **`llm/internal/prompts/scripts.ts`** — `generateIntro` gains an `artistMiss` param; when set it adds a hard context line + prompt clause telling the DJ to briefly own the miss and never imply the track is by the requested artist.

The hit case (`artistMiss` null) is byte-for-byte unchanged. This brings the stateless cascade in line with the **agent path**, which already handles misses honestly (e.g. *"no Baile Inolvidable in the SUB/WAVE vault, but…"*).

## Test

`npm run lint` (eslint + `tsc --noEmit`) passes with 0 errors in `controller/`.

### DIFF
--- a/controller/src/llm/internal/prompts/context.ts
+++ b/controller/src/llm/internal/prompts/context.ts
@@ -14,7 +14,7 @@ export const ANGLES = {
     'Skip the introduction entirely and start mid-thought, as if continuing a conversation.',
     'React to the request itself — what kind of request it is, what mood it suggests — before mentioning the track.',
     'Use a short personal observation about the moment (Tuesday energy, the rain holding off, etc.) as the doorway.',
-    'Lean into contrast: how this track sits against what came before, or against the time of day.',
+    'Lean into contrast: how this track sits against the time of day, the weather, or the mood of the moment.',
     'Just say one true sentence and let the music start.',
   ],
   link: [
--- a/controller/src/llm/internal/prompts/scripts.ts
+++ b/controller/src/llm/internal/prompts/scripts.ts
@@ -33,7 +33,7 @@ export async function generateIntro({ track, context, requestedBy = null, reques
   const missClause = artistMiss
     ? ` The listener asked for "${artistMiss}", but we don't have them — briefly own that ("no ${artistMiss} in the crates", or similar), then introduce what's actually coming up as a worthy stand-in. Never pretend the track is by "${artistMiss}".`
     : '';
-  const prompt = `Write an intro for this track. ${lengthPhrase('intro')}${budget ? ' ' + budget : ''} If the listener said something specific, acknowledge their words naturally — don't quote them verbatim, but weave the gist in. Never read the request out loud as-is.${missClause}\n\n${ctxLines.join('\n')}`;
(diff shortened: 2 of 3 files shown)


perminder-klair:
## Tested against the live LLM 🎙️

Also pushed a follow-up (`1094572`) so the **request intro never talks about the previously-played track** — the recently-played context line was already flagged *"do not mention these artists or titles"*, but the random `intro` angle *"how this track sits against **what came before**"* pulled the DJ straight back to it. Dropped that clause from the angle + added an explicit *"don't back-announce or talk about the track that was just playing"* instruction. Back-announce stays the job of `generateLink` (untouched).

Since this is pure prompt-construction, I drove `dj.generateIntro()` directly against the live LLM (Ollama `glm-5.1:cloud` — the station's configured model), with the real persona/settings. Two scenarios × 4 samples, **this branch vs. parent commit `a4430e6`** for a before/after.

- **A** — asks for *Katy Perry*, library only has *Daft Punk* (the bug in the PR); *Queen* + *Avicii* as the just-played tracks.
- **B** — control: asks for *Daft Punk*, gets *Daft Punk*.

### This branch — 0/8 flagged ✅
- Miss owned honestly every time: *"Sam, no Katy Perry in the crates, I'm afraid — but Daft Punk delivered something just as luminous…"* — never implies the track is hers.
- Previous track never mentioned in any of the 8 (no Queen, no Avicii).

### Baseline (parent `a4430e6`, "what came before" angle forced to reproduce the 1-of-7 draw) — ❌
- **2/8 named it outright:** *"after that one-two of **Queen and Avicii**…"*, *"after that **Avicii** escalation…"*
- effectively all 8 referenced it obliquely: *"those two outsized anthems"*, *"all that theatrical sprawl"*, *"after that run of bigness"*.

So the pre-change code talked about the previous track on essentially every contrast-angle draw; the fix eliminates it while keeping the honest-substitution behaviour intact.

> Note: no committed regression test — these are non-deterministic LLM outputs, so a checked-in assertion would be flaky. `npm run lint` (eslint + `tsc --noEmit`) passes clean.

https://claude.ai/code/session_01P7Wh8wCBVCPv4qY7RnEGk8

PR #463, opened 2026-06-20
Fix transient stream status from stopping playback

### LINKED ISSUE #461: Web player becomes silent while MP3 stream appears to stay active
## Summary

Hi, thank you for building SUB/WAVE. I am trying it as a local self-hosted station and ran into a playback issue that I am not fully sure how to diagnose.

The web player can play the station for a short while after a refresh, but after some time the browser becomes silent. Refreshing the page usually makes the audio come back again for a while.

From the logs I can see the broadcast side still appears to be active and the MP3 stream is still being produced, so I am wondering whether this could be a web-player/browser stream recovery issue, a configuration issue on my side, or something already known.

## Environment

- Host OS: Fedora Linux, kernel `6.19.14-300.fc44.x86_64`
- Docker: `29.5.3`
- Docker Compose: `5.1.4`
- Deployment: Docker Compose, using published GHCR images
- SUB/WAVE source checkout: `main`, commit `e59bb42`
- Running containers:
  - `ghcr.io/perminder-klair/subwave-caddy:latest`
  - `ghcr.io/perminder-klair/subwave-web:latest`
  - `ghcr.io/perminder-klair/subwave-controller:latest`
  - `ghcr.io/perminder-klair/subwave-broadcast:latest`
- Access URL: `http://localhost:7700`
- Stream URL: `http://localhost:7700/stream.mp3`
- Subsonic backend: Navidrome on localhost
- Current stream settings observed in `radio.log`:
  - `crossfade_duration=3`
  - `archive_enabled=false`
  - `opus_enabled=false`
  - MP3 stream is being used

## What I observe

1. Open the SUB/WAVE web player.
2. Start listening to the station.
3. Audio works for a while.
4. After some time, playback becomes silent.
5. Refreshing the page makes audio come back again temporarily.

I also noticed that the listener count often drops back to `0` shortly after the browser had been listening, while the broadcast process keeps running.

Examples from `state/listeners.jsonl`:

```json
{"t":"2026-06-20T10:29:00.286Z","count":1}
{"t":"2026-06-20T10:30:15.264Z","count":0}
{"t":"2026-06-20T10:32:00.289Z","count":1}
{"t":"2026-06-20T10:33:15.266Z","count":0}
{"t":"2026-06-20T10:37:00.288Z","count":1}
{"t":"2026-06-20T10:38:00.288Z","count":1}
{"t":"2026-06-20T10:39:00.267Z","count":0}
{"t":"2026-06-20T10:47:00.293Z","count":1}
{"t":"2026-06-20T10:48:15.272Z","count":0}
```

The broadcast log still shows normal queueing/decoding/crossfade activity around the same time, for example:

```text
2026/06/20 18:52:04 [lang:3] on_meta: title=MONTAGEM XONADA | artist=MXZI、DJ Samir、DJ Javi26 | album=MONTAGEM XONADA | id=LBMtXOqOwJnMuJQtrSl4e5
2026/06/20 18:52:04 [cross:3] Computing crossfade transition over overlapping 2.98s buffered data at start and end.
2026/06/20 18:53:17 [lang:3] on_meta: title=Capital Letters | artist=Andy Davis | album=Heartbreak Yellow | id=GXQriib4j2PSZ4UitwoXvt
2026/06/20 18:53:17 [cross:3] Computing crossfade transition over overlapping 2.96s buffered data at start and end.
2026/06/20 18:57:10 [lang:3] on_meta: title=my leaving | artist=Mabanua | album=Jazzy Folklore | id=mCUv60PLRHCFdo8lK1p4D5
2026/06/20 18:57:10 [cross:3] Computing crossfade transition over overlapping 2.97s buffered data at start and end.
```

I also tested the stream endpoint with curl and it returns MP3 data:

```text
HTTP 200
Content-Type: audio/mpeg
```

## Why I am filing this

I noticed the player code already has recovery logic for `waiting`, `stalled`, and `error` events, and I also saw related historical issues such as:

- #212, about Opus streaming going silent in Firefox
- #335, about audio hitches

This report is not necessarily the same problem, because my setup is using MP3 with Opus disabled. But the symptom feels related: the stream/source seems alive, while the web listener becomes silent or disconnects until refresh.

## Questions

- Is this a known limitation of browser playback with the current Icecast/MP3 stream setup?
- Is there a recommended way to diagnose whether the disconnect is caused by the browser, Caddy, Icecast/Liquidsoap, or the web player's recovery logic?
- Would it make sense for the web player to also have a "progress watchdog" that reconnects when the audio element appears to be playing but `currentTime` or `timeupdate` has not advanced for several seconds?
- Are there any recommended settings for crossfade/archive/stream format to reduce this issue?

Happy to provide more logs or test a patch if that helps.

kamjin3086:
## Summary

This PR makes the public listener player more tolerant of transient stream-status failures so a healthy audio element is not stopped by a short-lived `/api/now-playing` / Icecast status mismatch.

## Background

While investigating #461 on a local Docker deployment, I found that the direct Icecast mount (`/stream.mp3`) could keep playing reliably, while the React listener UI would stop after a short time. Instrumenting the player showed that the UI was not being disconnected by Icecast directly. Instead, one transient `streamOnline: false` update from the cached listener/status monitor caused `PlayerApp` to call `stop()`, which paused and cleared the active `<audio>` element.

That makes the status probe a single point of failure for playback, even when the actual audio stream is still alive.

## Changes

- Treat Icecast status fetch failures as unknown rather than proof that the stream is offline.
  - The listener monitor keeps the previous `online` / `bitrate` values and resets the current listener count when the status endpoint cannot be read.
- Debounce `streamOnline: false` in the listener UI.
  - The UI now requires several consecutive offline polls before presenting the station as offline.
- Remove the automatic `stop()` side effect from `PlayerApp` when `streamOnline === false`.
  - The audio element's own `waiting` / `stalled` / `error` handling remains responsible for actual stream recovery.
  - The status flag still informs UI state, but it no longer tears down active playback by itself.

## Why this should be safe

A transient status-probe failure is not equivalent to the broadcast source being gone. Keeping playback alive avoids false-positive disconnects, while real stream failures still surface through the browser media element and the existing player watchdog/error path.

## Verification

Tested on a local SUB/WAVE Docker deployment where the issue was reproducible:

- Confirmed `/stream.mp3` stayed alive while the UI stopped before this patch.
- Added temporary instrumentation and confirmed `offline-effect-stop` was the stop trigger.
- Applied this patch and verified listener UI playback stayed active for 90+ seconds:
  - `paused=false`
  - `readyState=4`
  - `currentTime` advanced continuously
- Ran `docker build -f web/Dockerfile ...` successfully with this change included.
- Ran controller checks locally:
  - `npm --prefix controller run lint` (passes with existing warnings)
  - `npm --prefix controller run test`

Fixes #461

### DIFF
--- a/controller/src/broadcast/listeners.ts
+++ b/controller/src/broadcast/listeners.ts
@@ -91,7 +91,14 @@ async function fetchCount() {
     lastStatus = { online, listeners: { current, peak: peakSeen }, bitrate };
   } catch {
     lastCount = null;
-    lastStatus = { online: false, listeners: { current: 0, peak: 0 }, bitrate: null };
+    // Treat Icecast status fetch failures as "unknown", not as proof the
+    // broadcast went offline. The listener UI polls this cached status and
+    // should not tear down a healthy audio element because the stats endpoint
+    // had a transient timeout.
+    lastStatus = {
+      ...lastStatus,
+      listeners: { current: 0, peak: peakSeen },
+    };
   }
   // Persist at most one row per wall-clock minute. Skip null samples — a
   // stats outage shouldn't leave a misleading "0 listeners" stripe in the
--- a/web/components/PlayerApp.tsx
+++ b/web/components/PlayerApp.tsx
@@ -50,7 +50,7 @@ export default function PlayerApp({ contained = false }: PlayerAppProps) {
   const { apiUrl } = useStationOrigin();
   const { nowPlaying, context, dj, activeShow, listeners, streamOnline, llmTokens, state, session, trackStartedAt, timezone } = useStationFeed();
   const boothFeed = session.messages;
-  const { audioRef, tunedIn, status, volume, setVolume, tune, stop, toggleMute, muted, idleStopped } = usePlayer();
+  const { audioRef, tunedIn, status, volume, setVolume, tune, toggleMute, muted, idleStopped } = usePlayer();
 
   // streamOnline is null until the first poll resolves — only treat an
   // explicit false as offline so the player never flashes "offline" on load.
@@ -64,12 +64,6 @@ export default function PlayerApp({ contained = false }: PlayerAppProps) {
   // normalise the feed's number | { current } | null shape to a plain count.
   const listenerCount =
     listeners == null ? null : typeof listeners === 'number' ? listeners : (listeners.current ?? null);
-
-  // If the station goes off air while someone is tuned in, tear playback down
-  // so the <audio> element isn't left retrying a dead mount.
-  useEffect(() => {
-    if (offline && tunedIn) stop();
-  }, [offline, tunedIn, stop]);
 
   // One-shot audience beacon: hand the controller the external referrer + any
   // UTM tag on first load. The referrer is browser-only knowledge — by the time
--- a/web/hooks/useStationFeed.ts
+++ b/web/hooks/useStationFeed.ts
@@ -38,6 +38,7 @@ export interface StationFeed {
 
 const EMPTY_STATE: StationState = { upcoming: [], history: [], djLog: [] };
 const EMPTY_SESSION: SessionPayload = { session: null, messages: [] };
+const OFFLINE_CONFIRM_POLLS = 4;
 
 // Only commit a freshly-parsed payload when it differs from what's already in
 // state — returning `prev` from the updater skips the re-render, so a quiet
@@ -64,6 +65,7 @@ export function useStationFeed(): StationFeed {
   const [trackStartedAt, setTrackStartedAt] = useState<number | null>(null);
   const [timezone, setTimezone] = useState<string | null>(null);
   const lastTrackKeyRef = useRef<string | null>(null);
+  const offlinePollsRef = useRef(0);
 
   useEffect(() => {
     const tick = async () => {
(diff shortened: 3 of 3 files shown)


perminder-klair:
Thanks for the excellent diagnosis and fix, @kamjin3086 — pinning it on a transient /api/now-playing status blip tearing down a healthy <audio> element is exactly right, and the instrumentation write-up made it easy to verify.

One workflow note: this repo merges to develop, not main,main is owned by release-please via develop→main release PRs. I've rebased your commit (your authorship is preserved) onto develop and layered two small follow-ups + a unit test in #466:

1. A fully-unreachable Icecast still reads offline. With this PR alone, if the status endpoint throws on every poll, online was held at its last value indefinitely, so the offline indicator could stay pinned "online" for a genuinely dead station. #466 bounds the tolerance to ~1 min of consecutive failures, after which it reports offline again. (The DJ fail-open path is untouched.)
2. No flashed 0. The cached listener count now holds its last-known value on a transient blip instead of resetting to 0, and only zeroes once the outage is confirmed sustained.
3. Extracted the transition into a pure statusAfterFailure() helper with unit tests, so the transient-vs-sustained branching is covered.

Closing in favour of #466 — credit stays with you. Thanks again for the great report and fix! 🙏

PR #470, opened 2026-06-20
feat(dj): per-persona tone dials (humour, local colour, warmth) + anti-cliché

perminder-klair:
## What

Adds three per-persona **tone dials** — humour, local colour, warmth — to the DJ persona, plus extends the default prompt's anti-cliché rule. The prompt-layer "persona settings" upgrade discussed alongside the locca picker work.

## Sliders

Each dial is **0–10, default 5 (neutral)**, sitting next to the existing talk-frequency / script-length / dj-mode knobs. They map to **three prompt bands**: 0–3 low, 7–10 high, 4–6 neutral. Only a band away from neutral appends a one-line directive (e.g. humour-high → "Lean into dry, playful wit; an aside or a wink is welcome").

- A persona left at the default **5/5/5 renders a byte-identical system prompt** — existing stations are unchanged, no migration.
- A model can't tell humour=6 from 7, so the 3-band mapping is the honest one; the editor surfaces the active band so operators don't expect fine gradations to matter.

## Why these three 

Dropped `news /10` (news is a dedicated between-track skill with its own cadence, not patter) and `intro style /10` (`scriptLength` already owns verbosity). Kept the three that are genuinely new personality axes with no overlap.

## Scope boundary (deliberate)

Tone is appended in `renderDjPrompt` (the free-text DJ voice: intros, links, idents, weather) and **kept off the tool-loop agent preamble** (picker / request / segment director). That preamble already warns that extra persona text derails small models — it's why its `rules` block is opt-out — so loading tone there would risk the picker. The spoken-voice path is where these dials belong.

## Anti-cliché

Extended the default template's existing rule to also ban back-announce tells (`"and that was"`, `"that was [song] by [artist]"`). Affects the default template; custom templates keep their own wording.

## Files

- `controller/src/settings.ts` — `TONE_DIALS`, pure `personaToneDirectives()` / `normalizeDial()`, wired into `normalizePersona()` + `renderDjPrompt()`; anti-cliché line.
- `controller/scripts/llm-pure.test.ts` — pins the band mapping and clamp.
- `web/components/admin/PersonasPanel.tsx` — three range sliders in the persona editor (load / default / save wiring).

## Verification

- `npm run test:llm`: all pass, including new `normalizeDial` / `personaToneDirectives` band tests.
- Controller `npm run lint` (eslint + tsc): 0 errors.
- Web `npm run lint` (eslint + tsc): 0 errors. `npm run build`: exit 0, `/admin/personas` compiles.

### DIFF
--- a/controller/scripts/llm-pure.test.ts
+++ b/controller/scripts/llm-pure.test.ts
@@ -14,7 +14,7 @@ import { agentPlan } from '../src/llm/internal/strategy/plan.js';
 import { introBudgetPhrase, enforceIntroBudget } from '../src/llm/internal/prompts/intro-budget.js';
 import { embeddingBaseUrl } from '../src/llm/internal/provider/embedding.js';
 import { DEFAULT_LOCCA_EMBED_BASE_URL } from '../src/llm/internal/provider/registry.js';
-import { personaToneDirectives, normalizeDial, DIAL_NEUTRAL } from '../src/settings.js';
+import { personaToneDirectives, normalizeDial, DIAL_NEUTRAL, validatePersonasStrict } from '../src/settings.js';
 
 let failures = 0;
 function test(name: string, fn: () => void | Promise<void>) {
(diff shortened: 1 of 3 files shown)


perminder-klair:
## Tested live — found & fixed a blocking bug (pushed as `10e3aab`)

Booted the stack from a worktree on this branch and ran the dials through a real save round-trip. Lint, `test:llm`, and `build` were all green, and the UI renders correctly — but **the dials don't persist**.

### The bug
Saving a persona silently drops `humour`/`localColour`/`warmth`. The slider moves, you hit **Save**, and the value resets to neutral — so `renderDjPrompt` never sees a non-neutral value and the feature is dead end-to-end.

**Root cause:** the dials were wired into `normalizePersona()` (the read/load path) and `renderDjPrompt()`, but **not** into `validatePersonasStrict()` (`settings.ts`, the `POST /settings` write path). That validator rebuilds each persona from an explicit field whitelist, and the three new fields weren't in it.

Proven server-side, independent of the UI — a direct `POST /settings` with `humour:9` returns `200`, then reads back with the dials gone:
```
=== re-read Marlowe after POST(humour=9,localColour=8,warmth=1) ===
humour: None  localColour: None  warmth: None
```
The existing checks couldn't catch this: lint/build don't exercise a save, and the `renderDjPrompt` unit test only ever sees a hand-built persona that already has the dials — which never survives a real save.

### The fix (`10e3aab`)
Carry the three dials through `validatePersonasStrict` via `normalizeDial` (absent → `DIAL_NEUTRAL`, so legacy personas still render byte-identical):
```diff
       djMode,
+      humour: normalizeDial(item.humour),
+      localColour: normalizeDial(item.localColour),
+      warmth: normalizeDial(item.warmth),
       soul,
```
Also exported the validator and added a regression test pinning that the dials round-trip the save path and clamp/default correctly — the test seam that was missing.

### Verified after the fix
- `npm run test:llm` — all pass, incl. the new `validatePersonasStrict carries the dials through the save path` test
- `npm run lint` (eslint + tsc) — 0 errors
- Live round-trip: save → `state/settings.json` on disk → UI reload all hold `9 / 8 / 1` ✅

PR #473, opened 2026-06-20
feat(tts): per-engine + per-persona DJ voice gain dial

perminder-klair:
## Why

DJ voice loudness is inconsistent **between TTS engines**. Only PocketTTS self-normalises its output, so it sits noticeably quieter than raw Piper/Kokoro/Chatterbox under the mic chain's fixed-threshold compressor — operators have been baking normalisation into their own TTS Docker images to compensate. This adds a first-class gain dial so they don't have to.

A static gain via Liquidsoap's `liq_amplify` annotation was chosen over an in-controller normaliser because engines emit **mixed formats** (Piper/Kokoro/Chatterbox/PocketTTS + OpenAI → WAV, **ElevenLabs → MP3**) and the controller image has no ffmpeg/decoder. `liq_amplify` is format-agnostic (Liquidsoap decodes everything) and reuses the exact mechanism the music loudness path already uses (`getAnnotatedUri` → `amplify(override="liq_amplify")`).

## What

- **settings**: per-engine `tts.gainDb` map (`piper`/`kokoro`/`chatterbox`/`pocket-tts`/`cloud`) + per-persona `tts.gainDb`, clamped ±12 dB (`TTS_GAIN_CLAMP_DB`/`clampTtsGain`). Old saves load at unity; the two gains **stack**.
- **tts**: `voiceGainDb(kind)` = resolved-engine gain + persona trim, clamped. No change to `speak()`'s contract.
- **queue**: `airVoice` wraps the handoff path as `annotate:liq_amplify="<n> dB":<path>` when non-zero (bare path at 0 → byte-for-byte today's behaviour). Clip duration is still read from the bare WAV header **before** wrapping.
- **radio.liq**: `amplify(override="liq_amplify", 1., …)` on both voice queues before `mic_chain`, mirroring the music line. The un-annotated silent lead-in stays at unity.
- **web**: per-engine "Voice level (dB)" sliders in Settings + a per-persona trim in Personas.

## Test plan

- ✅ `controller`: `npm run lint` (eslint + `tsc --noEmit`) clean — 0 errors.
- ✅ `web`: `npm run lint` (eslint + `tsc --noEmit`) clean.
- ⬜ Live smoke (dev): set PocketTTS `+6 dB` + a persona `+2 dB`, `docker compose -f docker-compose.dev.yml restart broadcast`, confirm broadcast log shows `DJ says: annotate:liq_amplify="8 dB":/…` and the voice is audibly louder vs a Piper persona at 0 dB.
- ⬜ No-op safety: all gains at 0 → `say.txt`/`intro.txt` carry a bare path (no `annotate:`).

### DIFF
--- a/controller/src/audio/tts.ts
+++ b/controller/src/audio/tts.ts
@@ -80,6 +80,23 @@ function resolveEngine(kind: string, personaTts: any) {
   return chosen;
 }
 
+// Effective voice level trim (dB) for a spoken segment of `kind`: the resolved
+// engine's per-engine gain (settings.tts.gainDb) plus the on-air persona's own
+// trim (persona.tts.gainDb), clamped to ±TTS_GAIN_CLAMP_DB. Applied downstream
+// by broadcast/queue.ts as a Liquidsoap `liq_amplify` annotation on the handoff
+// file — the same mechanism music loudness uses. 0 = unity (no annotation
+// written), i.e. today's behaviour. Uses the *resolved* engine (post
+// availability/key fallback), so the gain matches the engine that will actually
+// speak; the rare runtime-throw fallback inside speak() is an error path.
+export function voiceGainDb(kind: string): number {
+  const personaTts = djPersonaTts(kind);
+  const engine = resolveEngine(kind, personaTts);
+  const tts: any = settings.get().tts || {};
+  const engineGain = settings.clampTtsGain(tts.gainDb?.[engine]);
+  const personaGain = personaTts ? settings.clampTtsGain(personaTts.gainDb) : 0;
+  return settings.clampTtsGain(engineGain + personaGain);
+}
+
 async function speakWith(engine: string, text: string, opts: any, personaTts: any) {
   if (engine === 'kokoro') {
     const voice = (personaTts && personaTts.engine === 'kokoro' && personaTts.voice)
--- a/controller/src/broadcast/queue.ts
+++ b/controller/src/broadcast/queue.ts
@@ -9,7 +9,7 @@ import { config } from '../config.js';
 import * as subsonic from '../music/subsonic.js';
 import * as mix from '../music/mix.js';
 import * as library from '../music/library.js';
-import { speak } from '../audio/tts.js';
+import { speak, voiceGainDb } from '../audio/tts.js';
 import * as djAgent from './dj-agent.js';
 import * as sfx from './sfx.js';
 import * as session from './session.js';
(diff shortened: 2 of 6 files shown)


perminder-klair:
## ✅ Live smoke test — PASS

Ran the two unchecked items from the test plan (and then some) on a worktree of this branch (`6300789`) with the dev stack (`broadcast` + `controller`, `--build`). Drove the real path via **`POST /dj/say {mode:"raw"}`** (speaks verbatim → bypasses the LLM, renders via Piper → straight through `queue.announce → airVoice(…, voiceGainDb(kind))`) and read back the exact URI Liquidsoap consumed from `radio.log`.

### Core annotation behaviour
```
0 dB     → DJ says: /var/sub-wave/voice/869597ce3bdb.wav                                  (bare path, no annotate:)
+6 dB    → DJ says: annotate:liq_amplify="6 dB":/var/sub-wave/voice/f62797949a3e.wav
kind:link→ DJ talk-over: annotate:liq_amplify="6 dB":…                                    (intro.txt too)
-4 dB    → DJ says: annotate:liq_amplify="-4 dB":…                                        (sign correct)
99 dB    → annotate:liq_amplify="12 dB":…                                                 (clamped to ±12)
```

### Stacking + clamp
- engine `+3` + persona `+2` → `annotate:liq_amplify="5 dB"` ✅
- engine `10` + persona `5` (=15) → `annotate:liq_amplify="12 dB"` ✅ (outer clamp on the sum)
- reset both back to `0` → bare path again ✅ (no-op safety both directions)

### Validation / load
- Old save with no `gainDb` key → loads as all-zero map (unity). ✅
- `POST /settings {tts:{gainDb:{bogus:3}}}` → **HTTP 400** `unknown engine "bogus"; must be one of: piper, kokoro, chatterbox, pocket-tts, cloud` ✅
- Setting takes effect with `requiresRestart:false` — no mixer restart needed.

### Liquidsoap actually consumes it
For the +6 clip: ffmpeg decoded the WAV, `voice_queue Prepared "…f62797949a3e.wav" (RID 13)`, played to EOF, **zero errors/warnings** all session. The un-annotated `leadin.wav` plays at unity right before each clip, as documented.

### Web UI
Both sliders render and save end-to-end: Settings "Voice level (dB)" per engine → persisted `piper:4`; Personas per-persona trim → persisted `p_vesper.gainDb:3`. Full loop: UI-saved engine `+4` + persona `+3` → on-air segment logged `annotate:liq_amplify="7 dB"`.

### Notes
- 👍 **Resolved-engine logic confirmed through a real fallback:** the active persona is configured for `pocket-tts` (not installed in the dev build) so it falls back to Piper — and `voiceGainDb` correctly applied *Piper's* gain (the engine that actually speaks), matching the code comment.
- The only thing not measured is the *subjective* "audibly louder" delta (would need RMS capture of the ducked voice over music). The gain-application mechanism is identical to the in-production music loudness path (`radio.liq` `amplify(override="liq_amplify")`), and Liquidsoap applied it without error, so the contract holds.

🤖 Verified with Claude Code.

PR #491, opened 2026-06-21
fix(auth): harden admin auth with timing-safe comparison and brute-force lockoutFix/admin auth timing bruteforce

### LINKED ISSUE #488: Security: admin auth has no brute-force protection and passwords are stored as plaintext
While reviewing the auth flow, I found a few areas worth hardening. The requireAdmin middleware does a plain string comparison with no lockout or delay, making brute-force attacks trivial. Passwords are stored as raw strings in .env with no hashing; if the file leaks, credentials are immediately usable. Authentication is also opt-in per route rather than applied globally, so any new route added without requireAdmin is silently unprotected. Finally, there is no UI to change ADMIN_USER, ADMIN_PASS, or Icecast passwords after setup; only .env editing plus container restart works.

**Planning to raise a PR for this. Approach:**

  - Replace plain string compare in auth.ts with crypto.timingSafeEqual and add a
  per-IP attempt counter with a temporary lockout after 5 failed tries
  - Hash ADMIN_PASS at startup using bcrypt, compare at runtime
  - Move requireAdmin to a single group guard in server.ts instead of per-route opt-in
  - Add a POST /settings/credentials endpoint that validates the current password before
  updating, and a matching "Change Password" UI in the admin settings panel

  **Happy to discuss the approach before opening the PR.**

geekylakshya:
## What 
Replace plain `===` string comparison in admin auth with `crypto.timingSafeEqual` and add per-IP brute-force lockout after 5 consecutive failed attempts.

## Why
The `requireAdmin` middleware was vulnerable to timing side-channel attacks (plain string comparison leaks credential info via response timing) and had no protection against brute-force attempts. Any attacker with curl could enumerate credentials with no rate limiting.

Fixes #488 (partial)

## How tested
  - `cd controller && npm run lint` - 0 errors, 554 pre-existing warnings
  - Code review: timing-safe helper handles length mismatch with dummy comparison to
  prevent length leaking
  - Code review: lockout map follows same in-memory pattern as existing `ratelimit.ts`

## Notes for reviewers
  - This is PR 1 of a potential series for #488. Scope is intentionally narrow: timing-safe
  compare + brute-force lockout only.
  - Reuses the existing `clientIp()` helper from `ratelimit.ts` for IP extraction.
  - Lockout state is in-memory (resets on container restart), matching the existing rate
  limiter pattern.
  - Map cleanup triggers at 500 entries to prevent unbounded growth.
  - The web UI admin login doesn't currently handle 429 responses with a specific message.
  A follow-up could add that, but the lockout works correctly regardless.

### DIFF
--- a/controller/src/middleware/auth.ts
+++ b/controller/src/middleware/auth.ts
@@ -11,7 +11,10 @@ const ADMIN_USER = process.env.ADMIN_USER || '';
 const ADMIN_PASS = process.env.ADMIN_PASS || '';
 export const ADMIN_AUTH_REQUIRED = Boolean(ADMIN_USER && ADMIN_PASS);
 const IS_PROD = process.env.NODE_ENV === 'production';
-const MAX_AUTH_FAILURES = 5;
+// 10 strikes before a temporary lockout: brute-forcing a real password in 10
+// tries is implausible, while an operator (or a household behind one NAT IP)
+// fat-fingering the password a few times shouldn't get locked out for 15 min.
+const MAX_AUTH_FAILURES = 10;
 const AUTH_LOCKOUT_MS = 15 * 60 * 1000;
 const authAttempts = new Map<string, { failures: number; lockedUntil: number }>();
 
@@ -42,6 +45,12 @@ export function assertAdminConfigured() {
 export function requireAdmin(req, res, next) {
   if (!ADMIN_AUTH_REQUIRED) return next();
 
+  // Lockout keys on clientIp() — the left-most X-Forwarded-For — which is
+  // client-controlled and therefore spoofable: an attacker can rotate the
+  // header per request to dodge this counter. So this is defense-in-depth that
+  // slows casual brute-forcing from a single source, not a hard guarantee. For
+  // durable enforcement put a real rate limit at the edge (Caddy/Cloudflare),
+  // where the connecting IP is known before it gets flattened into a header.
   const ip = clientIp(req);
   const now = Date.now();
   const rec = authAttempts.get(ip);
--- a/controller/src/middleware/auth.ts
+++ b/controller/src/middleware/auth.ts
@@ -1,3 +1,5 @@
+import { timingSafeEqual } from 'node:crypto';
+
 // Admin basic auth. In production (NODE_ENV=production) ADMIN_USER and
 // ADMIN_PASS are MANDATORY — the controller refuses to start without them,
 // because /debug, /settings, and the jingle/tagger endpoints expose enough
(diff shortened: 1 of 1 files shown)


perminder-klair:
Pulled the branch and gave it a proper go. Lint's clean, and I drove `requireAdmin` directly with mock requests, correct creds pass, ten bad tries from one IP trips the 429 with a `Retry-After`, a locked IP is refused even with the right password, other IPs are unaffected, and a success clears the counter. The core logic does exactly what it says.

I pushed three small commits on top of yours (hope that's alright, maintainer edits were on):

- **Bumped the lockout from 5 to 10.** Against a real password the difference is noise, but 5 is easy to trip by accident, an operator mistyping, or a few people sharing one NAT IP and a 15-minute lockout for a fat-fingered password is annoying. 10 gives some headroom without meaningfully helping an attacker.
- **Documented the IP key.** Added a comment by the lookup spelling out that the key is the left-most `X-Forwarded-For`, which is client-controlled and therefore spoofable — someone brute-forcing can just rotate the header and walk past the counter. So this is defense-in-depth that slows casual probing from one source, not a hard control. Real enforcement wants to live at the edge (Cloudflare WAF, or `caddy-ratelimit`), where the connecting IP is known before it's flattened into a header.
- **Fixed a colon-in-password bug.** The parser was doing `split(':')`, which truncated any password containing a colon and rejected otherwise-correct creds. It's pre-existing (already on `develop`), but easy to fix while we're in here: split on the first colon only, per RFC 7617. Added a test case to confirm a multi-colon password now authenticates.

Two non-blocking notes for the record: behind a shared NAT, one bad actor locks everyone on that IP for 15 minutes (fine for a homelab, just a known trade-off), and every request with a missing/bad header counts as a failure, so a browser hitting an admin route before login nibbles at the counter, worth a glance at how the admin shell makes its first call.

On the rest of the #488 plan: when you get to the bcrypt step, hashing `ADMIN_PASS` at startup doesn't actually close the "if `.env` leaks, creds are usable" gap, since you still need the plaintext in `.env` to hash it on boot, storing the hash in `.env` would, but that collides with both setup wizards and `state/icecast-secrets.env`, so let's design that one before you build it. And for the global-guard step, keep it default-deny with an explicit public allowlist: `/health`, `/now-playing`, `/state`, `/dj`, and `/cover/:id` have to stay open or the player breaks.

Nice first cut. I think this is good to merge, give my commits a quick look and let me know.

geekylakshya:
Reviewed both commits, looks great. The colon-in-password fix is a good catch; it would have silently rejected valid credentials. The threshold bump to 10 makes sense for a homelab env.

**One question:** after the lockout expires, `failures` is still `>= 10`, so the very next wrong attempt immediately re-triggers a 15-minute lockout. The operator effectively gets 1 try every 15 minutes after the first lockout instead of another set of 10 attempts.

Is that deliberate, or a missed edge case?

If it is unintentional, resetting the counter when the lockout expires would fix it:

```ts
if (rec && rec.lockedUntil > 0 && rec.lockedUntil <= now) {
  rec.failures = 0;
  rec.lockedUntil = 0;
}
```

Happy to push this fix as a follow-up PR, or leave it to you.

PR #514, opened 2026-06-22
feat(controller): fetch Last.fm artist tags directly via the Last.fm API

perminder-klair:
## What

Library tag enrichment now fetches Last.fm artist tags **directly from the Last.fm REST API** instead of routing through Navidrome.

- **New `controller/src/music/lastfm.ts`** — read-only client:
  - `getArtistTopTags(artist, { count = 10 })` → `artist.getTopTags` (GET, `autocorrect=1`, `format=json`, `api_key` only — no signing, no session key). Parses `toptags.tag[]`, drops `count:0` noise + empties, lowercases/trims, slices to `count`. 5s `AbortController` timeout; returns `[]` on any failure (no retry, per project convention).
  - `hasLastfmKey()` — resolves `LASTFM_API_KEY || settings.scrobble.lastfm.apiKey`.
- **`music/tag-library.ts` `phaseEnrich`** — prefers the direct path when a key is present; keeps the Navidrome `getArtistInfo2` path as the fallback (custom Navidromes that *do* expose `tag[]`). Per-artist `artistTagCache` preserved.
- **`settings.ts`** — `embedding.enrichment.lastfmTags` default `false` → `null` (tri-state: `true` = always, `false` = never, `null` = auto). Gate: `lastfmTags === true || (lastfmTags !== false && hasLastfmKey())`.

## Why

Pulling tags through Navidrome's `getArtistInfo2` always returned empty on vanilla Navidrome — its agent only surfaces bio + images there, never the `tag[]` array — so the tagger always reported `lastfm: 0` and the feature was off by default. Operators already configure a Last.fm api_key for scrobbling; reusing it for read-only tag lookups actually works.

## Test plan

- `npm run lint` (controller) passes — eslint (0 errors) + `tsc --noEmit` (0 errors).
- Mocked unit smoke test: parsing, `count:0`/empty filtering, lowercase+trim, slice-to-10, read-only params (no `sk=`/`api_sig=`), error-field → `[]`, keyless short-circuit.
- Live request-shape test against the real Last.fm endpoint (invalid key → `403` → `[]`, graceful).
- Success path (real tags for a real artist) pending a valid api_key on the box — no key is currently configured here.

### DIFF
--- a/controller/src/music/lastfm.ts
+++ b/controller/src/music/lastfm.ts
@@ -1,0 +1,93 @@
+// Direct Last.fm tag client (read-only).
+//
+// Library tag enrichment wants Last.fm's crowd tags for an artist. The older
+// path went *through* Navidrome (subsonic.getArtistLastfmTags → getArtistInfo2),
+// but vanilla Navidrome's agent only surfaces bio + images there, never the
+// tag[] array — so tags always came back empty. This module skips Navidrome and
+// hits the Last.fm REST API directly, reusing the api_key the operator already
+// configured for scrobbling (LASTFM_API_KEY / settings.scrobble.lastfm.apiKey).
+//
+// Read methods (artist.getTopTags) are unauthenticated beyond the api_key — no
+// md5 signing, no session key (unlike the write calls in broadcast/scrobble.ts).
+// Every call is fire-and-forget with a 5s timeout and returns [] on any failure
+// so the caller can fall back; no retry (project convention — Last.fm's read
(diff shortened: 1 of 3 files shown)


MMagTech:
Does this now mean that in order to pull from last.fm we have to enable scrobbling?  Would be idea if it were possible to enable the setting but, disable scrobbling. This doesn't seem like an idea scenario where the two actions should be joined.  Would be idea if they were two independent features each still requiring the API and Secret key.

PR #521, opened 2026-06-22
feat(auth): replace plaintext ADMIN_PASS with scrypt-hashed credentials

### LINKED ISSUE #488: Security: admin auth has no brute-force protection and passwords are stored as plaintext

geekylakshya:
## What

Replace plaintext `ADMIN_PASS` comparison with scrypt-hashed credentials stored in `state/admin-hash.json`. Existing installs auto-migrate on first boot.

## Why

Closes #488 (items 2 and 3).

A leaked `.env` currently exposes usable admin credentials. Hashing removes the plaintext from the runtime auth path.

Scoped to hashing only per maintainer feedback. Change-password UI/endpoint and default-deny guard are separate PRs.

## How tested

* `cd controller && npm run lint` - 0 errors
* `cd cli && npx tsc --noEmit` - clean
* Verified scrypt params (`N=2^17`, `r=8`, `p=1`, `maxmem=256 MiB`) don't throw `ERR_CRYPTO_INVALID_SCRYPT_PARAMS`
* Checked migration path: hash file absent + `ADMIN_PASS` in env triggers hash-and-write, logs warning to remove env var manually
* Checked boot ordering: `initAdminCredentials()` runs before `app.listen()` so admin routes are never exposed during the boot window

## Notes for reviewers

* `maxmem: 256 * 1024 * 1024` is passed explicitly to all `scrypt()` calls (Node's default 32 MiB is too small for `N=2^17`, as you flagged)
* Migration does **not** auto-strip `ADMIN_PASS` from `.env`. It warns the operator to remove it manually. CLI `init` never writes it in the first place.
* `setup.sh` is intentionally unchanged. Pure bash can't hash with scrypt, and the controller auto-migrates on first boot.
* `maxmem` is omitted from on-disk `scryptParams` in `admin-hash.json` because it's a Node runtime cap, not a cost parameter.
* Change-password endpoint/UI and default-deny guard are deferred to follow-up PRs.

### DIFF
--- a/cli/src/commands/init.ts
+++ b/cli/src/commands/init.ts
@@ -10,7 +10,7 @@
 // After init, the operator runs `subwave start` (which brings docker up)
 // and `subwave setup` (the full wizard for Navidrome / LLM / TTS / DJ).
 
-import { existsSync, mkdirSync, writeFileSync } from 'node:fs';
+import { existsSync, mkdirSync, writeFileSync, chmodSync, renameSync } from 'node:fs';
 import { resolve } from 'node:path';
 import { homedir } from 'node:os';
 import crypto from 'node:crypto';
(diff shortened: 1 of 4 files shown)


perminder-klair:
Thanks for putting this together. The crypto side is solid, the atomic 0600 write, the maxmem fix, and the constant-time compare all look right.

After mulling it over though, I'm going to hold off on merging. For a self-hosted box the security win is pretty thin. Hashing only really helps if someone can read `.env` but can't get shell access, and on a single host that's a narrow gap. Meanwhile the cost is real: Basic auth sends creds on every request, so every admin call now runs a full scrypt, which is heavy on something like a Pi. It also adds a new stateful file in `state/`, and there's no password recovery path yet since the change-password UI got split off.

One concrete thing in the current diff too: `init` stops writing `ADMIN_PASS` to `.env`, but the CLI still reads it from there (`cli/src/api.ts`, `cli/src/doctor.ts`), so `subwave setup` would 401 on a fresh install.

The brute-force lockout from #491 already covers the higher-value half of the original issue, so I'm okay parking the hashing for now. Really appreciate the work, and I'd still gladly take the change-password UI on its own later.

geekylakshya:
Thanks for the thorough review and for laying out the reasoning. All fair points.

The scrypt-on-every-request cost on low-power hardware is something I underweighted, and the .env/CLI breakage on fresh install is a real bug in the diff. Good catch.

I agree that, with the brute-force lockout from #491 already in place, the marginal security gain doesn't justify the complexity right now.

Happy to bring back the change-password UI as a standalone PR when the time is right. Parking this one for now.

PR #525, opened 2026-06-22
fix(app): remove Dimezis BlurView that kills touch input on Android (New Arch)

### LINKED ISSUE #458: Android App, cannot connect.
@perminder-klair
Everything works fine from browser on the phone, player works, /health works this is all running locally as I am not exposing it outside.  Downloaded the Android App and enter in https://radio.{host}.com, just like on my phones browser and the "Resolving host" comes back ok.  but "Controller - /health" keeps failing, even though I can go to 
https://radio.{host}.com/health in the android browser and it returns {"status":"on-air"}

Running on locale server Docker Compose container.

Thanks!

perminder-klair:
## Problem

On the native player, the **entire UI goes dead to touch** — nothing is tappable (tune, volume, station switcher, FM-dial tabs), though the screen renders and now-playing updates. It's **device/ROM-flaky and OS-version-independent**: dead on a Pixel 7 Pro and Pixel 10, fine on a Pixel 8a and a OnePlus Pad 3 — same app build, same Android version, same station. (#458)

PR #490's `pointerEvents="none"` on the BlurViews was shipped as a **native** build and did **not** fix it.

## Root cause

The app forces the **New Architecture (Fabric)**. Fabric resolves whether a touch routes through the new- or old-arch event emitter by the **target view's id parity**. `expo-blur`'s `dimezisBlurView` is a custom **native** Android view whose internal Android-created child views get ids that break that resolution, throwing `Cannot find EventEmitter for receivedTouches` and **silently dropping the tap** (facebook/react-native#47218, RN ≥ 0.75.2 + New Arch). Whether it fires depends on which ids get allocated → device/ROM-flaky.

`pointerEvents="none"` can't fix it: the bug is event **routing**, not touch **capture** — the hazard is the native view's mere presence in the tree.

The two BlurViews were `absoluteFill` behind the **masthead** (station switcher + FM dial) and the **transport bar** (tune + volume) — i.e. *every* primary control — so the whole surface went dead.

## Fix

Remove both `BlurView`s; keep the existing translucent glass film. Since `expo-blur` ≥ 55 the bare `blurMethod="dimezisBlurView"` (no `BlurTargetView`) renders **no blur at all** (expo/expo#44165), so the frost has already been inert since the SDK 56 upgrade — **this is a touch-only fix with no visual change** from what's shipped. An in-code `NB` comment documents why a `BlurView` must not go back here.

## Verify

`tsc`/lint locally not run (Expo app has no `node_modules` on the build host); change is JS-only, deletes two JSX elements + one now-unused import per file (`StyleSheet`/`mode` still used). Confirm on an affected device via logcat while tapping a dead control:

```
adb logcat | grep -iE "receivedTouches|EventEmitter|SoftException"
```

`Cannot find EventEmitter for receivedTouches` on the current build = root cause confirmed; gone after this build = fixed.

Fixes #458. Supersedes #490.

### DIFF
--- a/app/src/player/PlayerScreen.tsx
+++ b/app/src/player/PlayerScreen.tsx
@@ -14,7 +14,6 @@
 // re-render the pages whose data actually changed (useStationFeed keeps
 // unchanged payloads reference-stable for exactly this reason).
 
-import { BlurView } from 'expo-blur';
 import { LinearGradient } from 'expo-linear-gradient';
 import * as Haptics from 'expo-haptics';
 import { memo, useCallback, useEffect, useMemo, useRef, useState } from 'react';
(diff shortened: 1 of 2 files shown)


perminder-klair:
Closing — on-device testing ruled the BlurView out as the cause (a native build with the BlurViews removed was still dead on an affected Pixel 7 Pro). The blur was already inert on SDK 56 anyway, so this is at most cosmetic cleanup, not a fix.

The real cause was found by bisection: `BottomSheetModalProvider` (gorhom bottom-sheet) wrapping the whole app intercepted touches on the New Architecture on some Android devices. The app's sheet is non-modal and never needed that provider. Fix is coming as a separate PR (branch `fix/android-bottomsheet-provider-touch`). See #458.

Issue #522, opened 2026-06-22
Bug: OpenRouter embedding provider doesn't work - "fetch failed"


geekylakshya:
## What

Selecting OpenRouter as the embedding provider in admin settings and clicking "Test embeddings" fails with:

`Can't reach provider "openrouter" - check network / baseUrl. (fetch failed)`

This happens even when:

* The OpenRouter API key is set and valid
* The network is working and OpenRouter is reachable from inside the container
* The same embedding model works when called manually via `fetch()`

The embedding provider dropdown offers OpenRouter as an option, but the backend does not actually support OpenRouter for embeddings. The request falls back to the Ollama embedding path, which causes the test to fail when no local Ollama instance is running.

The LLM/chat provider implementation already supports OpenRouter correctly. The issue appears to be limited to the embeddings provider implementation.

## Steps to Reproduce

1. Go to Admin Settings
2. Select `OpenRouter` as the embedding provider
3. Set the embedding model to `nvidia/llama-nemotron-embed-vl-1b-v2:free`
4. Click **Test embeddings**
5. Observe the error:
   `Can't reach provider "openrouter" - check network / baseUrl. (fetch failed)`

## Expected Behavior

Embedding requests should be sent to OpenRouter when OpenRouter is selected as the embedding provider, just as chat/LLM requests are.

I traced the issue to the embeddings provider implementation and would be happy to open a PR to add OpenRouter embedding support if that would be welcome.

perminder-klair:
Thanks for the clear write-up, and good catch. You're right: OpenRouter added an OpenAI-compatible embeddings endpoint, but our backend still treated it as chat-only, so the request never actually reached OpenRouter.

One note on the error you saw. That "can't reach provider openrouter (fetch failed)" line is from a build older than v0.26.0. In v0.26.0 we hid OpenRouter from the embedding picker instead, which was the wrong move now that they do support embeddings.

Opened #523 to fix it properly: OpenRouter embeddings now route through https://openrouter.ai/api/v1/embeddings, same transport as the OpenAI-compatible provider, and it's back in the picker. Default model is openai/text-embedding-3-small, but you can set your own (e.g. the nvidia/... one you tested). Appreciate you offering to PR it.

geekylakshya:
### The fix in #523 added `case 'openrouter'` to `buildEmbeddingModel()`, but there's a remaining bug that prevents it from actually working when the LLM provider and embedding provider use different API keys.

### The problem

[...]

PR #535, opened 2026-06-23
fix(settings): load embedding.apiKey from settings.json

geekylakshya:
## What

Add `embedding.apiKey` to the settings loader, updater, defaults, and redaction logic in `controller/src/settings.ts`.

## Why

The settings loader reads `embedding.provider`, `embedding.model`, `embedding.baseUrl`, and `embedding.ollamaUrl`, but it does not load `embedding.apiKey`.

As a result, `settings.get().embedding.apiKey` is always `undefined`, causing `embeddingCfg()` to fall back to `llm.apiKey`.

This breaks configurations where the embedding provider and chat LLM use different API keys. For example:

* OpenRouter for embeddings
* A local proxy or different provider for chat

In this setup, embedding requests are sent with the chat API key instead of the embedding API key and fail with a `401 Unauthorized`.

This is a remaining issue from #522 after #523 added OpenRouter embedding provider support.

## How Tested

* `cd controller && npx eslint src/settings.ts && npx tsc --noEmit` — 0 errors
* Will verify against a live OpenRouter embedding probe after deployment

## Notes for Reviewers

This follows the same pattern already used for `search.apiKey` and `llm.apiKey` across all four touch points:

* Defaults
* Loader
* Updater
* Redaction

The `'set'` guard in `update()` prevents the redacted placeholder value from overwriting the stored API key when settings are round-tripped through the admin UI.

### DIFF
--- a/controller/src/llm/internal/provider/embedding.ts
+++ b/controller/src/llm/internal/provider/embedding.ts
@@ -32,7 +32,15 @@ function embeddingCfg() {
     enabled: s.enabled !== false,
     provider: s.provider || llm.provider || 'ollama',
     model: s.model || '',
-    apiKey: s.apiKey || llm.apiKey || '',
+    // Key precedence: the saved settings field wins, then a dedicated
+    // `EMBEDDING_API_KEY` env var (the env path most installs use -- keys live in
+    // state/secrets.env, not settings.json), then the chat key. This is a
+    // runtime env read like config.ts does for SEARCH_API_KEY, so the env value
+    // never gets baked into the persisted settings.json. It covers every
+    // provider uniformly -- including openai-compatible/locca, which can't safely
+    // grab a provider-conventional env var (createOpenAI would otherwise reach
+    // for OPENAI_API_KEY against an arbitrary self-hosted server).
+    apiKey: s.apiKey || process.env.EMBEDDING_API_KEY || llm.apiKey || '',
     ollamaUrl: s.ollamaUrl || llm.ollamaUrl || '',
     baseUrl: s.baseUrl || llm.baseUrl || '',
   };
--- a/controller/src/llm/internal/provider/embedding.ts
+++ b/controller/src/llm/internal/provider/embedding.ts
@@ -160,7 +160,7 @@ export function buildEmbeddingModel(cfg: EmbeddingCfg) {
       // key is required — a missing one 401s with the 'unauthorized' message.
       const provider = createOpenAI({
         baseURL: 'https://openrouter.ai/api/v1',
-        apiKey: cfg.apiKey || 'unused',
+        apiKey: cfg.apiKey || process.env.OPENROUTER_API_KEY || 'unused',
         name: 'openrouter',
       });
       return provider.textEmbeddingModel(id);
--- a/controller/src/settings.ts
+++ b/controller/src/settings.ts
@@ -629,6 +629,7 @@ const DEFAULTS = {
     // e.g. Ollama). See issue #405.
     baseUrl: '',          // openai-compatible / locca embedding server URL (with /v1)
     ollamaUrl: '',        // Ollama embedding server URL (ollama provider)
+    apiKey: '',           // empty -> inherit settings.llm.apiKey
     seedCount: 0,         // 0 → auto max(200, ceil(sqrt(library)))
     knnNeighbours: 5,
     moodVoteThreshold: 0.6,
@@ -1155,6 +1156,10 @@ export async function load() {
         typeof stored.embedding?.ollamaUrl === 'string'
           ? stored.embedding.ollamaUrl.trim()
           : DEFAULTS.embedding.ollamaUrl,
+      apiKey:
+        typeof stored.embedding?.apiKey === 'string'
+          ? stored.embedding.apiKey.trim()
+          : DEFAULTS.embedding.apiKey,
       seedCount:
         Number.isFinite(stored.embedding?.seedCount) && stored.embedding.seedCount >= 0
           ? Math.floor(stored.embedding.seedCount)
@@ -1296,6 +1301,7 @@ export function getRedacted() {
   if (clone.llm?.fallback) clone.llm.fallback.apiKey = s.llm?.fallback?.apiKey ? 'set' : '';
   if (clone.tts?.cloud) clone.tts.cloud.apiKey = s.tts?.cloud?.apiKey ? 'set' : '';
   if (clone.search) clone.search.apiKey = s.search?.apiKey ? 'set' : '';
+  if (clone.embedding) clone.embedding.apiKey = s.embedding?.apiKey ? 'set' : '';
   if (Array.isArray(clone.webhooks)) {
     for (let i = 0; i < clone.webhooks.length; i++) {
       clone.webhooks[i].authHeader = s.webhooks?.[i]?.authHeader ? 'set' : '';
@@ -1983,6 +1989,11 @@ export async function update(patch) {
         throw new Error('embedding.ollamaUrl must start with http:// or https://');
       }
       next.embedding.ollamaUrl = v.replace(/\/+$/, '');
+    }
+    if (e.apiKey !== undefined && e.apiKey !== 'set') {
+      const v = String(e.apiKey).trim();
+      if (v.length > 200) throw new Error('embedding.apiKey must be 0-200 chars');
+      next.embedding.apiKey = v;
     }
     if (e.seedCount !== undefined) {
       const v = parseInt(e.seedCount, 10);


perminder-klair:
Thanks for tracking this down. The four touch points are clean and match how `search.apiKey` already works.

Before we merge though, I think this needs to go a bit further, because as it stands it won't fix the 401 for the way we actually run the station. Our keys live in env (`state/secrets.env`), not in settings.json. The loader now reads `settings.embedding.apiKey`, but if you never paste a key into settings that field stays empty, falls back to an empty `llm.apiKey`, and the OpenRouter embedding branch still sends `apiKey: 'unused'`. So the failing setup still fails.

The root cause is that the embedding builders don't honour env the way the chat path does. Chat OpenRouter is `createOpenRouter({ ...(cfg.apiKey ? { apiKey } : {}) })`, so a blank key falls through to `OPENROUTER_API_KEY`. The embedding side hardcodes `apiKey: cfg.apiKey || 'unused'`, which kills that fallback.

Could you fold the env path into this PR so it actually closes the issue? Two options:

1. Make the embedding builders read the provider env var, e.g. `apiKey: cfg.apiKey || process.env.OPENROUTER_API_KEY || 'unused'` (note: dropping `'unused'` alone won't do it, since `createOpenAI` with no key looks for `OPENAI_API_KEY`, not the OpenRouter one). Same idea for the openai-compatible and locca branches.
2. Or seed an `EMBEDDING_API_KEY` in the loader the way `SEARCH_API_KEY` is handled in config.ts, so the settings field gets populated from env.

Keep the settings loading you already have, it's correct and worth having. Just want the env case covered too before this goes in, since that's the path most installs use.

geekylakshya:
Thanks for the detailed review. You're right that fixing the settings loader alone does not cover the environment-variable-based setup.

I've pushed a second commit that addresses option 1 by making the OpenRouter embedding builder read `OPENROUTER_API_KEY` when `cfg.apiKey` is not configured:

```ts
apiKey: cfg.apiKey || process.env.OPENROUTER_API_KEY || 'unused',
```

This matches the behavior of the chat path, where leaving `cfg.apiKey` blank allows the SDK to pick up the environment variable automatically.

For the `openai-compatible` and `locca` branches, I intentionally left the logic as:

```ts
apiKey: cfg.apiKey || 'unused',
```

Those providers are primarily intended for self-hosted OpenAI-compatible servers, where there is no widely accepted provider-specific environment variable. Allowing `createOpenAI()` to implicitly fall back to `OPENAI_API_KEY` could be surprising if that variable is configured for a different provider.

If you'd prefer, I can follow up with a dedicated `EMBEDDING_API_KEY` environment variable for those branches instead.

### Validation

* ESLint: 0 errors
* `tsc --noEmit`: clean

perminder-klair:
This is great, thanks. Keep both your commits as they are. The settings loader is correct and the `OPENROUTER_API_KEY` fallback is exactly right for the OpenRouter case, so I don't want to drop it.

The only gap left is the other env-based providers. `openai-compatible`, `locca`, `openai` and `google` still have no env path, so an operator running embeddings on one of those with the key in `state/secrets.env` still hits the 401. Rather than teach each builder its own provider env var, and risk the implicit `OPENAI_API_KEY` grab we talked about, this goes with the `EMBEDDING_API_KEY` route from option 2. One dedicated var, resolved once in `embeddingCfg()`, covering every provider, sitting underneath your OpenRouter fallback so nothing you did changes.

I've written and verified the follow-up so you don't have to work it out. Could you apply this on top of your branch? `git apply` it, then commit with whatever message fits your style.

```diff
diff --git a/controller/src/llm/internal/provider/embedding.ts b/controller/src/llm/internal/provider/embedding.ts
index 9454365..059a0a2 100644
--- a/controller/src/llm/internal/provider/embedding.ts
+++ b/controller/src/llm/internal/provider/embedding.ts
@@ -32,7 +32,15 @@ function embeddingCfg() {
     enabled: s.enabled !== false,
     provider: s.provider || llm.provider || 'ollama',
     model: s.model || '',
-    apiKey: s.apiKey || llm.apiKey || '',
+    // Key precedence: the saved settings field wins, then a dedicated
+    // `EMBEDDING_API_KEY` env var (the env path most installs use — keys live in
+    // state/secrets.env, not settings.json), then the chat key. This is a
+    // runtime env read like config.ts does for SEARCH_API_KEY, so the env value
+    // never gets baked into the persisted settings.json. It covers every
+    // provider uniformly — including openai-compatible/locca, which can't safely
+    // grab a provider-conventional env var (createOpenAI would otherwise reach
+    // for OPENAI_API_KEY against an arbitrary self-hosted server).
+    apiKey: s.apiKey || process.env.EMBEDDING_API_KEY || llm.apiKey || '',
     ollamaUrl: s.ollamaUrl || llm.ollamaUrl || '',
     baseUrl: s.baseUrl || llm.baseUrl || '',
   };
diff --git a/controller/src/setup/secrets.ts b/controller/src/setup/secrets.ts
index 4865597..4462317 100644
--- a/controller/src/setup/secrets.ts
+++ b/controller/src/setup/secrets.ts
@@ -24,6 +24,10 @@ export const SECRET_ENV_KEYS = [
   'AI_GATEWAY_API_KEY',
   'ELEVENLABS_API_KEY',
   'SEARCH_API_KEY',
+  // Embeddings. Only needed when the embedding provider uses a different key
+  // than chat (e.g. OpenRouter for embeddings, a local proxy for chat) — see
+  // embedding.ts embeddingCfg(). Blank → embeddings inherit settings.llm.apiKey.
+  'EMBEDDING_API_KEY',
   // Scrobbling. See broadcast/scrobble.ts. Env always wins over settings.json,
   // so a host with these in compose env_file works without ever touching the UI.
   'LASTFM_API_KEY',
diff --git a/cli/src/util.ts b/cli/src/util.ts
index e4734b6..8a0d812 100644
--- a/cli/src/util.ts
+++ b/cli/src/util.ts
@@ -231,6 +231,7 @@ export const WIZARD_SECRET_KEYS = [
   'AI_GATEWAY_API_KEY',
   'ELEVENLABS_API_KEY',
   'SEARCH_API_KEY',
+  'EMBEDDING_API_KEY',
 ] as const;
 
 // Merge a batch of API keys into state/secrets.env (mode 0600), preserving
diff --git a/.env.example b/.env.example
index 92aa685..48558c5 100644
--- a/.env.example
+++ b/.env.example
@@ -77,6 +77,10 @@ SITE_URL=
 # AI_GATEWAY_API_KEY=
 # ELEVENLABS_API_KEY=
 # SEARCH_API_KEY=tvly-...
+# Embedding key — only when embeddings use a different provider/key than chat
+# (e.g. OpenRouter for embeddings, a local proxy for chat). Blank → inherit the
+# chat key. OpenRouter embeddings also fall back to OPENROUTER_API_KEY above.
+# EMBEDDING_API_KEY=
 
 # ───────── Acoustic analysis (CLAP audio-similarity + Demucs vocals) ─────────
 # The analysis pass (npm run analyze / admin "Analyze audio") computes bpm, key,
diff --git a/cli/src/assets.generated.ts b/cli/src/assets.generated.ts
index 914fdbf..2bddd5f 100644
--- a/cli/src/assets.generated.ts
+++ b/cli/src/assets.generated.ts
@@ -731,6 +731,10 @@ SITE_URL=
 # AI_GATEWAY_API_KEY=
 # ELEVENLABS_API_KEY=
 # SEARCH_API_KEY=tvly-...
+# Embedding key — only when embeddings use a different provider/key than chat
+# (e.g. OpenRouter for embeddings, a local proxy for chat). Blank → inherit the
+# chat key. OpenRouter embeddings also fall back to OPENROUTER_API_KEY above.
+# EMBEDDING_API_KEY=
 
 # ───────── Acoustic analysis (CLAP audio-similarity + Demucs vocals) ─────────
 # The analysis pass (npm run analyze / admin "Analyze audio") computes bpm, key,
```

What it does:

- `embeddingCfg()` reads `EMBEDDING_API_KEY` between the settings field and the chat key, so `cfg.apiKey` is populated for all five builders.
- Adds `EMBEDDING_API_KEY` to the secrets allowlist in `setup/secrets.ts`, plus the CLI mirror, otherwise `state/secrets.env` never sources it.
- Documents it in `.env.example`. `cli/src/assets.generated.ts` is the embedded copy of that file, so it has to move with it. Regenerate rather than hand-edit if you prefer: `npm --prefix cli run embed-assets`.

I checked it here: controller and cli `tsc` are clean, controller eslint is 0 errors, and `embed-assets` regenerates with no extra diff. Ping me once it's pushed and I'll re-review.

geekylakshya:
Applied your patch as a third commit.

The new `EMBEDDING_API_KEY` precedence in `embeddingCfg()` now sits between the saved settings value and the existing chat-key fallback, covering all embedding providers uniformly without introducing the risk of an implicit `OPENAI_API_KEY` fallback.

I also audited the rest of the controller for similar configuration gaps:

* All hosted-provider chat builders in `registry.ts` already use the conditional-spread pattern for API keys.
* The fallback path in `legs.ts` inherits the same behavior.
* The embedding probe endpoint routes through `embeddingCfg()`, so it benefits from the same fix.

I did not find any additional provider-specific API key gaps.

### Validation

* ESLint: 0 errors
* `tsc --noEmit`: clean

perminder-klair:
Thanks for all the iterations here @geekylakshya. I've rolled this up into #549 so we can land the complete fix in one go — your three commits are preserved there, authorship and all, with one extra commit for the part that was still missing.

The gap was the env path: `embeddingCfg()` reads `EMBEDDING_API_KEY`, but for keys in `state/secrets.env` the var also has to be on the `SECRET_ENV_KEYS` allowlist or `loadSecretsIntoEnv()` drops it. Without that entry the embedding request still goes out as `Bearer unused`. I confirmed it end to end in dev against a mock server: with the allowlist entry the operator's key lands on the wire, without it you get `Bearer unused`. #549 adds the allowlist entry, the CLI mirror, and the `.env.example` docs on top of your work.

Closing this in favour of #549. Really appreciate the fix — credited you as co-author.

geekylakshya:
Thanks @perminder-klair, that's the missing piece. The `SECRET_ENV_KEYS` allowlist is not obvious from the loader, and I likely would have shipped the same gap.

Happy to see it land as #549, and I appreciate the co-author credit.

I'll keep that in mind for any future secret additions.

perminder-klair:
thank @geekylakshya for all your work <3

PR #538, opened 2026-06-23
fix(queue): deduplicate track pushes to prevent stacking

### LINKED ISSUE #537: Bug: queue.push() allows duplicate tracks to stack in the queue
## What

`queue.push()` currently has no deduplication logic. The same track can be pushed into the queue multiple times, resulting in duplicate entries in `upcoming` even when the track is already queued or currently playing.

In some cases, this causes the same track to appear 3-4 times consecutively in the queue.

## Steps to Reproduce

1. Have a library with poor genre coverage (for example, 84% of tracks untagged).
2. Request a genre that the library cannot match well (for example, "hindi" or "bollywood").
3. Allow the DJ agent and fallback picker to select tracks.
4. Observe both paths converging on the same track.
5. The same track appears multiple times consecutively in the queue.

## Expected Behavior

`queue.push()` should check both `upcoming` and `current` before accepting a track.

If the track is already queued or currently playing, the push should be skipped (or the caller should be notified that the track was rejected as a duplicate).

## Where

`controller/src/broadcast/queue.ts` around line 280, inside the `push()` function.

Neither the agent path nor the fallback/cascade path appears to perform duplicate checks before calling `queue.push()`.

## Environment

* SubWave latest on `develop`
* Observed with a low-coverage Navidrome library (829 tracks, 84% untagged)

geekylakshya:
## What

Add a deduplication guard to `queue.push()` that rejects tracks already present in `upcoming` or `current`. Also feed queued track IDs into the picker, DJ agent, and request cascade so they prefer alternative selections instead of repeatedly choosing tracks that are already queued.

## Why

`queue.push()` previously had no duplicate check. When the DJ agent or request cascade converged on the same track (for example, due to poor genre coverage causing repeated fallback selections), the same song could be queued multiple times in succession.

This resulted in listeners hearing the same track back-to-back with little or no variety.

Fixes #537.

## How Tested

* `cd controller && npx eslint src/broadcast/queue.ts src/broadcast/dj-agent.ts src/routes/request.ts && npx tsc --noEmit` - 0 errors
* Will verify with a live stream after deployment by requesting a song that is already queued and confirming that the system selects an alternative track or skips the duplicate

## Notes for Reviewers

Three small commits across three files (+21 lines total):

1. **`queue.ts`**

   * Adds a new `queuedIds()` accessor that returns track IDs from both `upcoming` and `current`
   * Adds a deduplication guard in `push()` that returns `-1` when a duplicate track is detected
   * Existing fire-and-forget callers remain unaffected because they already ignore the return value

2. **`dj-agent.ts`**

   * Merges `queuedIds()` into `recentIds` in both `pickViaAgent()` and `runRequestViaAgent()`
   * Causes the picker and agent to treat queued tracks the same way as recently played tracks and prefer alternatives
   * `sampleWithRecentFallback()` still falls back to the full pool if all candidates are filtered out

3. **`request.ts`**

   * Applies the same `queuedIds()` merge in the "more like this" and stateless cascade paths

Deduplication is based solely on the Subsonic track ID, not title or artist, to avoid false positives for remixes, alternate releases, or live versions.

Returning `-1` from `push()` avoids breaking existing callers that do not check the return value while still allowing future callers to detect duplicate rejections if needed.

### DIFF
--- a/controller/src/broadcast/dj-agent.ts
+++ b/controller/src/broadcast/dj-agent.ts
@@ -342,6 +342,7 @@ async function pickViaAgent(queue, { wantLink, audioWaypoint = null }: { wantLin
   // catalogues keep the long anti-repeat guard, while small-artist libraries
   // do not exclude every real candidate before the picker sees it.
   const { ids: recentIds, keys: recentKeys } = queue.recentlyPlayed(windows.trackHours);
+  for (const id of queue.queuedIds()) recentIds.add(id);
   const recentArtists = queue.recentArtistsSince(windows.artistHours);
 
   const { object, steps, toolCalls, extras } = await pickerAgent.run({
@@ -521,6 +522,7 @@ async function runRequestViaAgent(queue: any, { requester }: { requester: string
     // song from earlier in the day. 2h covers the "don't repeat the song still
     // ringing in their ears" case and nothing more.
     const recentIds = queue.recentlyPlayedIds(2);
+    for (const id of queue.queuedIds()) recentIds.add(id);
 
     const { object, toolCalls, extras } = await requestAgent.run({
       messages: session.windowMessages(),
--- a/controller/src/broadcast/queue.ts
+++ b/controller/src/broadcast/queue.ts
@@ -285,7 +285,7 @@ class Queue {
     introKind?: string;
     aiPicked?: boolean;
   }) {
-    if (track?.id) {
+    if (aiPicked && track?.id) {
       const dominated = this.upcoming.some(i => i.track?.id === track.id)
         || (this.current?.track?.id === track.id);
       if (dominated) {
--- a/controller/src/broadcast/queue.ts
+++ b/controller/src/broadcast/queue.ts
@@ -285,6 +285,14 @@ class Queue {
     introKind?: string;
     aiPicked?: boolean;
   }) {
+    if (track?.id) {
+      const dominated = this.upcoming.some(i => i.track?.id === track.id)
+        || (this.current?.track?.id === track.id);
+      if (dominated) {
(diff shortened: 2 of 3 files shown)


perminder-klair:
Nice catch on the root cause, the missing dedup in `push()` is real and the ID-based approach is the right call.

One thing holds me back though. The guard returns `-1` and drops the track, but none of the callers check it. So on an explicit request the route still generates the intro, appends the session turn, and replies "coming up" to the listener even though nothing got queued. That happens most on exactly the low-coverage library from #537, since the `recentIds` fallback re-picks a queued track and the guard then eats it. The duplicate is gone, but now the request silently fails and tells the user it worked. `/dj/queue-track` has the same issue, it returns `ok: true` with `queuePosition: -1`.

The `recentIds` merges are good, keep those. For the guard I'd suggest one of:

1. Scope it to autonomous picks (apply only when `aiPicked`). That kills the stacking from #537, which is all agent/fallback convergence, and never swallows a human request. Smallest change.
2. Keep the broad guard but have callers check for `-1` and reply honestly ("that's already coming up" / "playing right now").

I lean towards option 1 as the merge bar. Happy to look again once that's in.

geekylakshya:
Good call. I've added a fourth commit that scopes the deduplication guard to autonomous picks only:

```ts
if (aiPicked && track?.id) {
```

Human requests now pass through `push()` unconditionally, so a listener will never receive a false "coming up" confirmation for a track that was silently dropped as a duplicate.

The `recentIds` merges remain unchanged. They only influence the picker's candidate pool and help steer autonomous selections away from tracks that are already queued, without affecting explicit user requests.

### Validation

* ESLint: 0 errors
* `tsc --noEmit`: clean

PR #539, opened 2026-06-23
feat: add Requesty as an OpenAI-compatible LLM provider

Thibaultjaigu:
## What

Adds Requesty as a selectable, OpenAI-compatible LLM provider, mirroring the existing OpenRouter provider. Requesty (https://requesty.ai) is an OpenAI-compatible gateway that routes one API across many models using the same `provider/model` naming as OpenRouter, so it reuses the existing OpenAI-compatible client path with a fixed default base URL (`https://router.requesty.ai/v1`) and `REQUESTY_API_KEY`.

## Changes (mirrors OpenRouter at each wiring site)

**controller**
- `src/llm/internal/provider/registry.ts` — `DEFAULT_REQUESTY_BASE_URL` + a `case 'requesty'` building the OpenAI-compatible client at `router.requesty.ai/v1` (pass-through semantics matching openrouter).
- `src/llm/provider.ts` — re-export `DEFAULT_REQUESTY_BASE_URL`.
- `src/llm/internal/provider/capabilities.ts` — requesty capability descriptor.
- `src/llm/internal/provider/embedding.ts` — requesty default embedding model + builder case.
- `src/settings.ts` — `requesty` added to `LLM_PROVIDERS` and `EMBEDDING_PROVIDERS`.
- `src/routes/onboarding.ts` — `case 'requesty'` in the test-LLM switch.
- `src/routes/settings.ts` — `REQUESTY_API_KEY` presence check.
- `src/setup/secrets.ts` — `REQUESTY_API_KEY` in the secret allow-list.
- `scripts/picker-test.mjs` — requesty in the provider usage string.

**web**
- `components/onboarding/steps.tsx` — `{ id: 'requesty', label: 'Requesty' }` in `LLM_PROVIDERS`.
- `components/onboarding/useWizard.ts` — `requesty → REQUESTY_API_KEY` in the key-stitch map.
- `components/admin/SettingsPanel.tsx` — env-var map, provider labels, embed-model suggestions, model-id hint, embed-providers list.
- `components/what/YourStack.tsx` — `Requesty` added to the displayed stack list.

**cli**
- `src/commands/setup.ts` — `CloudProvider` type, `CLOUD_ENV_VAR`, provider select list, example model, probe wiring.
- `src/probes.ts` — `probeRequesty` (GET `router.requesty.ai/v1/models`).
- `src/util.ts` — `REQUESTY_API_KEY` in the wizard key allow-list.
- `.env.example` + `cli/src/assets.generated.ts` — `# REQUESTY_API_KEY=` (kept in sync).

## Testing

- `controller`: `tsc --noEmit` clean (exit 0); `npm run test:llm` (llm-pure picker suite) passes.
- `web`: `tsc --noEmit` clean (0 errors).
- `cli`: `tsc --noEmit` clean (exit 0).
- Live: `POST https://router.requesty.ai/v1/chat/completions` with `model: openai/gpt-4o-mini` returned HTTP 200 with a real completion.

Docs: https://docs.requesty.ai • Keys: https://app.requesty.ai/api-keys

I work at Requesty. This mirrors the existing OpenRouter provider as closely as possible. Happy to adjust or close it if it's not a fit.


perminder-klair:
Thanks for this, it's a tidy PR and tracks the OpenRouter wiring closely. One thing will stop it working as shipped though.

The key never reaches the client. Cloud keys in this project don't live in settings.json. Both the onboarding wizard and the admin UI write them to `state/secrets.env`, so they end up in `process.env` and `settings.llm.apiKey` stays empty by design. Every other cloud provider works because its SDK reads its own env var, OpenRouter included (`createOpenRouter` picks up `OPENROUTER_API_KEY`).

The Requesty case uses `createOpenAI({ apiKey: cfg.apiKey || 'unused' })`. `createOpenAI` only ever reads `OPENAI_API_KEY`, never `REQUESTY_API_KEY`, and since `'unused'` is passed explicitly it won't fall back to the env at all. So with the key sitting in secrets.env and `settings.apiKey` empty, every request goes out with a bogus token and 401s.

It's easy to miss because the onboarding Test button passes. That path forwards the key you typed straight into the test client, so you see green, save, and then live picks quietly 401 and fall back.

Two ways to fix it:

1. Use the official provider, `@requesty/ai-sdk` (`createRequesty`). It reads `REQUESTY_API_KEY` from the env by default at the same base URL, so it becomes a true mirror of the OpenRouter case. Costs a dependency, and embeddings would still go through `createOpenAI` since the dedicated provider doesn't cover them.

2. Keep `createOpenAI` and read the env yourself: `cfg.apiKey || process.env.REQUESTY_API_KEY || 'unused'`, in both the chat case (`registry.ts`) and the embedding case (`embedding.ts`). Smallest change, no new dependency, keeps `debugFetch` and one transport for chat and embeddings.

I'd lean towards 2 for this codebase, but either is fine.

Minor: a few display/docs spots that name OpenRouter weren't updated (`Prerequisites.tsx`, `ManualInstall.tsx`, `ModelsAndTokens.tsx`, the cloud embedding hint in `music/embeddings.ts`, README). Not blockers.

Thibaultjaigu:
Thanks for the precise diagnosis — you're exactly right. Since cloud keys live in `process.env` (from `state/secrets.env`) and `createOpenAI` only reads `OPENAI_API_KEY`, the explicit `'unused'` meant Requesty requests went out with a bogus token.

I went with your option 2 (smallest change, no new dependency): `cfg.apiKey || process.env.REQUESTY_API_KEY || 'unused'`, applied to both the chat case (`registry.ts`) and the embedding case (`embedding.ts`), keeping `debugFetch` and the single transport for both. Pushed; controller `tsc --noEmit` is clean.

I left the OpenRouter embedding branch as-is to keep the diff scoped to Requesty — happy to fold the same env fallback into it in this PR if you'd prefer the two stay symmetric. I can also tidy the display/docs spots that still say OpenRouter (`Prerequisites.tsx`, `ManualInstall.tsx`, `ModelsAndTokens.tsx`, `music/embeddings.ts`, README) — just say the word.

perminder-klair:
thank you @Thibaultjaigu

PR #548, opened 2026-06-23
fix(dj-agent): don't record a phantom pick when the dedup guard drops a track

### LINKED ISSUE #537: Bug: queue.push() allows duplicate tracks to stack in the queue

### LINKED ISSUE #547: DJ session records a phantom pick when the dedup guard drops an autonomous track
Follow-up to #538 (which fixed #537).

## Background

#538 added an ID-based dedup guard to `queue.push()`, scoped to autonomous picks (`aiPicked`). When the picker selects a track already in `upcoming`/`current`, `push()` logs `dedup-skip` and returns `-1` instead of queuing it.

Since `runTrackEvent` only fires when `upcoming` is empty, the guard only ever collides with `current` — i.e. the picker re-picked the currently-playing track. `recentlyPlayed()` already includes `current` (`controller/src/broadcast/queue.ts:692-694`), so this happens in two cases:
1. **Degenerate tiny library** — every candidate is filtered out, so `sampleWithRecentFallback()` re-includes recents (incl. `current`) and the model picks it. (This is #537's scenario.)
2. **LLM misbehaves** — healthy pool with fresh candidates, but the model picks `current` anyway.

## The wart

`enqueuePick()` ignores the `-1` return, so `pickViaAgent` still runs `session.appendTurn({... "Selected X" })` (`controller/src/broadcast/dj-agent.ts:377-385`) even though nothing was queued. The DJ session then records a **phantom pick** for a track that never airs — the same class of inconsistency the maintainer flagged for the listener-facing request path in #538, just internal to the session window.

The slot itself is covered: with nothing in `upcoming`, Liquidsoap falls to `auto.m3u` and the next track-start re-triggers `runTrackEvent`. So this is not user-facing breakage — just a session-history inconsistency (and, in case 2, an avoidable lost AI-pick slot).

## Proposed fix (small, no retry loop)

Reuse the existing pool-fallback wiring instead of adding a retry:
- `enqueuePick()` returns the `push()` result.
- On `-1`, `pickViaAgent` skips the `session.appendTurn` and returns `false` → `runTrackEvent` falls through to `pickViaPool` (recovers the slot in case 2).
- `pickViaPool` likewise skips its phantom turn on `-1` and lets `auto.m3u` backstop (case 1 — no further retry, since the collapsed pool would just re-pick `current`).

A naive retry loop is deliberately avoided: in the degenerate tiny-library case there is no fresh track to pick, so `auto.m3u` is the correct backstop.

## Scope

Files: `controller/src/broadcast/dj-agent.ts`, `controller/src/broadcast/queue.ts`. ~12 lines. Low priority — behaviour is already strictly better than the pre-#538 bug.

perminder-klair:
## What

Closes #547 (follow-up to #538, which fixed #537).

When #538's dedup guard drops an autonomous (`aiPicked`) pick — `queue.push()` returns `-1` because the picked track is already queued/on-air — `enqueuePick()` ignored the result. So `pickViaAgent`/`pickViaPool` still logged an `ai-pick`, wrote a durable picks-log record, and appended a `"Selected X"` turn to the DJ session for a track that **never aired**. The playback slot itself was fine (Liquidsoap falls to `auto.m3u`), but the session history recorded a phantom pick.

## How

- **`enqueuePick()`** now pushes first and returns the queue position (or `-1`). On a drop it skips the `ai-pick` log **and** `recordPick`, so neither reports a no-op (`push()` already logs `dedup-skip`).
- **`pickViaAgent`** returns `false` on a drop and skips its session turn. `runTrackEvent` then falls through to the pool for a fresh pick — the agent itself is healthy, so the breaker is **not** tripped (`breakerSuccess()` still fires).
- **`pickViaPool`** skips its session turn on a drop and lets `auto.m3u` backstop; the next track-start re-triggers `runTrackEvent`.

## Why no retry loop

The guard only ever collides with the currently-playing track (`runTrackEvent` fires with an empty `upcoming`), and `recentlyPlayed()` already excludes `current`. A drop therefore means either (1) a tiny library whose candidate pool collapsed to recents, or (2) the LLM ignored guidance. Case 2 is recovered by the pool fall-through; case 1 has no fresh track to pick, so `auto.m3u` is the correct backstop. A retry would just re-hit the collapsed pool.

`windowMessages()` already filters stale pick events and coalesces same-role turns, so even a rare agent+pool double-dedup leaves the session window well-formed — no user-message-coalescing hazard.

## Testing

- `cd controller && npx eslint . && npx tsc --noEmit` — 0 errors (pre-existing `no-explicit-any` warnings only), tsc clean.

### DIFF
--- a/controller/src/broadcast/dj-agent.ts
+++ b/controller/src/broadcast/dj-agent.ts
@@ -318,10 +318,13 @@ function trackFields(song) {
 // playing. It's attached to the queued item so the queue airs it at the
 // transition INTO this track (queue.airIntro), not over whatever is currently
 // on-air when the pick is made — which is one track earlier (issue #189).
-async function enqueuePick(queue, song, reason, source, link: string | null = null) {
-  queue.log('ai-pick', `${song.title} — ${song.artist}`, { reason, source });
-  recordPick({ song, reason, source });
-  await queue.push({
+// Returns the queue position, or -1 when push()'s dedup guard dropped the pick
+// because that track is already queued/on-air. On a drop we skip the ai-pick log
+// AND the durable picks-log record so neither reports a phantom pick that never
+// aired (push() has already logged the dedup-skip). Callers fall back on -1
+// (agent → pool → auto.m3u) instead of recording a session turn for a no-op.
+async function enqueuePick(queue, song, reason, source, link: string | null = null): Promise<number> {
+  const pos = await queue.push({
     track: trackFields(song),
     requestedBy: null,
     intent: reason || 'ai pick',
@@ -329,13 +332,17 @@ async function enqueuePick(queue, song, reason, source, link: string | null = nu
     introKind: 'link',
     aiPicked: true,
   });
+  if (pos === -1) return -1;
+  queue.log('ai-pick', `${song.title} — ${song.artist}`, { reason, source });
+  recordPick({ song, reason, source });
+  return pos;
 }
 
 // ---------------------------------------------------------------------------
 // Track event — a track started; pick the next one and maybe air a link.
 // ---------------------------------------------------------------------------
 
-async function pickViaAgent(queue, { wantLink, audioWaypoint = null }: { wantLink: boolean; audioWaypoint?: number[] | null }) {
+async function pickViaAgent(queue, { wantLink, audioWaypoint = null }: { wantLink: boolean; audioWaypoint?: number[] | null }): Promise<boolean> {
   await library.load();
   const windows = recencyWindowsForLibrary(library.stats().distinctArtists);
   // Scale the recency windows to the tagged library's artist diversity: dense
@@ -375,7 +382,11 @@ async function pickViaAgent(queue, { wantLink, audioWaypoint = null }: { wantLin
   const say = (djMode && rawSay) ? dj.enforceIntroBudget(rawSay, introMsOf(song)) : rawSay;
   // Attach the link to the pick so it airs as the pick starts (back-announcing
   // the track on-air now), instead of immediately over that on-air track (#189).
-  await enqueuePick(queue, song, object.reason, 'agent', (wantLink && say) ? say : null);
+  const queued = await enqueuePick(queue, song, object.reason, 'agent', (wantLink && say) ? say : null);
+  // Pick was already queued/on-air and got deduped — don't record a session turn
+  // for a track that never airs. Returning false lets runTrackEvent fall through
+  // to the pool for a fresh pick.
+  if (queued === -1) return false;
   session.appendTurn({
     role: 'dj', kind: 'pick',
     text: object.reason || `Selected "${song.title}".`,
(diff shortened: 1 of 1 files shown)


perminder-klair:
## Smoke test ✅

Ran a live dev-stack smoke test from this branch (`docker-compose.dev.yml`, controller source baked into the image), plus an isolated contract test.

**Live stack**
- Boots clean — controller `{"status":"on-air"}`, broadcast healthy, stream live.
- No runtime errors from this change. (The only log error was environmental — the LLM endpoint was unreachable in the test box.)
- The refactored control flow ran for real: `runTrackEvent` → `pickViaAgent` failed (LLM down) → **fell through to `pickViaPool`** → pool built a 15-candidate set → playback continued. The agent→pool fall-through this PR touches is intact.

**Dedup contract** (isolated, throwaway `STATE_DIR`, real `queue` module):
```
[queued]      Track A                    # first autonomous push → queued
[dedup-skip]  Track A (already queued)   # autonomous duplicate → push() returns -1 ✅
[queued]      Track B                    # fresh autonomous → queued
[queued]      Track B                    # human (non-aiPicked) re-request → passes through, NOT deduped ✅
```
So `push()` returns `-1` on an autonomous duplicate (which `enqueuePick`/`pickViaAgent`/`pickViaPool` now handle by skipping the phantom session turn) and lets human re-requests through.

**Not triggered in a single live event:** the full chain *agent returns a deduped pick → session turn skipped → fall-through*, since it needs the LLM up **and** a degenerate collision (picker re-picking the currently-playing track). Each link is verified independently though — `push()→-1` proven live, the `queued === -1` guard branches are additive + tsc-checked, and the fall-through is proven live.

**Static gates:** `eslint .` 0 errors, `tsc --noEmit` clean, `llm-pure` + `picker-recency-regression` pass. CI green (lint controller + web).

PR #554, opened 2026-06-23
fix(app): drop unused BottomSheetModalProvider that killed touch on Android (New Arch)

### LINKED ISSUE #458: Android App, cannot connect.

perminder-klair:
## Problem

On the native Android player, the UI rendered and kept live-updating (now-playing, DJ lines) but was **completely dead to touch** — no control responded. It was **device-specific and independent of OS version and GPU**: dead on Pixel 7 Pro, Pixel 10, and Galaxy S22 Ultra; fine on Pixel 8a, Pixel 9 Pro XL, and OnePlus Pad 3 — same app build, same Android version. No crash, no ANR. (#458)

## Root cause

`BottomSheetModalProvider` (`@gorhom/bottom-sheet`) wrapped the entire app in `_layout.tsx`. On the **New Architecture** it leaves an Android touch-interceptor over the whole tree on some devices, swallowing every tap before it reaches the controls while the JS thread keeps running (hence: renders fine, updates fine, no crash). Whether it triggers depends on device/ROM timing, which is why it tracked neither the Android version nor the GPU vendor.

Critically, the app's **only** sheet (the Themes picker) uses gorhom's **non-modal** `<BottomSheet>`, which does **not** require `BottomSheetModalProvider`. The provider was dead weight that only added the touch hazard.

## Fix

Remove the unused `BottomSheetModalProvider`. `GestureHandlerRootView` stays (the non-modal sheet's pan-to-dismiss needs it) and the Themes sheet is unchanged — so there's **no functional or visual loss**.

## How it was found (bisection on real devices)

Blur, the Skia visualiser, and the nav transition were each ruled out by on-device test builds. Then:

- A build with the **whole** gesture-handler + bottom-sheet stack removed → **touch restored** on an affected Pixel 7 Pro (confirmed by @R0ckGam3r).
- This build, removing **only** `BottomSheetModalProvider` (sheet + gestures kept) → **confirmed working** by @R0ckGam3r and @diablo581 ("works on the first try"), Themes sheet included.

## Testing

- Device-confirmed on a previously-dead Pixel 7 Pro and another affected device.
- `tsc --noEmit` clean.

Fixes #458.

### DIFF
--- a/app/src/app/_layout.tsx
+++ b/app/src/app/_layout.tsx
@@ -19,7 +19,6 @@ import * as SplashScreen from 'expo-splash-screen';
 import { StatusBar } from 'expo-status-bar';
 import { useEffect } from 'react';
 import { GestureHandlerRootView } from 'react-native-gesture-handler';
-import { BottomSheetModalProvider } from '@gorhom/bottom-sheet';
 import { SafeAreaProvider } from 'react-native-safe-area-context';
 import ErrorScreen from '@/components/ErrorScreen';
 import { StationProvider, useStation } from '@/config/StationContext';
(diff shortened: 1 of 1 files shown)


perminder-klair:
Closing — on-device testing showed removing only BottomSheetModalProvider is NOT enough: the UI was still dead (and flaky enough to pass a first-try test before failing on relaunch). The touch interceptor is the wider gesture-handler + bottom-sheet stack, not just the provider.

The real fix is on `fix/android-dead-touch-rn-sheet`: rebuild the Themes sheet on React Native's core <Modal> and drop GestureHandlerRootView + BottomSheetModalProvider entirely (matching the stripped build that two testers confirmed). A new PR will follow once it's device-confirmed across multiple launches. See #458.

PR #571, opened 2026-06-24
Fix/gate empty embedding picker tools

Grafting1:

### DIFF
--- a/controller/src/llm/internal/tools/picker-tools.ts
+++ b/controller/src/llm/internal/tools/picker-tools.ts
@@ -228,11 +228,14 @@ export function buildPickerTools({
     }),
 
     // Only registered when the controller's own text/mood embedding index has
-    // been built (withEmbedding > 0) AND a text-embedding provider is live.
-    // When both are missing every call returns [] — and the old description
-    // said "Prefer this to similarSongs", actively steering the model into a
-    // dead tool. Gate it off entirely so the model never sees an unusable option.
-    ...(hasTextEmbeddings && hasEmbeddingProvider ? {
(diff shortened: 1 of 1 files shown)


perminder-klair:
Thanks for this, the gating approach is right and it solves a real problem. Audio embeddings are off by default so `tracksThatSoundLikeThis` was a dead tool on every default install, and I confirmed the change passes lint and typecheck locally.

Two things before merge:

**a) Please re-target `develop`, not `main`.** `main` is the release branch and only `develop` (plus the automated release-please PR) merges into it, so CI will block a PR aimed at `main`. Switch the base with the dropdown above.

**b) Consider dropping `&& hasEmbeddingProvider` from the two `tracksLikeThis` gates.** That tool resolves through `library.tracksLikeThis` -> `db.knnById`, which does KNN over the seed track's stored vector and never calls the embedding provider at query time. So it works whenever the index exists, even if the provider is later turned off. Gating it on `hasTextEmbeddings` alone (the way `tracksThatSoundLikeThis` already gates on `hasAudioEmbeddings`) would be more precise. `searchByLyrics` does embed the query, so keep the provider check there.

Neither is a blocker, b is just a nice-to-have for precision.

Grafting1:
Thanks for the quick review! Both addressed:

- Re-targeted to `develop`.
- Dropped `&& hasEmbeddingProvider` on `tracksLikeThis` — agreed, it KNNs over the stored vector via `db.knnById` and never embeds at query time, so it works whenever the index exists; now gates on `hasTextEmbeddings` alone (mirroring `tracksThatSoundLikeThis`). Kept the provider check on `searchByLyrics` since that one embeds the query.

PR #573, opened 2026-06-24
Feat/api key management UI

### LINKED ISSUE #572: feat(settings): add API key management UI for all providers
## Problem

Adding or rotating API keys currently requires SSH access and manual edits to `state/secrets.env`.

This creates unnecessary friction for operators, especially since nearly all other configuration is already managed through the admin UI.

Examples include:

* `ANTHROPIC_API_KEY`
* `OPENAI_API_KEY`
* `OPENROUTER_API_KEY`
* `ELEVENLABS_API_KEY`
* `EMBEDDING_API_KEY`
* Other keys listed in `SECRET_ENV_KEYS`

## Proposed Solution

Add an **API Keys** card to the Settings panel that exposes all entries from `SECRET_ENV_KEYS`.

The existing scrobbling UI (`ScrobbleSection`) already implements the required UX pattern and can serve as the reference implementation.

## Behaviour

Match the existing scrobbling pattern exactly:

* One `type="password"` input per key
* Placeholder displays:

  * `•••••• (on file)` when a key exists
  * A provider-specific hint (for example, `sk-ant-...`) when a key is not configured
* On load:

  * The controller returns presence information only
  * Secret values are never returned to the client
* On save:

  * Blank field = no change
  * Non-empty field = update the stored secret
  * Updates merge into existing secrets rather than replacing the entire file
  * Changes take effect immediately in-process without requiring a restart
* Display a `KeyStatus` badge beside each field indicating whether the key is currently set or missing

## Benefits

* Removes the need for SSH access for routine key management
* Makes API key rotation significantly easier
* Reuses an existing, proven UI pattern
* Preserves current security behavior by never exposing stored secret values
* Keeps all provider credentials manageable from the same admin interface

geekylakshya:
## What
  Adds inline API key management to the admin Settings panel — operators can paste provider keys directly in the LLM, TTS, and Embedding sections instead of SSH-editing
  `state/secrets.env`.

  Closes #572

  ## Why
  Adding or rotating API keys required SSH access and manual file edits. Every other setting is manageable from the admin UI; keys should be too.

  ## How tested
  - Deployed to a test LXC with rebuilt controller + web images
  - Verified all 13 SECRET_ENV_KEYS hot-reload after save (no restart needed)
  - Verified Anthropic direct provider works end-to-end with key saved from UI
  - Verified embedding probe passes with OpenRouter key saved from UI
  - Verified KeyStatus badges update after save without page reload
  - Verified keys with special characters (%/=/space) save correctly after the quote-merge fix
  - Full backend audit of all 13 key paths: 12 PASS, 1 was FAIL (SEARCH_API_KEY frozen at boot) and fixed in this PR
  - Lint clean on both controller and web

  ## Notes for reviewers
  - `POST /settings/secrets` is the new endpoint (admin-gated, merges into `state/secrets.env`, values never returned to client)
  - Key inputs appear contextually: select Anthropic in LLM section and the ANTHROPIC_API_KEY field appears inline. Switch to Ollama and it disappears. Same pattern for TTS
  cloud and embedding.
  - `saveSecrets()` in `setup/secrets.ts` had a quote-merge bug: it read existing values raw (with quotes) then `envEscape` choked on the `'`. Fixed by stripping quotes on
  read, same as `loadSecretsIntoEnv` already does.
  - `web-search.ts` read `SEARCH_API_KEY` from a boot-frozen `config.search.apiKey` instead of live `process.env`. Fixed to read env at call time so runtime key saves take
  effect without restart.
  - Known limitation: `envEscape` rejects values containing a single quote (`'`) with a 400. Rare in real API keys but worth noting.


perminder-klair:
Reviewed the diff and traced the secrets path, admin gate, CORS, and logging. This looks good to merge. Write-only design (GET only exposes booleans, password inputs, keys never returned), whitelist enforced in both the route and saveSecrets, admin-gated, and secrets.env is not shell-sourced anywhere so there is no injection vector. Nice catches on the quote-merge and SEARCH_API_KEY boot-freeze bugs too.

A few non-blocking nits, none of them blockers:

1. Error leakage: `res.status(400).json({ error: err.message })` can surface filesystem paths to the client. It is behind the admin gate so impact is low, but a generic message to the client with the detail logged server-side would be cleaner.
2. No length or shape validation on values. An admin could write a large blob up to the 600kb body limit. Negligible since it is admin-gated, but a small sanity cap would not hurt.
3. Shared-key surprise: `OPENAI_API_KEY` is one env var across LLM, cloud TTS, and the embedding fallback. Saving it in the TTS section overwrites it for LLM too. Technically correct since it is the same key, but might surprise an operator who expected per-section keys. Only `EMBEDDING_API_KEY` is decoupled.
4. Minor future note: secrets.env is not shell-sourced today, but the file comment invites hand-editing and calls it "shell-style". If anything ever sources it, a crafted value could inject, so worth a guard if that ever changes.

PR #575, opened 2026-06-24
Feat/api key test button

### LINKED ISSUE #574: feat(settings): "Test" button to verify API key before saving
## Summary

  Add a small "Test" button beside each API key input in the Settings panel so operators can verify a key works before (or after) saving it.

  ## Motivation

  Currently there's no feedback when a key is wrong - the first sign of failure is a broken LLM/TTS/embedding call at runtime.

  ## Proposed implementation

  ### Backend

  New endpoint: `POST /settings/secrets/test`

  Request: `{ key: string, value: string }` (the env var name + value to test)
  Response: `{ ok: boolean, message: string, latencyMs: number }`

  Probe strategy per key:
  | Key | Probe |
  |-----|-------|
  | `OLLAMA_BASE_URL` | `GET /api/version` |
  | `OPENAI_COMPATIBLE_BASE_URL` | `GET /models` |
  | `ANTHROPIC_API_KEY` / `OPENAI_API_KEY` / `OPENROUTER_API_KEY` / `GEMINI_API_KEY` | 1-token `generateText` call |
  | `ELEVENLABS_API_KEY` | `GET /v1/user` |
  | `VOYAGE_API_KEY` / `OPENAI_EMBEDDING_API_KEY` etc. | reuse existing `probeEmbeddingConfig()` |

  ### Frontend

  - Small `Btn` (sm variant) labelled "Test" beside each key input, same pattern
    as the test button in `ScrobbleSection`
  - Shows spinner while in-flight
  - On success: green inline text with latency (e.g. `✓ 142 ms`)
  - On failure: red inline text with message from response

  ## Open design question

  Should "Test" probe the **typed (unsaved) value** in the input, or only the **already-saved key** in `secrets.env`?

  - Typed value: more useful (test before committing), but means the key is sent to the backend without being saved
  - Saved only: simpler, but less useful if the user is entering a new key

  Suggest: test the typed value (send in request body, don't persist).

geekylakshya:
## What

  Adds a "Test key" button beside every cloud API key input in the admin Settings panel. Operators can verify a key works against its provider before or after saving it.

  Closes #574

  ## Why

  There was no validation feedback for API keys. The first sign of a bad key was a broken LLM/TTS/search call at runtime. The inline inputs from  #573 made key entry easy, but operators still had no way to confirm a key works without triggering a real DJ call.

  ## How tested

  - Deployed to test LXC  with rebuilt controller + web images
  - Tested bad keys for Anthropic, ElevenLabs, Tavily: all return friendly error messages ("Key rejected", "Rate limited", "Timed out") instead of raw SDK stack traces
  - Tested valid Anthropic key: returns "Anthropic key valid, model responded" with latency
  - Verified Test button is disabled when input is empty
  - Verified test result clears when provider changes
  - Verified result box follows custom theme colours (var(--accent) / var(--danger))
  - Verified embedding section has no Test key button (existing "Test embeddings" covers it)
  - Verified endpoint returns 401 without admin auth
  - Controller and web lint clean (0 errors)

  ## Notes for reviewers

  - `POST /settings/secrets/test` is non-mutating: it builds a one-off provider client with the typed value, never writes to process.env or secrets.env
  - Each LLM probe uses a 1-token generateText call with a 15s AbortSignal timeout to prevent hangs
  - Tavily probe uses `Authorization: Bearer` header (matches the production call path in web-search.ts)
  - Embedding key test was removed from the frontend because LibrarySection already has "Test embeddings" which probes the full embedding pipeline
  - `briefLlmError()` distils raw AI SDK errors into one-line actionable messages (maps HTTP status codes and common error patterns to plain
  English)
  - AI Gateway key returns ok:true with an informational message since there is no standalone probe for it
  - KeyTestResult uses the same compact rounded-box style as the embedding probe result, themed via CSS variables

### DIFF
--- a/controller/src/routes/settings.ts
+++ b/controller/src/routes/settings.ts
@@ -262,7 +262,7 @@ async function probeKey(
       try {
         const model = activeModel('anthropic') || 'claude-haiku-4-5-20251001';
         const m = createAnthropic({ apiKey: value })(model);
-        const out = await generateText({ model: m, prompt: 'Reply with the single word OK.', maxOutputTokens: 8, abortSignal: AbortSignal.timeout(15000) });
+        await generateText({ model: m, prompt: 'Reply with the single word OK.', maxOutputTokens: 8, abortSignal: AbortSignal.timeout(15000) });
         return { ok: true, message: `✓ Anthropic key valid · model responded` };
       } catch (err) { return { ok: false, message: briefLlmError(err) }; }
     }
@@ -270,7 +270,7 @@ async function probeKey(
       try {
         const model = activeModel('openai') || 'gpt-4o-mini';
         const m = createOpenAI({ apiKey: value })(model);
-        const out = await generateText({ model: m, prompt: 'Reply with the single word OK.', maxOutputTokens: 8, abortSignal: AbortSignal.timeout(15000) });
+        await generateText({ model: m, prompt: 'Reply with the single word OK.', maxOutputTokens: 8, abortSignal: AbortSignal.timeout(15000) });
         return { ok: true, message: `✓ OpenAI key valid · model responded` };
       } catch (err) { return { ok: false, message: briefLlmError(err) }; }
     }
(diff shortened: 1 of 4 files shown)


perminder-klair:
thanks @geekylakshya

PR #595, opened 2026-06-25
Feat/OpenAI compat api key

### LINKED ISSUE #594: feat(admin): add API key field for openai-compatible LLM provider
# Problem

When `openai-compatible` is selected as the LLM provider (for example, a self-hosted proxy such as CLIProxyAPI), the Settings UI does not display an API key field. The Base URL and Model inputs are available, but there is no way to configure `llm.apiKey` from the admin panel.

The application already supports `llm.apiKey` for this provider. If the value is added manually to `state/settings.json`, it is persisted and used correctly. However, there is currently no UI for viewing, updating, or testing the key.

# Why

* Operators using OpenAI-compatible endpoints that require bearer authentication (CLIProxyAPI, LiteLLM, Azure OpenAI proxies, etc.) cannot configure credentials through the UI
* All other providers that support API keys (OpenAI, Anthropic, Google, DeepSeek, OpenRouter, Gateway) expose an inline API key field and test workflow
* The same limitation exists for the fallback LLM configuration when `openai-compatible` is selected

# Proposed Solution

Add an optional API key field to the `openai-compatible` provider section in `LlmSection` for both the primary and fallback LLM configurations.

The field should:

* Read from and save to `llm.apiKey` in `settings.json`
* Use a placeholder such as `Bearer token (optional)` since many self-hosted OpenAI-compatible servers do not require authentication
* Follow the existing API key UX used by other providers, including the Test button
* Store the value in `settings.json` rather than `secrets.env`, consistent with how `baseUrl` and `model` are currently handled for this provider

geekylakshya:
What

  Add a bearer token (API key) field and "Test connection" button to the admin Settings UI when openai-compatible is selected as the LLM provider, for both primary and
  fallback configurations.

  Why

  Operators using OpenAI-compatible endpoints that require bearer authentication (CLIProxyAPI, LiteLLM, Azure OpenAI proxies, etc.) had no way to configure credentials
  through the UI. The key had to be manually added to settings.json. Every other keyed provider already had an inline key field and test workflow.

  Closes #594

  How tested

  - Deployed to test LXC  with docker compose up -d --build
  - Switched LLM provider to openai-compatible in the admin UI
  - Verified bearer token field appears below Server base URL with correct placeholder
  - Verified "Test connection" button probes the endpoint live without saving
  - Verified key persists to settings.json on save (not secrets.env)
  - Verified placeholder shows "on file" after save + page reload
  - Verified switching provider clears the input and test result
  - Verified fallback section shows the same field when fallback provider is openai-compatible
  - npm run lint (controller + web) passes clean

  Notes for reviewers

  - The compat key is intentionally stored in settings.json (not secrets.env) -- consistent with how baseUrl and model already work for this provider. The existing
  LLM_ENV_VARS / SECRET_ENV_KEYS / saveKey() path is not touched.
  - The "Test connection" button is enabled when a base URL is set, even without a key -- many self-hosted servers don't require auth, and the probe confirms reachability
  either way.
  - KeyStatus component is not used (that's for env-var-backed keys). Instead the Input placeholder shows "on file" when a key is already saved, matching the redaction from
  getRedacted().

### DIFF
--- a/controller/src/routes/settings.ts
+++ b/controller/src/routes/settings.ts
@@ -443,7 +443,7 @@ router.post('/settings/llm/probe-compat', requireAdmin, async (req, res) => {
     const m = createOpenAI({
       apiKey: (typeof apiKey === 'string' && apiKey.trim()) ? apiKey.trim() : 'no-key',
       baseURL: baseUrl.trim().replace(/\/+$/, ''),
-    })(model.trim());
+    }).chat(model.trim());
     await generateText({
       model: m,
       prompt: 'Reply with the single word OK.',
--- a/controller/src/routes/settings.ts
+++ b/controller/src/routes/settings.ts
@@ -426,6 +426,37 @@ router.get('/settings/llm/discover', requireAdmin, async (req, res) => {
 });
 
 // ---------------------------------------------------------------------------
+// POST /settings/llm/probe-compat — live probe for an openai-compatible key.
+// Body: { apiKey: string, baseUrl: string, model: string }
+// Always 200s with { ok, message, latencyMs }. The key is NOT saved.
+// ---------------------------------------------------------------------------
+router.post('/settings/llm/probe-compat', requireAdmin, async (req, res) => {
+  const { apiKey, baseUrl, model } = req.body || {};
+  if (!baseUrl || typeof baseUrl !== 'string' || !baseUrl.trim()) {
+    return res.status(400).json({ ok: false, message: 'baseUrl is required', latencyMs: 0 });
+  }
+  if (!model || typeof model !== 'string' || !model.trim()) {
+    return res.status(400).json({ ok: false, message: 'model is required', latencyMs: 0 });
+  }
+  const t0 = Date.now();
+  try {
+    const m = createOpenAI({
+      apiKey: (typeof apiKey === 'string' && apiKey.trim()) ? apiKey.trim() : 'no-key',
+      baseURL: baseUrl.trim().replace(/\/+$/, ''),
+    })(model.trim());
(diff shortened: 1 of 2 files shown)


perminder-klair:
Nice work, this fills a real gap and the persistence/redaction handling is clean. One thing to fix before merge though.

The Test connection probe builds the model with `createOpenAI({...})(model)`, which on `@ai-sdk/openai` v3 targets the Responses API (`/v1/responses`). But the actual runtime path for `openai-compatible` uses `.chat(id)` on purpose, since these servers only speak `/v1/chat/completions`. So the button ends up testing a different endpoint than the DJ actually uses. For llama.cpp, vLLM, LM Studio, CLIProxyAPI and friends that 404 on `/v1/responses`, the test will report failure even when the config is correct.

Quick fix:

```ts
})(model.trim());  ->  }).chat(model.trim());
```

The OpenAI key probe above it is fine as-is, since real OpenAI does support Responses. Only the compat one needs `.chat()`.

Minor: `apiKey` got added to `LlmForm` / `LlmFallbackForm` but nothing reads it (the placeholder uses `data.values?.llm?.apiKey` and save uses `compatKeyInput`), so it can go.

Happy to approve once the `.chat()` change is in.

geekylakshya:
@perminder-klair Good catch on the .chat() path -- fixed in 00e749f.

  - createOpenAI({...})(model) changed to createOpenAI({...}).chat(model) so the probe targets /v1/chat/completions like the runtime does
  - Removed the unused apiKey from LlmForm / LlmFallbackForm interfaces and their init code

  Both lints clean. Ready for re-review.

PR #615, opened 2026-06-26
feat(admin): searchable model discovery dropdown for all providers

### LINKED ISSUE #598: feat(admin): model discovery dropdown for LLM providers
# Problem

Operators currently have to type model names manually into a free-text field for every LLM provider. This is error-prone - a typo such as `gemni-3-flash` can silently fail at runtime—and it increases cognitive load because operators must know the exact model ID required by the provider.

# Proposed Solution

Add model discovery support to all LLM provider configurations.

When a provider is selected, display a **"Discover Models"** button (or automatically fetch available models) that queries the provider's model-list endpoint and populates a `<Select>` dropdown.

If model discovery fails, or if the provider does not support model listing, fall back to the existing free-text input. Operators should also be able to switch to manual entry at any time.

## Discovery Sources

- **OpenAI-compatible / Locca**
  - `GET {baseUrl}/models`
  - Existing implementation: `GET /settings/llm/discover`

- **Ollama**
  - `GET {ollamaUrl}/api/tags`

- **Cloud Providers**
  - OpenAI
  - Anthropic
  - Google
  - DeepSeek
  - OpenRouter
  - Requesty
  - Gateway

  Query each provider's model-list API using the configured API key and populate the available models automatically.

## Fallback Behavior

- Keep the current free-text model field as a manual override.
- Allow operators to enter custom model IDs even when discovery is available.
- Gracefully handle discovery failures without blocking configuration.

# Scope

Apply model discovery to:

- Primary LLM model field
- Fallback LLM model field
- Embedding model field
- Cloud TTS model field

geekylakshya:
## What changed

Resolves #598

This PR adds **searchable model discovery dropdowns** across the entire admin Settings panel -- LLM primary, LLM fallback, TTS, and embedding model pickers -- replacing the plain text inputs with auto-populated, filterable lists.

### Model Discovery Engine

- **New `useModelDiscovery` hook** (`web/hooks/useModelDiscovery.ts`) -- fires on provider change, returns `{ models, loading, error, refresh }`
- **Unified `/settings/llm/models` route** (`controller/src/routes/settings.ts`) -- single endpoint that discovers models from any provider:
  - **Ollama/Locca**: hits `/api/tags` on the configured server
  - **OpenAI / DeepSeek / Anthropic / Google**: uses provider API with env key
  - **OpenRouter**: fetches from `openrouter.ai/api/v1/models`
  - **OpenAI-compatible**: probes `baseUrl/v1/models` with bearer token
- **Scoped embedding discovery** -- when `?scope=embedding` is passed:
  - **OpenRouter**: uses the undocumented `?output_modalities=embeddings` parameter to return only embedding models (previously the embedding section showed chat models because OpenRouter has no separate embedding endpoint)
  - **OpenAI/Google**: filters by regex (`embed|text-embedding`)
  - Ensures embedding pickers never show chat models

### Searchable ModelCombobox

- Built on **cmdk** (`Command` primitives) with `createPortal` to escape `overflow: hidden` card containers
- **Smart viewport positioning**: measures `getBoundingClientRect()` space below trigger -- if < 300px, opens **upward** with inverted layout (search bar at bottom, closest to trigger)
- Closes on outside click and page scroll (but not when scrolling inside the dropdown list itself)
- Inline refresh button shows on error or empty list

### API Key Testing Improvements

- **Test with key on file**: when the bearer token input is empty, the "Test" button sends an empty `apiKey` to the backend, which resolves the saved key from `settings.llm.apiKey` or `settings.llm.fallback.apiKey` depending on the target URL
- **Auto-save on success**: when a key test passes, the key is automatically persisted to `secrets.env`
- Works for both standard provider keys (`/settings/secrets/test`) and OpenAI-compatible endpoints (`/settings/llm/probe-compat`)

### UI / Lint

- Added `wrapperClassName` prop to `CommandInput` for inverted border styling
- Exempted `SettingsPanel.tsx` from `react/forbid-dom-props` inline style rule (portal absolute positioning requires it)
- Moved TTS "API key found in environment" status alert below the voice level trim bar

### DIFF
--- a/controller/src/routes/settings.ts
+++ b/controller/src/routes/settings.ts
@@ -16,7 +16,7 @@ import { restartLiquidsoap, startStream, stopStream, streamStatus } from '../bro
 import { invalidateWeatherCache } from '../context.js';
 import { requireAdmin } from '../middleware/auth.js';
 import { saveSecrets, SECRET_ENV_KEYS } from '../setup/secrets.js';
-import { generateText } from 'ai';
+import { generateText, createGateway } from 'ai';
 import { createAnthropic } from '@ai-sdk/anthropic';
 import { createOpenAI } from '@ai-sdk/openai';
 import { createGoogleGenerativeAI } from '@ai-sdk/google';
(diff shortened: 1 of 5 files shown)


perminder-klair:
beautiful work @geekylakshya thank you!

PR #617, opened 2026-06-26
fix(picker): let agent identify songs from pasted lyrics via web search

### LINKED ISSUE #616: Bug: Lyrics-based song requests resolve to wrong track - agent skips web search
## Bug

When a listener pastes song lyrics into the request field on `/listen`, the DJ agent resolves the request to the wrong track.

The `identifyRequestedTrack` tool (web search → identify song → match to library) is available and can correctly identify the song from lyrics, but the agent never calls it because of the current tool description:

> "Do NOT use when the listener already gives an artist or a title. Use `searchLibrary` for that."

The agent interprets a block of lyrics as user-provided text rather than something that needs to be resolved. As a result, it falls back to `searchLibrary`, which performs a naive Navidrome text match on lyric fragments and often returns unrelated songs.

## Root cause

The `identifyRequestedTrack` tool description in:

```text
controller/src/llm/internal/tools/picker-tools.ts
```

labels the tool as a **last resort** and explicitly says not to use it when the listener provides an artist or title.

Lyrics are neither a description nor an artist/title, leaving them as an unhandled input type. The model follows the instruction and calls `searchLibrary` instead, causing lyric fragments to be matched against Navidrome song titles.

For the same lyrics, SearXNG correctly identifies the song (for example, returning **Jugni** by **Diljit Dosanjh** from hinditracks.in), but that resolution path is never used.

## Fix

Updated the `identifyRequestedTrack` tool description to explicitly include **lyrics** as a trigger for using the tool.

Verified with `gemini-3.1-flash-lite`:

* The agent calls `identifyRequestedTrack`.
* The lyrics are searched on the web.
* The correct song is identified (for example, **Jugni** by **Diljit Dosanjh**).
* The resolved song is matched against the library.
* The correct track is queued instead of an unrelated title.

geekylakshya:
## What

Updated the `identifyRequestedTrack` tool description to explicitly include **song lyrics** as a trigger case, so the agent uses web search to identify the song instead of naively text-matching lyric fragments with `searchLibrary`.

## Why

When a listener pastes song lyrics into the request field, the agent skips `identifyRequestedTrack` because the previous tool description described it as a **last resort** and instructed the model not to use it when the listener already provides an artist or title.

Lyrics are neither a description nor an artist/title, leaving them as an unhandled input type. As a result, the agent falls through to `searchLibrary`, which matches arbitrary lyric fragments against Navidrome titles and frequently queues the wrong track. For example, the lyric fragment **"Billi akh"** from *Jugni* matched **"Billi Billi"** by Sukhbir instead of identifying the requested song.

Resolves #616.

## How tested

* Sent the same Punjabi lyrics request 3 times before the fix and received 3 different incorrect tracks:

  * Billi Billi
  * Gal Sun
  * Lemonade
  
* Rebuilt the controller with the updated tool description.
* Sent the same lyrics request again.
* Confirmed the agent called `identifyRequestedTrack`.
* Verified SearXNG identified **Jugni** by **Diljit Dosanjh**, matched it in the library, and queued the correct track.
* Tested with `gemini-3.1-flash-lite` to confirm even the smallest model in rotation follows the updated guidance.

## Notes for reviewers

This change updates a single tool description string only.

There are **no logic, schema, or control flow changes**.

The key behavior change is replacing the previous "LAST RESORT" / "Do NOT use when the listener already gives an artist or a title" guidance with explicit instructions to use `identifyRequestedTrack` when the request appears to contain song lyrics, such as repeated phrases, verse-like structure, or lyrics in another language.


perminder-klair:
thank you @geekylakshya !

PR #637, opened 2026-06-26
fix(queue): replace _autoMisses heuristic with Liquidsoap telnet sync…

### LINKED ISSUE #632: Bug : Queue items vanish from UI but still play - _autoMisses wipe clears upcoming while   tracks remain in Liquidsoap
## Bug

Songs queued via listener requests or AI picks disappear from the queue section of the web UI, but still play in the correct order. The queue UI appears empty, then sometimes briefly shows items again after a skip, then goes empty again.

## Background

The controller maintains its own `upcoming` queue for the web UI while Liquidsoap maintains the actual playback queue (`dj_queue` / `next.txt`).

To recover from stale controller state after a restart, `onTrackStarted()` tracks consecutive failures to match the currently playing track against `upcoming`. After three consecutive misses, it assumes the controller's queue is stale and clears it.

That recovery logic is valid after a full restart, but it is also triggered during normal playback.

## Root cause

`onTrackStarted()` in `queue.ts` (~line 554) matches incoming `now-playing.json` metadata against upcoming items by `subsonic_id`, falling back to exact `title + artist` string comparison.

When the match fails, `_autoMisses` increments. After three consecutive misses with items still in `upcoming`, the array is cleared:

```ts
this._autoMisses++;

if (this._autoMisses >= 3 && this.upcoming.length > 0) {
  this.upcoming = [];
}
```

The items are already in Liquidsoap's `dj_queue` because they were written to `next.txt` by `drainToLiquidsoap()` when they were queued.

As a result:

* Playback continues in the correct order.
* `upcoming` becomes empty.
* `/state` returns an empty queue.
* The web UI shows no upcoming tracks.

The controller and Liquidsoap become out of sync.

## Match failures can come from

* A jingle playing between queued tracks (no `subsonic_id`, no title match against a music track)
* Artist string variance between the annotated URI and what Liquidsoap reports back via `on_metadata` (for example `feat.` vs `featuring`, Unicode dashes, trailing whitespace)
* A crossfade transition firing `on_metadata` for the outgoing track (the pre-cross hook placement should prevent this, but edge cases exist)

The threshold of three was chosen to tolerate an interleaved jingle without clearing a genuine pending pick, but in practice a jingle, one artist string mismatch, and one more jingle (or another mismatch) reaches three and wipes the queue.

## Reproduction

1. Queue 2-3 songs via `/listen` requests.
2. Wait for a jingle to interleave (or for an artist string mismatch on the first queued track).
3. Observe that:

   * The queue section in the UI becomes empty.
   * Skipping tracks shows the songs still play in the expected order.

More likely with `jingleRatio` set low (more frequent jingles) or with tracks whose artist metadata has formatting or transliteration differences between Navidrome and Liquidsoap.

## Files

* `controller/src/broadcast/queue.ts:554-598` - Match logic and `_autoMisses` wipe
* `controller/src/broadcast/queue.ts:671` - Auto-pick gated on `upcoming.length === 0`

## Possible fix

The fix needs to preserve restart recovery while avoiding queue wipes caused by benign metadata mismatches. Options include:

* Ignore jingles and other non-library metadata when counting `_autoMisses`.
* Normalize artist strings before comparison (trim whitespace, normalize Unicode, handle `feat.` vs `featuring`, etc.).
* Verify that Liquidsoap's queue has actually diverged before clearing `upcoming`, instead of relying only on consecutive misses.
* Match queued tracks using a stable identifier wherever possible instead of metadata strings.

## Input needed

@perminder-klair, the queue wipe exists for a legitimate reason after a full restart, but it is also clearing live queues due to normal metadata mismatches.

Would you prefer improving the matching logic, or replacing the three-miss heuristic with a more explicit controller and Liquidsoap synchronization check?

geekylakshya:
## What

Replace the `_autoMisses` counter and threshold in `queue.ts` with a real Liquidsoap telnet sync check. Instead of guessing whether upcoming items are stale after three consecutive metadata mismatches, query Liquidsoap's `dj_queue` directly and reconcile the queue state.

## Why

The `_autoMisses >= 3` heuristic could clear `upcoming` during normal playback when jingles or artist string mismatches caused consecutive match failures. The tracks were already present in Liquidsoap's `dj_queue`, so playback continued correctly, but the web UI showed an empty queue.

Fixes #632.

## How tested

* `cd controller && npm run lint` with 0 errors.

## Notes for reviewers

* Added `getDjQueueIds()` in `liquidsoap-control.ts`, which performs two telnet queries:

  * `dj_queue.queue` to retrieve request IDs.
  * `request.metadata <rid>` for each request ID to extract `subsonic_id`.
* Added a 4-second timestamp cache and in-flight promise deduplication so rapid transitions do not generate repeated telnet requests.
* `reconcileWithDjQueue()` is invoked as fire-and-forget (`void this.reconcileWithDjQueue()`), so the 1.5-second watcher tick is never blocked.
* On telnet failure, the code logs the error and returns without clearing the queue based on guesswork.
* `invalidateDjQueueCache()` is exported for future use but is not wired up yet.

### DIFF
--- a/controller/src/broadcast/liquidsoap-control.ts
+++ b/controller/src/broadcast/liquidsoap-control.ts
@@ -120,8 +120,8 @@ export async function getDjQueueIds(): Promise<Set<string>> {
           if (match && match[1]) {
             subsonicIds.add(match[1]);
           }
-        } catch {
-          // Per-RID errors swallowed (RID may be consumed/played between calls)
+        } catch (ridErr: any) {
+          console.warn(`[liquidsoap] request.metadata ${rid} failed: ${ridErr.message}`);
         }
       }
 
--- a/controller/src/broadcast/liquidsoap-control.ts
+++ b/controller/src/broadcast/liquidsoap-control.ts
@@ -87,3 +87,59 @@ export async function streamStatus() {
   const res = await sendCommand('stream_status', 2000);
   return /\bon\b/i.test(res);
 }
+
+interface DjQueueCache {
+  timestamp: number;
+  ids: Set<string>;
+}
+let _djQueueCache: DjQueueCache | null = null;
+let _djQueueInflight: Promise<Set<string>> | null = null;
+
+// Query Liquidsoap's dj_queue using two telnet hops:
+// 1. dj_queue.queue returns space-separated request IDs.
+// 2. request.metadata <rid> returns metadata for each request ID.
+// Returns a Set of subsonic_ids currently in the queue.
+export async function getDjQueueIds(): Promise<Set<string>> {
+  if (_djQueueCache && Date.now() - _djQueueCache.timestamp < 4000) {
+    return _djQueueCache.ids;
+  }
+  if (_djQueueInflight) {
+    return _djQueueInflight;
+  }
+
+  _djQueueInflight = (async () => {
+    try {
+      const res = await sendCommand('dj_queue.queue', 2000);
+      const rids = res.trim().split(/\s+/).filter(Boolean);
+      const subsonicIds = new Set<string>();
+
+      for (const rid of rids) {
+        try {
+          const meta = await sendCommand(`request.metadata ${rid}`, 2000);
+          const match = /^subsonic_id="([^"]*)"/m.exec(meta);
+          if (match && match[1]) {
+            subsonicIds.add(match[1]);
+          }
+        } catch {
+          // Per-RID errors swallowed (RID may be consumed/played between calls)
+        }
+      }
+
+      const ids = subsonicIds;
+      _djQueueCache = {
+        timestamp: Date.now(),
+        ids
+      };
+      return ids;
+    } finally {
+      _djQueueInflight = null;
+    }
+  })();
+
+  return _djQueueInflight;
+}
+
+export function invalidateDjQueueCache() {
+  _djQueueCache = null;
+}
+
--- a/controller/src/broadcast/queue.ts
+++ b/controller/src/broadcast/queue.ts
@@ -19,6 +19,7 @@ import { logEvent } from '../observability/events.js';
 import { djCallsAllowed } from './listeners.js';
 import * as webhooks from './webhooks.js';
 import * as scrobble from './scrobble.js';
+import * as liquidsoapControl from './liquidsoap-control.js';
 
 // Random gap between DJ links on auto-played tracks. The frequency setting
 // scales how chatty the DJ is:
@@ -85,7 +86,6 @@ class Queue {
   _persistTimer: NodeJS.Timeout | null = null; // debounce for the queue.json snapshot
   _recentPlaysTimer: NodeJS.Timeout | null = null; // debounce for the recent-plays.json sidecar
   _recentPlays: { id: string | null; title: string | null; artist: string | null; endedAt: string }[] = [];
-  _autoMisses = 0;             // consecutive untracked plays — see onTrackStarted
 
   // Snapshot upcoming/current/history to disk. The queue is otherwise purely
   // in-memory, so a controller restart (every `--build controller` rebuild)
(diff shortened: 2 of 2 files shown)


perminder-klair:
Nice direction. Querying Liquidsoap directly instead of guessing with `_autoMisses` is the right call, and I confirmed the telnet contract is solid (`dj_queue.queue` returns the RIDs and `request.metadata` gives back `subsonic_id="..."`, so the regex matches fine). Two things I'd want sorted before merging though:

**1. A just-sent pick can get dropped before Liquidsoap has it.** `item.sent` is set right after we rename `next.txt`, but Liquidsoap only pulls it into `dj_queue` on its next poll up to a second later. If a reconcile fires in that gap (say a jingle starts right after the DJ agent queued its pick), the pick isn't in `liveIds` yet so it gets removed from `upcoming`. That loses its pre-rendered intro and drops `upcoming` to 0, which opens the auto-pick gate and queues a second track. The 4s cache widens the window. It basically reproduces the #632 symptom on a normal autonomous station. Safest fix is to only drop an item once it's been seen in `dj_queue` at least once (or is older than the poll+push latency), and treat an empty or partial `liveIds` as "unknown, drop nothing."

**2. Lost the self-heal for `sent:false` orphans.** Reconcile and the new gate only ever touch `sent:true` items. The 500ms persist debounce can snapshot an item as `sent:false` while drain is still mid-TTS, `recover()` reloads it with no sent filter, and nothing re-drains it at boot. That pins `upcoming.length > 0` forever, so the auto-DJ never picks again and the station just coasts on `auto.m3u`. The old 3-miss clear used to rescue this case.

Couple of smaller ones: the per-rid `catch {}` will silently evict an item if one `request.metadata` call times out, and the filter drops any sent item with no `track.id` since there's no title+artist fallback like `onTrackStarted` has. Also `invalidateDjQueueCache` is exported but unused right now.

Happy to talk through the reconcile timing if useful.

geekylakshya:
Thanks for the detailed review. All points have been addressed in the follow-up commit ([adaaa6c](https://github.com/perminder-klair/subwave/pull/637/commits/adaaa6c7bc6214e163b2269b89377658a4a4d227)).

**Race (just-sent pick dropped before Liquidsoap polls)**

Added a `confirmedInLiquidsoap` flag to queue items. Reconcile now runs in two passes. The first pass marks items currently present in `liveIds` as confirmed. The second pass only drops items that were previously confirmed as present and are now missing. Items that have never been seen in `dj_queue` yet (the in-flight grace period) are retained.

Also added an early bail when `liveIds` is empty while sent items exist. At that point we cannot distinguish between a mid-poll state and a restart, so nothing is dropped.

**Lost self-heal for orphans**

`recover()` now immediately calls `drainToLiquidsoap()` if any `sent:false` items are found, covering the case where the process was snapshotted mid-TTS during a crash.

If any `sent:true` items are present, `recover()` also schedules `reconcileWithDjQueue()` after 3 seconds to give Liquidsoap time to accept telnet connections during startup.

This restores the self-healing behavior that the old `_autoMisses` clear previously provided.

**Smaller fixes**

* Per-RID `catch {}` blocks now log a `console.warn` with the RID and error message instead of silently swallowing the exception.
* Items without a `track.id` in the reconcile filter are now retained (`return true`) instead of being dropped.
* `invalidateDjQueueCache` remains exported as-is since it appears intentional for future use.

perminder-klair:
Pushed a small follow-up to the branch. The empty `dj_queue` check was bailing on every read, so after a full restart the old sent items stayed stuck in `upcoming` and the auto DJ stopped picking. I swapped it for a counter that only clears after 3 empty reads in a row. A jingle or a metadata mismatch gives a non-empty read and resets it, so the #632 wipe can't come back, but a real restart still recovers.

Tested it on a live mixer: queued a track, restarted broadcast, watched the stale item clear after 3 reads and the DJ start picking again. Lint's clean too. Nice work on this one.

PR #662, opened 2026-06-27
perf(tagger): parallel enrichment, phase timings, heavy-embedding-model warning

perminder-klair:
Three library-tagger performance/observability improvements, prompted by a slow tagging run on a NAS (`~250 in a few hours` with `ollama:bge-m3`). Benchmark on a local CPU Ollama: **bge-m3 is ~2.6× slower per embedding than nomic-embed-text** (196 vs 74 ms, 1024-d vs 768-d, 1.22 GB vs 0.38 GB) — but even bge-m3 embeds the whole library in ~90 min, so embeddings are *not* the real bottleneck. The dominant cost is the chat-model tagging phase (and, on low-RAM boxes, model thrash). These changes attack the actual sinks and make them visible.

### 1. Parallelize Phase-0 enrichment (`perf`)
`phaseEnrich` ran a strictly-serial `await`-per-track loop over Last.fm + Navidrome-lyrics fetches — almost all wall-time blocked on the network. Now drains the id list with a bounded worker pool (default **6**, tunable via `TAG_ENRICH_CONCURRENCY`). Per-artist Last.fm lookups cache the in-flight **Promise** (not the resolved value) so concurrent workers on the same artist share one API call. DB writes stay on better-sqlite3 (synchronous) → naturally serialised on the event loop, no locking.

### 2. Per-phase timing breakdown (`feat`)
The tagger now records cumulative wall-clock per phase and surfaces it:
- CLI: `[tag] phase breakdown: seed 8m · learn 6m · embed 2m · …` (slowest first)
- A new optional `timings` field on the terminal `done` progress event
- A **"Last run"** breakdown row in the admin tagging panel

Turns "tagging is slow" into "phase X is most of the time."

### 3. Heavy-embedding-model warning (`feat`)
New name-based `isHeavyEmbeddingModel()` + `embeddingPerfAdvisory()` (gates on provider — cloud models run off-box, never flagged). The **doctor** renders a deterministic `embedding model` finding under *Navidrome & library*: warns when a heavy *local* model (bge-m3 / *-large / qwen) is selected, with a hint pointing at nomic-embed-text + a re-embed. Catches exactly the case that triggered this.

## Testing
- `npm run lint` (eslint + `tsc --noEmit`) clean in both `controller/` and `web/`.
- `isHeavyEmbeddingModel` unit-checked against 10 model names (bge-m3/*-large/qwen → heavy; nomic/3-small/minilm → not).
- Worker-pool semantics verified: every track processed exactly once, all enriched, per-artist fetch fires once per distinct artist under concurrency.

No behaviour change to the music/playback path; embedding model default stays nomic-embed-text.


perminder-klair:
Added regression tests for the new logic (commits `598d303`, `87fbe2a`):

- **`scripts/embedding-heavy.test.ts`** — `isHeavyEmbeddingModel` (bge-m3 / \*-large / qwen → heavy; nomic / 3-small / unknown → not) and `isLocalEmbeddingProvider` (cloud providers never flagged).
- **`scripts/tagger-perf.test.ts`** — `mapPool` (each item once, cap honoured *and* reached, empty / over-cap / NaN-degrades-to-serial, throwing worker rejects), `memoizeByKey` (one underlying call per distinct key under concurrency — the per-artist Last.fm dedup), and the `sortedPhaseTimings`/`formatPhaseBreakdown` shape.

To make these importable I extracted the testable cores into `src/util/async-pool.ts` + helpers in `tagger-progress.ts`/`embedding.ts` (behaviour-preserving; `phaseEnrich` now uses `mapPool`/`memoizeByKey` instead of the hand-rolled loop). Both files are wired into `npm test`.

Heads-up: `npm test` was **already red on develop** — `llm-pure.test.ts` still asserted the pre-#618 "Genre lock" wording. Commit `87fbe2a` refreshes those two assertions to the live "{genre}-only" text (no production change), so the whole chain is green again.

PR #672, opened 2026-06-28
feat(tts): add remote TTS engine for self-hosted HTTP endpoints

### LINKED ISSUE #584: Feature: first-class self-hosted HTTP TTS engine (configurable endpoint) + per-persona voice description
## Summary
Add a TTS engine that points to a **user-configurable HTTP endpoint** speaking the existing `tts-heavy` `/speak` contract, selectable per persona in the UI,  the TTS counterpart of the LLM provider's "OpenAI-compatible / custom base URL" option. As a companion, let a persona carry a free-text **voice description** that is forwarded to the engine (for VoiceDesign / prompt-driven voice models). Related to #579?

## Motivation / use case
In order to get the best performance and stay local, I run a self-hosted GPU voice model (Qwen3-TTS VoiceDesign) in a small sidecar that already implements Subwave's documented `/speak` + `/health` contract. Today the **only** way to make the controller use it without forking is to:

- point `TTS_HEAVY_URL` at my sidecar, and
- have the sidecar **advertise an existing engine name** (`pocket-tts`) so the dispatcher routes it.

That works but it's effectively impersonating a built-in engine, there can only be **one** such endpoint, and the routing is confusing to reason about. A first-class "remote endpoint" engine would make this a clean, supported deployment instead of a clever workaround which is exactly the spirit of `tts-heavy` (keep heavy engines *out* of the controller image).

## Current behavior (for context)
- `controller/src/audio/tts.ts`: `ENGINES = ['piper','kokoro','chatterbox','pocket-tts','cloud']`.
- `cloud` already has a **sub-provider** (`cloudProvider`: `openai`, `elevenlabs`) with per-persona  i.e. the codebase already has the "engine + configurable provider/voice" pattern.
- `ttsHeavyClient.ts` already speaks a clean HTTP contract: `GET /health → {ok, engines[]}` and POST /speak {engine,text,voice,reference_wav,out} → {ok,path,voice_used,fell_back,...}`.
- The `voice` field is already forwarded to the sidecar, it's just that built-in sidecars treat as a voice **id**.

## Proposed solution
1. **New engine** value, e.g. `remote` (or `http` / `self-hosted`), added to `ENGINES`.
2. **Settings**: `tts.remote = { url, /* optional: headers */ }` (mirrors the LLM custom base URL).
3. **Client** `remoteTts.ts` modeled on `ttsHeavyClient.ts`, reuse the `/health` probe loop and `/speak` POST, just with a per-engine configurable URL instead of the single TTS_HEAVY_URL`.
4. **Per-persona voice description**: when `persona.tts.engine === 'remote'`, treat persona.tts.voice` as a free-text description and forward it (already passed as `voice`).  unlocks **distinct voices per persona on a single VoiceDesign model**, today a 
prompt-driven engine can only render one voice because there's nowhere to put the prompt.
5. **UI**: add the engine to the persona/settings dropdown, with a **URL** field (like the LLM custom base URL) and a **"voice description"** textarea.

## Why it fits
- Mirrors the existing `cloud` / `cloudProvider` shape.
- Reuses the already-documented `/speak` + `/health` contract, no new protocol.
- Advances `tts-heavy`'s stated goal: run heavy/GPU engines as sidecars without rebuilding the image.

## Scope (files)
- `controller/src/audio/tts.ts` (`ENGINES`, `resolveEngine`, `speakWith`), new controller/src/audio/remoteTts.ts`, settings schema, web persona/settings form.

## Alternatives considered
- **`cloud` engine**: designed for hosted APIs with keys/models, not an arbitrary self-hosted endpoint.
- **`TTS_HEAVY_URL` + advertising an existing engine name** (what I do today): works but built-in engine, supports a single endpoint, and obscures routing.

## Willing to contribute
Happy to open a PR (branched off `develop`, Conventional Commits title `feat(controller): …`, `.env.example` updated for any new config) when I can find some free time if the direction sounds good, wanted to align on the shape first: engine name (`remote` / `http` / `self-hosted`?), and whether the voice description should be a **new field** vs. **repurposing `persona.tts.voice`** for this engine.

## AI disclosure 

MokipoCode:
## What

Add a `remote` TTS engine — a first-class self-hosted HTTP endpoint speaking the Subwave-native `/speak` + `/health` contract, mirroring the LLM's custom base URL.

## Why

Today running a self-hosted TTS sidecar (Qwen3-TTS, F5-TTS, CosyVoice, etc.) requires impersonating `pocket-tts` — advertising a fake engine name so the dispatcher routes to `TTS_HEAVY_URL`. That works but it's a hack: single
endpoint, confusing routing, no UI for the URL.

A `remote` engine with a configurable URL makes this a clean, supported deployment — exactly the spirit of tts-heavy (keep heavy engines out of the controller image).

Closes #584.

## How tested

- `controller/` — `tsc --noEmit` clean, all test suites pass (`npm test`)
- `web/` — `tsc --noEmit` clean
- Manually verified engine appears in settings dropdowns, persona engine , onboarding wizard, and status badges
- Reviewed against existing engine patterns (chatterbox, pocket-tts, cloud)  consistency in `resolveEngine`, `speakWith`, `availableEngines`, validateTtsBlock`, and `update()`

AI disclosure: YES

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ https://github.com/user-attachments/assets/fac7b723-c54f-438d-8440-bad1e0471e85
 - **Plain-language requests.** "Play something more upbeat" or "anything by Radiohead" works.
 - **Your own music library.** Pulls from Navidrome over the Subsonic API. No external catalogue.
 - **Swappable LLM provider.** Ollama, Anthropic, OpenAI, Google, DeepSeek, OpenRouter, Vercel AI Gateway, or any OpenAI-compatible server. Change it from the admin UI with no redeploy.
-- **Five TTS engines.** Piper and Kokoro in-process for fast local speech, plus an optional `tts-heavy` sidecar (`docker compose --profile tts-heavy up -d`) that adds Chatterbox (zero-shot voice cloning) and PocketTTS (6× real-time, EN/FR/DE/IT/ES/PT). Cloud (OpenAI / ElevenLabs) is also available. Pick a different engine per kind of speech.
+- **Six TTS engines.** Piper and Kokoro in-process for fast local speech, plus an optional `tts-heavy` sidecar (`docker compose --profile tts-heavy up -d`) that adds Chatterbox (zero-shot voice cloning) and PocketTTS (6× real-time, EN/FR/DE/IT/ES/PT). Cloud (OpenAI / ElevenLabs) and a Remote engine (any self-hosted HTTP endpoint, audio over the wire) round it out. Pick a different engine per kind of speech.
 - **Multiple DJ personas.** Up to 10 souls in rotation, each with its own voice and writing style.
 - **Dual-codec broadcast.** MP3 128 kbps for Sonos, hardware radios, and cars; Ogg-Opus 96 kbps for modern browsers. The web player picks automatically.
 - **Native apps and PWA.** Native iOS (on the App Store) and Android (on Google Play) players — background audio, lock-screen / CarPlay / Android Auto controls, multi-station — plus an installable PWA on phone and desktop.
--- a/controller/src/audio/remoteTts.ts
+++ b/controller/src/audio/remoteTts.ts
@@ -1,0 +1,155 @@
+// HTTP client for a user-configured remote TTS engine.
+//
+// When settings.tts.remote.url is set, the `remote` engine speaks the
+// Subwave-native /speak + /health contract against that endpoint — the same
+// protocol the built-in tts-heavy sidecar uses (docker/tts-heavy/server.py).
+// This is the TTS equivalent of the LLM's custom base URL: a first-class,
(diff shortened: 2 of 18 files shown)


perminder-klair:
Thanks for this, it's a clean addition. I pushed a few changes on top.

The main one: the remote engine now returns the audio in the `/speak` response body instead of a shared-volume path. The old path-return only worked if the server shared the controller's filesystem, but the UI points people at a LAN or Tailscale box, so a reachable-but-unshared endpoint went silently dead. Audio over the wire fixes that, no shared volume needed.

Also hardened the rest while I was in there: the health probe now starts after settings load and re-probes when you change the URL (no more 30s stale window), URL validation rejects a malformed host or port at save time, and I dropped an unused helper. Added docs on the Voices manual page and a short news post too.

Tested end to end against a mock endpoint before pushing. Have a look and let me know if anything reads wrong.

PR #673, opened 2026-06-28
feat(web): grey out unavailable TTS engines in the selector

### LINKED ISSUE #585: Enhancement: reflect TTS engine availability in the persona/settings UI (data already exposed)
## Summary
The persona/settings UI lets an operator pick **any** TTS engine, even ones whose backend isn't present in the running image/deployment. Picking an unavailable engine **silently falls back to Piper** at speak time, so the operator believes their selection took effect when it didn't. The backend already exposes exactly the data needed to fix this, so this is a small, front-end-only change.

## Current behavior
- `GET /settings` already returns `{ engines: tts.ENGINES, available: tts.availableEngines() }` (`controller/src/routes/settings.ts`).
- `availableEngines()` reports real availability: `piper: true`; `chatterbox` / `pocket-tts` by build args (venv present) or a reachable sidecar; `cloud` gated by `cloud.isConfigured()` (per provider: `openai`, `elevenlabs`).
- But the UI lists every engine as selectable. At runtime `resolveEngine()` quietly routes an unavailable choice to the saved default / Piper → the segment plays in the **wrong voice** with signal in the UI.

## Proposed solution
In the engine dropdown (persona TTS + global default), **disable / grey out** any option where`available[engine] === false`, with a short tooltip explaining why, e.g.:
- "not built into this image" (chatterbox / pocket-tts without the build arg or sidecar)
- "no API key configured" (cloud / per provider)
- "sidecar offline" (remote/tts-heavy not responding)

Optionally a small "unavailable" badge. Purely consumes the already-shipped `/settings.available` map : no backend change required.

## Why it fits
- The data is already there; this just surfaces it.
- Same spirit as #238 (making silent voice substitutions visible), prevent the operator from running in a fallback voice.

## Scope (files)
- Web settings/persona form only; consume the existing `available` field from `GET /settings`.

## Willing to contribute
Happy to PR this (off `develop`, Conventional Commits title `feat(web): …`) when I can find some free time alongside or independently of the remote-engine request, it's a self-contained front-end change.

## AI disclosure

MokipoCode:
## What

Disable TTS engine cards in the selector when `available[engine] === false`, with a hover tooltip explaining why. The currently-selected engine stays enabled so an existing setup is never stranded.

Persona page derives per-persona cloud availability from `cloudByProvider`, so an ElevenLabs persona on an OpenAI-only station sees Cloud as unavailable.

## Why

Today every engine is selectable regardless of availability. Picking a dead engine silently falls back to Piper — the operator believes their selection took effect when it didn't. The backend already ships the `available` map in `GET /settings`; this just surfaces it in the UI.

Closes #585.

## How tested

- `web/` — `tsc --noEmit` clean
- Reviewed all 6 engine cards with each availability state (piper always, kokoro/chatterbox/pocket-tts/cloud/remote disabled with tooltip unavailable, currently-selected engine always clickable)
- Verified per-persona cloud availability: switching provider in the dropdown the Cloud card's disabled state

AI disclosure: YES

### DIFF
--- a/web/components/admin/SettingsPanel.tsx
+++ b/web/components/admin/SettingsPanel.tsx
@@ -20,6 +20,7 @@ import {
 } from '../ui/select';
 import { Card, Btn, Pill, Eyebrow, Seg, Metric } from './ui';
 import { EngineSelector } from './tts/EngineSelector';
+import type { TtsAvailable } from './tts/engineMeta';
 import { VoicePreviewButton } from './tts/VoicePreviewButton';
 import { ProviderSelector } from './llm/ProviderSelector';
 import { EmbeddingProviderSelector } from './embedding/EmbeddingProviderSelector';
@@ -306,7 +307,7 @@ interface SettingsData {
   };
   tts?: {
     engines?: string[];
-    available?: Record<string, boolean>;
+    available?: TtsAvailable;
     kokoroVoices?: Array<{ id: string; label: string }>;
     chatterboxVoices?: string[];
     // `voiceDir` is the new shared name (issue #213). `chatterboxVoiceDir` is
--- a/web/components/admin/personas/PersonaVoiceCard.tsx
+++ b/web/components/admin/personas/PersonaVoiceCard.tsx
@@ -70,6 +70,25 @@ export function PersonaVoiceCard({ persona, data, defaultEngine, cloudIssueText,
     updateTts(patch);
   };
 
+  // Derive a per-persona availability map. The global `available` map has a
+  // single `cloud` flag (any provider configured). Per-persona, cloud is only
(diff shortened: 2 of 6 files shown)


perminder-klair:
Thanks for this, the code is clean and does exactly what it describes.

Testing it though, I hit a snag with the core approach. Greying out the unavailable engines also hides the inline help that tells you how to turn them on, so you land in a catch-22: you can't open Chatterbox or Cloud to read the "run docker compose --profile tts-heavy up -d" steps, and you can't enable them without those steps. The hover tooltip on its own isn't enough (no command, hover only, rough on touch and screen readers).

I'd rather solve #585 in a way that keeps that guidance visible, so I'm going to close this one. Really appreciate the effort and the clear write-up, hope you'll send more our way.

PR #676, opened 2026-06-29
feat(boardcast): configurable stream bitrate

TomiEckert:
MP3 bitrate setting

## What
Added an option to change the bitrate of the mp3 stream. Same exact implementation as the archive bitrate setting, except the UI part is in the Danger Zone.

## Why
I wanted to change the bitrate from 192 to 320 kbps.

## How tested
Built the docker containers from scratch, played a song at default 192 kbps, then tried 64 kbps, then 320 kbps. Clear difference in quality.

image

### DIFF
--- a/controller/src/broadcast/liquidsoap-control.ts
+++ b/controller/src/broadcast/liquidsoap-control.ts
@@ -52,15 +52,61 @@ export function sendCommand(cmd: string, timeoutMs = 3000): Promise<string> {
   });
 }
 
+const sleep = (ms: number) => new Promise(resolve => setTimeout(resolve, ms));
+
+// Quick liveness probe — can we open a TCP connection to the telnet port?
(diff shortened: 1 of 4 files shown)


perminder-klair:
Tested this locally and the stream re-encodes cleanly at every setting (checked 64, 192 and 320 with ffprobe). Nice work, and it follows the archive bitrate path closely.

I pushed two small follow-ups to the branch:
- renamed ARCHIVE_BITRATES to MP3_BITRATES, since the set now feeds both the archive and the live stream, and reworded the 192 comment so it reads as the default rather than a one-off bump
- fixed a separate Restart Mixer bug I hit while testing, where it could report success without actually cycling liquidsoap, so a new bitrate never applied until a manual restart

Good to merge once CI is green.

PR #697, opened 2026-06-30
Feat/persona music only

### LINKED ISSUE #666: Per-persona voice-off (music-only) mode
## Summary

Add a **per-persona** way to silence the spoken DJ entirely while keeping the smart AI track picker running. Some shows should be able to talk while others stay pure music.

Two listeners on r/selfhosted have asked for "this but no audible DJ" / "how do I turn off the AI voice for a show." There is currently no first-class way to do it.

## Current state

- The least-talkative lever today is the persona `frequency` ladder (`quiet | moderate | aggressive`). `quiet` reduces but does not eliminate voice: occasional station IDs, an hourly time check, and between-track links still fire.
- Personas (`validatePersonasStrict` in `controller/src/settings.ts`) and shows have **no** silent/mute flag.
- The only way to get true zero-voice right now is the blunt token-cap hard tier, which also disables the smart picker and coasts on the plain `auto.m3u` shuffle — wrong tradeoff.

## Proposal

Make it a per-persona setting so it inherits naturally through shows (each show already references a `personaId`). Two shapes to consider:

1. A `'silent'` rung added to the frequency ladder, **or**
2. A dedicated `voice: on | off` (or `musicOnly: true`) flag on the persona.

When voice is off for the on-air persona:
- Keep the AI picker / session DJ agent fully active (track selection, recency, show genre/era steering all unchanged).
- Suppress **all** spoken output for that persona: between-track links, station IDs, hourly time check, and the optional segments (weather/news/curiosity/etc.).
- Listener requests: decide whether the request still plays the track silently (likely yes) vs. speaks an intro (no, when muted).

## Implementation notes

- The internals already know how to mute optional segments — `broadcast/dj-budget.ts` (`optionalSegmentsAllowed()`) gates links / station IDs / hourly / segments in soft/hard budget mode. A persona-level mute can reuse the same suppression chokepoints rather than threading a new flag through every caller.
- Surfaces to gate: `broadcast/dj-agent.ts` (link generation in `runTrackEvent`), `broadcast/dj-gate.ts` (`shouldFire` for stationId/hourly), the segment-director floor in `skills/_agent.ts`, and request-intro TTS in `broadcast/queue.ts` / `routes/request.ts`.
- Add the field to `DEFAULTS`, the persona validator, and the admin persona editor (`web/`).
- Picker path must stay untouched — this is a voice mute, not a model-call kill switch (distinct from the budget hard tier).

## Out of scope

- Global one-click mute (per-persona covers it; a station-wide default can come later if wanted).

geekylakshya:
Closes #666.

## What

Adds a `musicOnly` boolean to personas. When enabled, the AI picker and session agent continue running normally, including track selection, recency handling, and show steering, but all spoken output is suppressed.

## How it works

A single `voiceSilenced()` helper in `settings.ts` reads `getEffectivePersona()?.musicOnly`. Every voice output path checks this helper.

* `shouldFire()` in `dj-gate.ts` returns `false` for all announcement types (`stationId`, `hourly`, and `segments`).
* `agenticTick()` in `skills/_agent.ts` returns early before making any LLM calls.
* `runTrackEvent()` in `dj-agent.ts` forces `wantLink` to `false` before the `djMode` and budget branches.
* `runRequest()` in `dj-agent.ts` sets `introScript` to `null`.
* Both `generateIntro()` call sites in `routes/request.ts` are short-circuited.
* `drainToLiquidsoap()` in `queue.ts` skips TTS rendering. This covers the case where the persona switches after a track is queued but before TTS is rendered.
* `airIntro()` in `queue.ts` skips airing the rendered WAV. This covers the case where the persona switches after rendering but before the track starts.
* `announce()` in `queue.ts` returns early, covering `/dj/say` and manually triggered skills.

## UI

Replaced the separate `djMode` and `musicOnly` toggles with a single **DJ Style** three-option picker:

* Music only
* Narrator
* Full DJ

Screenshot 2026-06-30 142613

The selected option derives both fields from a single source, preventing conflicting combinations. A live status line below the picker explains the behavior of the currently selected mode.

## How tested

Deployed on the testing LXC and verified the following:

* Music only enabled: no links, station IDs, hourly announcements, or segments are generated.
* Listener requests are queued and played without spoken output.
* The AI picker continues selecting tracks normally.
* Switching back to Narrator or Full DJ restores voice output on the next track.
* `cd controller && npm run lint` reports 0 errors.
* `cd web && npm run lint` reports 0 errors.


perminder-klair:
Thanks for this, Lakshay. Great work and a really clear write-up. We ended up shipping the same capability in #922, where a new "silent" frequency stop suppresses all links, idents, hourlies, banter and segments while still letting listener requests and manual segment triggers speak. That covers the music-only behaviour you built here for #666, just as a frequency rung instead of a separate toggle. Going to close this one, but genuinely appreciate you taking the time on it.

Issue #238, opened 2026-06-01
Pocket-TTS - voice selection not taking effect


WhatKateDoes:
Thanks for the update - Pocket-TTS voices can now be selected and saved/applied, but they are not taking effect, it's using one of the default male voices regardless of selection.

perminder-klair:
Thanks! tracked it down. The voice selection was saving fine; the problem is specifically cloned .wav voices.

PocketTTS voice cloning needs Kyutai's gated kyutai/pocket-tts model, which only downloads with a Hugging Face token. The image builds without one, so it loaded the non-cloning weights and quietly fell back to a built-in voice, with no error shown, which is why it looked like the selection was ignored. (Built-in voices like Alba/Charles were never affected — only clones.)

Fix is in #240:
- Cloning now works if you opt in: accept the terms at https://huggingface.co/kyutai/pocket-tts, add HF_TOKEN=hf_... to your .env, and restart the tts-heavy sidecar.
- No more silent fallback, if cloning isn't available, the admin UI now warns you on the persona, and the logs say exactly which voice actually played vs. what was requested.

It'll land in the next release. If you'd rather not set up a token, the built-in voices work with no extra steps.

PR #702, opened 2026-06-30
feat(web): surface TTS engine availability before selection (#585)

### LINKED ISSUE #585: Enhancement: reflect TTS engine availability in the persona/settings UI (data already exposed)

MokipoCode:
## What

Mute unavailable engine cards in the TTS selector (persona + global default) and show a persistent, accessible note below the grid when the selected engine is unavailable, carrying the reason and the exact enable step.

Closes #585.

## Why

#673 disabled unavailable cards, which created a catch-22: the inline setup guidance (#238) is gated on the engine being selected, so a greyed-out card hid the very steps needed to enable it.

This retry keeps every card clickable and adds a `role="status"` DOM strip instead of a hover `title`. Visible on touch and screen readers, with the actual command (`docker compose --profile tts-heavy up -d`, etc.).

## How

- `engineMeta.ts`: new `TtsAvailable` typed interface (no `[key:string]:any`)+ pure `engineEnableHint()` sibling to `engineStatus()`.
- `EngineSelector.tsx`: muted (`opacity-60`) but always clickable; persistent note below the grid when the selected engine is unavailable.
- `PersonaVoiceCard.tsx`: per-persona cloud availability via `cloudByProvider[provider]`, not just the global flag.
- `SettingsPanel.tsx` / `personas/types.ts`: `available` typed via `TtsAvailable`. Existing `HeavyEngineSetupGuide` / #238 callouts untouched.

## Tested

- `cd web && npm run lint`: clean
- `/admin/settings` (Voice tab) + `/admin/personas`: unavailable engines muted but clickable; selecting one shows the persistent note with the exact command; saved-but-unavailable engine stays selectable (no catch-22); ElevenLabs persona on OpenAI-only station sees cloud as unavailable.

AI disclosure: YES

### DIFF
--- a/web/components/admin/personas/PersonaVoiceCard.tsx
+++ b/web/components/admin/personas/PersonaVoiceCard.tsx
@@ -12,7 +12,7 @@ import { CB_DEFAULT_VOICE, KOKORO_RE, CHATTERBOX_VOICE_RE, POCKET_TTS_VOICE_RE }
 import { Card, Seg } from '../ui';
 import { EngineSelector } from '../tts/EngineSelector';
 import { VoicePreviewButton } from '../tts/VoicePreviewButton';
-import { ENGINES } from '../tts/engineMeta';
+import { ENGINES, type EngineAvailability } from '../tts/engineMeta';
 import { Input } from '../../ui/input';
 import { Label } from '../../ui/label';
 import {
@@ -72,6 +72,21 @@ export function PersonaVoiceCard({ persona, data, defaultEngine, cloudIssueText,
     updateTts(patch);
   };
 
+  // Resolve Cloud against this persona's saved provider even before the Cloud
+  // card is selected. openai-compatible has no key-based availability entry and
+  // is trusted by the server, while unknown providers retain the global status.
+  const globalAvail = data?.tts?.available as EngineAvailability | undefined;
+  let selectorAvailable = globalAvail;
+  if (globalAvail) {
+    const prov = persona.tts.cloudProvider;
+    const cloudByProv = globalAvail.cloudByProvider;
+    if (cloudByProv && prov in cloudByProv) {
+      selectorAvailable = { ...globalAvail, cloud: cloudByProv[prov] };
+    } else if (prov === 'openai-compatible') {
+      selectorAvailable = { ...globalAvail, cloud: true };
+    }
+  }
+
   return (
     <Card flat title="Voice" sub="text-to-speech engine">
       {/* Engine — radio-card grid, full width above the two-column body. */}
@@ -80,7 +95,7 @@ export function PersonaVoiceCard({ persona, data, defaultEngine, cloudIssueText,
         <EngineSelector
           value={persona.tts.engine}
           engineIds={ENGINE_IDS}
-          available={data?.tts?.available}
+          available={selectorAvailable}
           onChange={selectEngine}
         />
         <div className="field-hint max-w-[70ch]">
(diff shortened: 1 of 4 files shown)


perminder-klair:
Nice fix, this sorts out the #673/#238 catch-22 well. Keeping the cards clickable and putting the reason plus enable command in a persistent note is the right call. I checked that `TtsAvailable` matches the controller's `availableEngines()` shape and the casts typecheck under strict. The per-persona `cloudByProvider` check and using `role="status"` over a hover title are both good improvements.

A few notes:

1. Worth fixing before merge: the note's closing line, "Until then this persona falls back to the default engine", reads wrong on the Settings default-engine tab. There's no persona there, and the engine being configured is the default, so it falls back to itself. Maybe pass a prop or word it neutrally.
2. Optional: the cloud card only reflects the persona's provider once cloud is selected, not before.
3. Optional: the note uses the danger colour where the card badge uses the softer warn tone.
4. Branch looks slightly behind develop, so a rebase would be worth it.

Only item 1 feels worth blocking on.

MokipoCode:
Hi, sorry for the slow follow-up; I have been busy lately. Thanks for the thoughtful review and for calling out the Settings wording issue.

I have now rebased the branch onto the current develop and refreshed the change against the new EngineAvailability / heavyEnabled status model. The shared note is neutral now, so it reads correctly on both Settings and Personas. I also kept the per-provider Cloud status visible before Cloud is selected, switched the note to the softer warning treatment, and made sure workers that are still starting remain visually active while genuinely unavailable cards stay muted and clickable.

I verified the refreshed branch locally on Node 22: npm run lint and npm run build both pass.

perminder-klair:
Thanks again for this — the approach held up well in review. develop moved underneath it (#759 and #1009 reshaped `PersonaVoiceCard`), so I've rebased it in #1061 with your original commit preserved, plus a small follow-up commit that folds `engineEnableHint()` into `engineStatus()` (single branch tree, machine-readable `state` instead of comparing badge labels) and keeps the `role="status"` region mounted. Closing this one in favour of #1061.

PR #703, opened 2026-06-30
Feat/plex music source

### LINKED ISSUE #692: Add pluggable music sources (Plex + local folder) alongside Navidrome
## Context

SUB/WAVE originally had exactly one music backend: Navidrome, reached through the Subsonic API. This issue scopes making the music backend **pluggable** so operators can run the station on a source other than Navidrome.

It started as "add Plex" and grew into a proper source abstraction: one active source at a time, chosen in Admin → Settings (and in the first-run wizard), mirroring the existing LLM provider pattern (registry + pure capabilities table + `MUSIC_SOURCES` enum + card-grid selector).

## Scope

Three sources behind one `MusicSource` facade (`music/source.ts`); no call site branches on the source id:

- **Navidrome / Subsonic** — the existing client, now the first `MusicSource` implementation. Behaviour is byte-identical.
- **Plex** — a Plex Media Server over its HTTP API (`PLEX_URL` + `PLEX_TOKEN`, optional `PLEX_LIBRARY`). Keeps the listening-history signals (playlists, ratings-as-stars, play-count ranking); no Last.fm graph, bios, or lyrics.
- **Local folder** — point SUB/WAVE at a directory of audio files (default `state/music`), no server at all.

When a source lacks a capability, the facade returns a neutral empty and the DJ's picker never offers the corresponding tool — nothing breaks, the station just has fewer discovery signals.

## Design

- **Interface + registry.** `music/sources/types.ts` (the `MusicSource` contract, core methods required, discovery methods optional), `music/sources/capabilities.ts` (pure per-source capability table), `music/sources/registry.ts` (resolves the active source from `settings.music.source`, signature-cached — same pattern as the LLM registry). `music/source.ts` is the facade every call site imports.
- **Frozen wire name.** The `subsonic_id` annotate field stays verbatim across radio.liq, liquidsoap-control, now-playing.json, and the web + native apps — it just means "track id".
- **Track-id stability.** Mood/analysis rows are keyed by track id. Local ids are path-derived (renames re-mint them) and a Navidrome rescan can re-mint too, so `music/track-identity.ts` + `library-db.ts` carry tags/analysis/embeddings across id changes before anything is pruned. On a source switch, matching rows migrate first, then unmatched rows are removed.
- **Config / settings / onboarding.** `settings.music.source` enum + admin exposure; `doctor` and `firstRun.needsSetup` are source-aware; the wizard's first step becomes "Music source". A server-less source flips `needsSetup` false with no Navidrome creds, so a fresh install can go local-only end to end.

## Out of scope

- No merged dual-library (selector, one source at a time). Ids are source-tagged so this can be revisited.
- No remote/HTTP plugin source yet — the `MusicSource` interface is the seam it will implement later.
- No reliance on Plex Pass sonic analysis.

## Status

Delivered in #843 (pluggable `MusicSource` layer — Subsonic, Plex & local-folder). The earlier Plex-only community PR #703 was closed in favour of it.

geekylakshya:
Closes #692.

## What

Adds Plex as a fully supported music source alongside Navidrome. Operators can switch between the two in the admin settings. All major subsystems, including library walks, cover art, tagging, enrichment, playback, the doctor, and the coverage counter, now route through the active source instead of assuming Navidrome.

## Why

Plex users did not have a fully working integration.

* Cover art requests returned 404s because Plex track artwork is resolved through the album rather than the track.
* The library was never walked or tagged because the tagger was hardcoded to Navidrome.
* Last.fm enrichment was silently skipped because the fallback path always called Navidrome, even when Plex was the active source.
* Tracks with ambiguous metadata could become permanently stuck in the untagged pool because an empty mood response still wrote `tagged_at`, marking the track as processed without valid tags.

## How tested

* Switched the music source to Plex in the admin settings and verified that credentials persist across restarts and source changes.
* Ran a full library walk and tagger against a Plex library. Confirmed that cover art, playback URIs, and mood tags are handled correctly.
* Verified that cover art resolves correctly for tracks that have not yet been imported into the database by using the live API fallback.
* Confirmed that tracks receiving empty mood results from the LLM remain in the untagged pool instead of being permanently marked as tagged.
* Verified that Last.fm enrichment runs correctly for Plex artists, both with and without an API key configured.
* Confirmed that the coverage counter updates automatically after a tagger run without requiring a manual refresh.
* Regression tested the Navidrome path to ensure all existing behavior is preserved.

## Notes for reviewers

`getCoverArtUrl()` on the Plex source is now asynchronous, while the Navidrome implementation remains synchronous. `public.ts` branches on the `plex:` ID prefix and awaits only the Plex implementation when required.

For tagging, the Plex **Mood** field (acoustic and MusicBrainz-derived tags stored per track) is used as the equivalent contextual signal that lyrics provide on the Navidrome path. It serves the same purpose for the LLM tagger, even though the underlying data differs.

Migration 11 adds the `plex_thumb` and `plex_part_key` columns. When Plex is the active source, the tagger backfills both fields during the first tagging pass.

### DIFF
--- a/cli/src/doctor.ts
+++ b/cli/src/doctor.ts
@@ -187,7 +187,7 @@ async function checkController(compose: ComposeStatus): Promise<Finding[]> {
       out.push({
         label: 'setup',
         status: 'fail',
-        detail: 'incomplete — Navidrome + LLM not configured',
+        detail: 'incomplete — music source + LLM not configured',
         hint: `Run \`subwave setup\` or open ${webBaseFor(compose.env)}/onboarding to finish configuration.`,
       });
     } else if (needsSetup === false) {
--- a/controller/src/broadcast/dj-agent.ts
+++ b/controller/src/broadcast/dj-agent.ts
@@ -340,6 +340,8 @@ function trackFields(song) {
     album: song.album,
     year: song.year,
     genre: song.genre,
+    path: song.path ?? null,
+    _partKey: song.plexPartKey || song._partKey || null,
   };
 }
 
--- a/controller/src/broadcast/queue.ts
+++ b/controller/src/broadcast/queue.ts
@@ -6,7 +6,7 @@ import { writeFile, readFile } from 'node:fs/promises';
 import { existsSync, readFileSync, openSync, readSync, closeSync, statSync } from 'node:fs';
 import { stat, rename } from 'node:fs/promises';
 import { config } from '../config.js';
-import * as subsonic from '../music/subsonic.js';
+import { getSource } from '../music/source/index.js';
 import * as mix from '../music/mix.js';
 import * as library from '../music/library.js';
 import { speak, voiceGainDb } from '../audio/tts.js';
@@ -488,7 +488,7 @@ class Queue {
         // (requestedBy set) stay exempt — a requested long mix plays in full,
         // mirroring the request path's selection-cap exemption in picker-tools.
         const maxDurationSec = item.requestedBy ? null : settings.effectiveMaxTrackSec();
-        const uri = subsonic.getAnnotatedUri(item.track, { maxDurationSec });
+        const uri = getSource().getAnnotatedUri(item.track, { maxDurationSec });
         await writeHandoff(config.liquidsoap.queueFile, uri);
         item.sent = true;
         this.persist();  // record the sent flag — these are now live in dj_queue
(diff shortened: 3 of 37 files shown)


geekylakshya:
image

perminder-klair:
Thanks for this, Lakshay, and for the thorough testing notes. We went with a broader approach in #843 that puts Navidrome, Plex and a local-folder option behind one pluggable music-source layer, so the Plex support you built here landed as part of that. Really appreciate you digging into the cover-art and tagger edge cases, those writeups were genuinely helpful. Closing this out in favour of #843.

PR #706, opened 2026-06-30
Fix/acoustic analysis disabled state

geekylakshya:
## What

The **Re-analyse acoustics** pass in the **Rescan** tab was always clickable, even when no acoustic analysis engine was running. The **Analyze acoustics** pass in the **Run** tab already had the correct disabled logic, but there was no visual styling to indicate that it was disabled.

## Why

The **Rescan** tab was missing `disabled={p.analysisOff}` entirely, most likely due to a copy-and-paste oversight when the tab was implemented. As a result, users could select the pass and start a rescan that would silently do nothing, with no indication of why.

The **Run** tab already had the correct disabled guard, but `.lib-pass:disabled` had no corresponding CSS rule. Disabled buttons therefore appeared identical to enabled ones across both tabs.

## How tested

* Verified that with no `tts-heavy` sidecar running, both **Analyze acoustics** (Run tab) and **Re-analyse acoustics** (Rescan tab) are greyed out and cannot be clicked.
* Started the `tts-heavy` sidecar and confirmed that both passes become fully interactive.
* Verified that there are no regressions for the other library passes (`enrich`, `embed`, `tag`, `reseed`, and `upgrade`).

### DIFF
--- a/web/app/globals.css
+++ b/web/app/globals.css
@@ -3387,6 +3387,7 @@ html.lite *::after {
     padding: 12px 13px; cursor: pointer; text-align: left; font-family: inherit;
 }
 .admin-root .lib-pass:hover { border-color: var(--ink); }
+.admin-root .lib-pass:disabled { opacity: 0.4; cursor: not-allowed; pointer-events: none; }
 .admin-root .lib-pass.on { border-color: var(--accent); background: var(--accent-soft); }
 .admin-root .lib-pass .box {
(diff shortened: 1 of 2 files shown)


perminder-klair:
Superseded by #767, which re-applies this fix against current `develop` (this branch had gone stale — merge conflicts + the hint referenced the tts-heavy sidecar, but analysis is now the default-on standalone `analyzer` service since #717).

PR #711, opened 2026-06-30
feat(tts-heavy): add DISABLE_CHATTERBOX and DISABLE_POCKET_TTS env vars

### LINKED ISSUE #710: tts-heavy: add DISABLE_CHATTERBOX env var to skip loading unused engines
## Problem

When `tts-heavy` starts, it always initializes all TTS workers (Chatterbox, PocketTTS, and Kokoro), regardless of which engine is selected in the settings.

On CPU-only hosts using `defaultEngine: pocket-tts`, the Chatterbox worker still loads its large neural TTS model during startup even though it is never used. The model then remains resident for the lifetime of the process.

Although the worker is idle, the kernel eventually pages much of the model out to swap after peak memory usage. Those swapped pages are not proactively reclaimed, so swap usage remains high despite plenty of free RAM.

Observed on a stock installation with `TTS_HEAVY_DEVICE=cpu` and `defaultEngine: pocket-tts`:

* `chatterbox_worker`: ~300 MB RSS, ~3 GB swap
* `pocket_tts_worker`: ~200 MB RSS, ~450 MB swap
* Total swap usage: ~3.5 GB out of 4 GB, while approximately 5.9 GB of RAM remained free

## Proposed fix

Add a `DISABLE_CHATTERBOX=1` environment variable (and ideally `DISABLE_POCKET_TTS=1`) that prevents the corresponding worker from starting during the FastAPI lifespan.

The worker objects can still be instantiated so that routes such as `/health` and `/speak` continue to function correctly. Only the background `run()` task should be skipped, preventing the model from being loaded until that engine is actually enabled.

For example:

```python
DISABLE_CHATTERBOX = os.environ.get("DISABLE_CHATTERBOX", "").lower() in ("1", "true", "yes")

# In lifespan:
tasks = [pocket_worker.run(), analyzer_worker.run()]
if not DISABLE_CHATTERBOX:
    tasks.insert(0, chatterbox_worker.run())
```

Operators who do not use Chatterbox could then simply add the following to their `.env` file:

```text
DISABLE_CHATTERBOX=1
```

This avoids loading an unused model, significantly reduces memory pressure on CPU-only systems, and prevents several gigabytes of unnecessary swap usage.

geekylakshya:
Fixes #710.

## What

Adds `DISABLE_CHATTERBOX=1` and `DISABLE_POCKET_TTS=1` environment variables to the `tts-heavy` sidecar.

When either variable is set, the corresponding TTS worker supervisor is not started, preventing that worker's model weights from being loaded into memory.

## Why

On CPU-only hosts, both TTS workers load their full model weights during startup regardless of which engine is selected in the settings.

Although unused workers remain idle, the kernel eventually pages much of their model weights to swap after peak memory usage. Those pages are not proactively reclaimed, resulting in high swap usage despite ample free RAM. On a stock installation with `defaultEngine: pocket-tts`, the idle Chatterbox worker alone consumed approximately 3 GB of swap while more than 5 GB of RAM remained available.

## How tested

* Set `DISABLE_CHATTERBOX=1` in `.env`, recreated the sidecar container, and confirmed that `chatterbox_worker.run()` is never started and no Chatterbox process appears in `ps aux`.
* Verified that `/health` still returns `ok: true` with `chatterbox_loaded: false`, and that the controller falls back to Piper as expected.
* Confirmed that PocketTTS continues to function normally when it is the active engine.
* Verified `DISABLE_CHATTERBOX=1` and `DISABLE_POCKET_TTS=1` together in analysis-only mode. The analyzer worker starts, `/analyze` responds successfully, and no TTS worker processes appear in `ps`.

## Notes for reviewers

The worker objects (`chatterbox_worker` and `pocket_worker`) are still instantiated. Only their background `run()` supervisor tasks are skipped. This allows endpoints such as `/health` and `/speak` to return a clean `{"ok": false, "engines": [...]}` response or an HTTP 500 `"worker not ready"` error instead of failing due to missing worker state.

There is no behavior change for existing installations that do not set these environment variables, as both default to an empty string and therefore remain disabled by default.

### DIFF
--- a/cli/src/assets.generated.ts
+++ b/cli/src/assets.generated.ts
@@ -248,6 +248,12 @@ services:
       # host + nvidia runtime; the server gracefully falls back to cpu when
       # CUDA isn't actually available.
       - TTS_HEAVY_DEVICE=\${TTS_HEAVY_DEVICE:-cpu}
+      # Set DISABLE_CHATTERBOX=1 or DISABLE_POCKET_TTS=1 in your root .env to
+      # skip loading that engine entirely. The model is never started so no RAM
+      # or swap is consumed. Useful when you only use one TTS engine, or when
+      # you run tts-heavy solely for acoustic analysis (both engines disabled).
+      - DISABLE_CHATTERBOX=\${DISABLE_CHATTERBOX:-}
+      - DISABLE_POCKET_TTS=\${DISABLE_POCKET_TTS:-}
       # Default voice id used by PocketTTS when a persona doesn't override it.
       - POCKET_TTS_VOICE=\${POCKET_TTS_VOICE:-alba}
       # Optional — enables PocketTTS zero-shot voice CLONING (issue #238). The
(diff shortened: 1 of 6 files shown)


perminder-klair:
closing this as we going with this direction instead https://github.com/perminder-klair/subwave/pull/717

Issue #700, opened 2026-06-30
Feature Request: PostgreSQL support


Wayne-Harbart:
Currently Subwave uses SQLite. Would love PostgreSQL (ideally with pgvector) for production deployments on Kubernetes where shared storage for SQLite is a pain and vector similarity search on embeddings would unlock better DJ picks.

perminder-klair:
currently out of scope, so closing this.

PR #727, opened 2026-07-01
feat(db): optional PostgreSQL + pgvector storage backend

Wayne-Harbart:
Closes https://github.com/perminder-klair/subwave/issues/700

## Summary

Adds an optional **PostgreSQL + pgvector storage backend**, selected by setting `DATABASE_URL`. Nothing changes for anyone who doesn't set it — SQLite (`better-sqlite3` + `sqlite-vec`) remains the default, and both backends now pass an identical integration contract suite so they can't drift.

This PR originally shipped as "Phase 1 — write-path only"; it has since grown into the complete feature (read + write), because merging the write-path alone would have left a documented split-brain (tagger writes Postgres, live picker reads SQLite). What's here now is: **set the env var, everything works.**

```bash
DATABASE_URL=postgres://user:pass@host:5432/subwave
```

Full docs in `POSTGRES.md` (repo root).

## What's in it

- **`LibraryDbAdapter` interface** (`controller/src/music/db/adapter.ts`) with two implementations: `SqliteAdapter` (existing backend, behaviour unchanged) and `PostgresAdapter` (new; JSONB columns, native `vector` type, idempotent auto-migrations, advisory-locked so concurrent starts can't race DDL).
- **Async read-path** — `library.ts` and everything above it (picker, scheduler, DJ agent, LLM tools, routes) is now `async`. Mostly mechanical `await` insertion; the SQLite path just wraps its sync calls in resolved Promises.
- **Enqueue-time hydration** — hot-path helpers that used to do sync per-call library lookups now read off the track object, hydrated once at `queue.push()` / before the show-energy lean. No behaviour lost for Subsonic-sourced tracks (mix transitions, loudness normalisation, intro budgets, energy leans all still work).
- **HNSW vector indexes** — IVFFlat trains centroids at `CREATE INDEX` time, and these indexes are created on empty tables, so default `probes=1` silently dropped KNN rows (the new contract test caught a 3-row table returning 2 neighbours). HNSW has no training step. Migration v10 converts existing installs automatically.
- **Adapter contract test suite + CI** — see Testing below.

## Review map

~1,700 lines changed, but most of it is mechanical. Where to spend your attention:

| Needs real eyes | Why |
|---|---|
| `src/music/db/postgres.ts` | The new adapter (~1,300 lines): SQL, migrations, vector handling |
| `src/broadcast/queue.ts` | `hydrateAcoustics()` at `push()`; three helpers now read off the track object |
| `src/music/show-filter.ts` | `preferEnergyHydrated()` — energy backfill for Subsonic candidates |
| `src/broadcast/dj-agent.ts` | `advanceRun()` now async so the sonic-journey attach can't race the first waypoint |
| `src/llm/internal/tools/picker-tools.ts` | `slim()`/`collect()` went async; 17 call sites |

| Mechanical (safe to skim) | |
|---|---|
| `src/music/library.ts`, `journey.ts`, `genre-suggest.ts`, `picker.ts`, `scheduler.ts`, `doctor.ts` | `await` insertion, signatures → `Promise<T>` |
| `src/routes/*` (7 files), `agent-factory.ts`, `intro-budget.ts` | Same |

Commits are layered for review: cascade+hydration → HNSW migration → tests+CI → docs.

## Testing

**`scripts/adapter-contract.ts`** runs the *same 25 assertions* against both backends — migration idempotency, tag/enrichment/analysis writes, text + audio KNN ordering and seed exclusion, filter facets and pagination, stats, prune, and the reseed/`adoptStoredDim` dim-negotiation guards:

- `npm run test:sqlite` — default backend in a temp `STATE_DIR`, no services needed (skips cleanly on a Node build without better-sqlite3 bindings)
- `npm run test:pg` — real Postgres via `TEST_DATABASE_URL` (skips when unset; refuses to run against a db whose name doesn't contain "test", since it wipes tables)

Both are part of `npm test`. The new **`.github/workflows/test.yml`** runs the full unit suite plus both contract suites on every PR with a `pgvector/pgvector:pg16` service container — `npm test` previously wasn't exercised in CI at all.

Beyond the suites, this has been soaking on a live station against a real library: **76,068 tracks / 76k text vectors (1536-d) / 8.7k audio CLAP vectors**. Picker, DJ agent, requests, admin browse, and the tagger all running against Postgres. 15-NN over the 76k HNSW index: **~8 ms**.

## Behaviour notes / honest limitations

- **v10 migration rebuilds the vector index once** on first boot after upgrade (Postgres installs only). On a large library this takes minutes at the default `maintenance_work_mem`; POSTGRES.md documents bumping it to speed this up. SQLite installs are untouched.
- **No SQLite → Postgres data migrator.** The recommended path is a fresh `npm run tag` run against the new backend. Documented in POSTGRES.md.
- **Admin backup/restore routes are SQLite-only** — they return a clear "use `pg_dump`/`pg_restore`" message in Postgres mode rather than pretending to work.
- **better-sqlite3 has no Node 26 prebuilds** (pre-existing; the Docker image pins a supported Node). The sqlite contract test detects this and skips instead of failing.


perminder-klair:
closing this, as its out-dated now. sorry about this.

PR #751, opened 2026-07-02
fix(llm): detect rate limit 429s and honor Retry-After for failover

### LINKED ISSUE #738: Rate-Limiting detection and support for switching LLM providers in that case
You can get excellent use out of free tiers of several LLM providers. Most do not supply enough requests per day to run the station 24/7. The current LLM fall-back mechanism does not detect rate limiting.

Adding some logic to detect possible rate limits, together with more detailed fall-back options (e.g. more than one fall-back LLM provider/model) would allow to keep the station on air 24/7 without incurring cost.

Alternatively, setting the LLM provide and model per show would accomplish a similar feat. You could set up shows on the off hours with a free provider and your daily commute regulars on a paid provider. That way, you'd also be online 24/7 with minimal LLM cost.

geekylakshya:
Fixes #738. A plain 429 with no quota/usage-limit wording (a free-tier per-minute or per-day request cap) was previously only retried twice on the same leg with a fixed delay, then hard-failed with no attempt at the configured fallback provider. This adds isRateLimited() to trigger failover once same-leg retries are exhausted, and honors a provider's Retry-After header, capped at 30s, instead of guessing the retry delay.


perminder-klair:
I dug into this one and I don't think it's ready to merge yet. The intent is right, but as written the two new mechanisms mostly don't fire on the paths #738 cares about, and the parts that do fire add latency.

The core problem: both `isRateLimited` and `retryAfterMs` inspect the wrong error shape. Every DJ path (`djText`/`djObject`/`djAgent`) calls the AI SDK with the default `maxRetries: 2` (only `onboarding.ts` sets `0`). On a retryable 429 the SDK retries internally, already honouring `Retry-After` via its own `getRetryDelayInMs`, and then throws a `RetryError` (`name: 'AI_RetryError'`). That wrapper has no `statusCode`, `status`, `cause`, or `responseHeaders` on it. The underlying `APICallError` is tucked away on `err.errors[]` / `err.lastError`.

So against a real `RetryError`, both functions fall through to the message regex, which needs a literal `429` plus rate-limit wording. I ran them against the actual shape to be sure:

```
RAW APICallError (what the tests use):
  isRateLimited = true   retryAfterMs = 30000
RetryError w/ OpenAI "Rate limit reached for gpt-4o..." (real generateText path):
  isRateLimited = false  retryAfterMs = null
RetryError w/ Gemini "Resource has been exhausted":
  isRateLimited = false  retryAfterMs = null
```

The unit tests pass because they feed raw `{ statusCode: 429, responseHeaders: ... }` literals, but those never reach `withTransientRetry`/`withFailover` on these code paths. So for exactly the free-tier providers the issue names, there's still no failover and no header honouring.

A few more things I'd want sorted before this lands:

1. The `Retry-After` sleep runs inside `withDeadline`. `agent.ts` `runDeadlined` wraps `withTransientRetry` inside `withDeadline(timeoutMs, ...)`, so a 30s clamped sleep (plus the SDK's own internal `Retry-After` waits, which also count against the deadline) can outlive the ceiling. `withDeadline` then throws `AgentDeadlineError`, which by design matches none of the failover classifiers, so the backup leg is never tried on the primary pick path.

2. Clamp-and-retry contradicts the comment. The comment says "a longer wait should fail over", but `Math.min(ms, 30_000)` clamps a `Retry-After: 3600` down to 30s and retries the same exhausted leg twice before failover. `djText`/`djObject` have no deadline at all, so `matchRequest` inside a request or a scheduled segment can block ~60s before switching legs, versus ~2s today. A `Retry-After` beyond the cap should fail over immediately instead of sleeping.

3. Any bare 429 now triggers failover. A self-hosted llama.cpp/vLLM/LiteLLM that returns 429 for momentary concurrency (no `Retry-After`, retries exhaust in ~2s) will now switch to a paid cloud fallback for a blip that same-leg retries used to ride out. Worth gating on `Retry-After` length or wording before treating a plain 429 as a leg switch.

4. Minor: the sleep never observes the `AbortSignal`, so after a deadline abort the orphaned retry can wake up to 30s later and fire a ghost call at the very provider that asked for backoff. And the `Retry-After` branch drops the jitter, so concurrent callers with the same header all retry at the same instant.

Suggested direction: unwrap `RetryError` (`err.errors` / `err.lastError`) or key off `APICallError.isInstance(err)` so classification sees the real status and headers; bound the retry sleep by the remaining deadline (or fail over when `Retry-After` exceeds the budget); and add a test that goes through a real `generateText` call with a mocked transport rather than synthetic error objects, since that's where the current version and the tests diverge.

Happy to pair on the rework if useful.

geekylakshya:
Thanks for the detailed review, you were right on all counts.

Pushed **8655144**:

- Added `unwrapSdkError()` so every classifier and `retryAfterMs` sees through `AI_RetryError` to the wrapped `APICallError`.
- A `Retry-After` beyond a **15s same-leg budget** now throws immediately, so `withFailover` tries the backup instead of sleeping out the reset window (and stays under the **45s agent deadline**).
- A bare `429` without rate-limit wording or a `Retry-After` header no longer triggers failover, so self-hosted concurrency blips stay same-leg retries.
- The backoff sleep now respects the deadline's `AbortSignal` and keeps jitter on honoured waits.
- Tests now include a real `generateText` call against `MockLanguageModelV3` throwing a real `APICallError` to pin the wrapper shape end to end.

perminder-klair:
Thanks for the rework, this landed exactly where I hoped it would. The unwrapSdkError approach is the right call, and I really appreciate the MockLanguageModelV3 test pinning the real AI_RetryError shape. That's the test that would have caught the original gap, so it's good insurance going forward.

The bare-429 guard for self-hosted boxes and the immediate fail-over on a long Retry-After both read well too. Nice work. Thank you @geekylakshya !

PR #759, opened 2026-07-03
feat(tts): add support for kokoro multilingual capabilities

### LINKED ISSUE #725: Wrong text to speech language
I have a problem... if i set as language "Italian" in a person it will generate the text in italian, but during the text to speech it will read as the text is in english. There is a way tho fix this? I can't find an option in the admin pannel.
Thanks!

P.S.: That's a very amazing project! I LOVE IT

InspectorCaracal:
## What
Adds config options and display to allow selection from full suite of kokoro voices, along with phonemization in all of kokoro's available languages via misaki's espeak phonemizer.

## Why
One of kokoro's many advantages aside from its speed is the number of languages it can handle; not supporting it is a waste. Also should resolve #725 as it provides an Italian phonemizer.

## How tested
I tested it by hand in a development instance. Also verified no lint errors.

### DIFF
--- a/controller/scripts/kokoro_worker.py
+++ b/controller/scripts/kokoro_worker.py
@@ -60,25 +60,31 @@ def main():
     log("ready")
     emit({"id": None, "ready": True})
 
+    lang_mapping = {
+        "a": "en-us",
+        "b": "en-gb",
+        "e": "es",
+        "i": "it",
+        "f": "fr",
+        "h": "hi",
+        "p": "pt-br",
+        "j": "ja",
+        "z": "cmn",
+    }
+    # EspeakG2P construction isn't free, and there are only a handful of
+    # languages, so build each phonemizer once and reuse it across requests.
+    g2p_cache = {}
+
     def _phonemize(voice_code, lang=None):
-        """Build a language-aware phonemizer. When `lang` is explicitly
+        """Return a cached language-aware phonemizer. When `lang` is explicitly
         provided use it directly (voice timbre, accent preserved); otherwise
         auto-detect from the voice code prefix character."""
-        code_char = voice_code[0]
-        mapping = {
-            "a": "en-us",
-            "b": "en-gb",
-            "e": "es",
-            "i": "it",
-            "f": "fr",
-            "h": "hi",
-            "p": "pt-br",
-            "j": "ja",
-            "z": "cmn",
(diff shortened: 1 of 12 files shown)


InspectorCaracal (commented):
A few issues with those changes:
- misaki[en] is not necessary, it pulls in dependencies for the `en` package not the `espeak` one. "en" optional dependencies will also install torch, which by default installs cuda, drastically bloating the install with unused dependencies
- the commit includes the local pycache which is incorrect for python code

The caching improvement is good.

perminder-klair:
@InspectorCaracal thank you! will get back to this.

perminder-klair:
You're right on both, thanks for catching them. I've pushed `050f3c1` which reverts the `misaki[en]` change and drops the stray `__pycache__` file, keeping just the caching.

Good call on `[en]` pulling spacy and torch. I'd missed that the extra brings the whole English G2P stack rather than just the espeak backend. Bare misaki is the right choice.

The only thing I'd still gently raise: `from misaki import espeak` works today because `espeakng-loader` and `phonemizer-fork` come in transitively through kokoro-onnx, not because misaki itself declares them. That's fine as is, just a little implicit. If you ever want to make it explicit without the bloat, pinning those two directly (they're tiny and already installed, no torch) would document the dependency. Totally your call, not a blocker.

InspectorCaracal:
Thanks. A note regarding the voiced concern:

The versions of kokoro-onnx and misaki are pinned, so there is no reason to worry about future changes in dependencies. This is in fact THE reason for pinning dependency versions. Explicit dependencies for these specific libraries can be added *if necessary* in a future where the project updates the pinned versions to some hypothetical future versions that no longer have these dependencies.

perminder-klair:
Thanks for this, and for the back and forth on the dependencies. Fair point that pinning kokoro-onnx and misaki effectively pins the espeak stuff too, so I'll leave the explicit pins out.

I pushed a small cleanup on top (786398d2): dropped the `formatKokoroVoiceLabel` export since nothing actually imports it (the UI components have their own inline `fmt` helper), and fixed a comment in `settings.ts` that referenced a `_espeak_code()` that doesn't exist. No behaviour change, just tidying.

Really glad to have all the languages exposed instead of just the British subset. Appreciate the work.

PR #769, opened 2026-07-03
fix(app): send URL basic-auth as a header so iOS AVPlayer can stream (#764)

### LINKED ISSUE #764: http basic auth not working with iOS app
HTTP basic auth (as in using an authenticating reverse proxy that is setup to allow access via https://user:password@subwave.mydomain.net) works without problem when using a web browser (tested on iOS and Windows).

It does not work with the official app on iOS. The app shows that it connects to the server, shows a signal strength, but when starting the stream it displays "scanning the dial" and does not progress from there.

Setup
- SUB/WAVE on Unraid
- VPS running Pangolin Authenticating Proxy that tunnels directly into the Unraid box. Username and password for basic auth set (bypassing Pangolin internal authentication).
- iOS application using https://user:password@subwave.mydomain.net

perminder-klair:
Fixes #764.

## Problem

A station URL with embedded HTTP basic-auth credentials (`https://user:password@subwave.mydomain.net`, e.g. via an authenticating reverse proxy) plays fine in browsers but stalls forever on **"scanning the dial"** in the native iOS app. The app connects and shows signal strength, but the stream never starts.

## Root cause

The app stores the base URL verbatim (credentials included) and uses it two ways:

- **API polls & artwork** go through RN `fetch` and `<Image>` → NSURLSession (iOS) / OkHttp (Android), which **do** honour `user:pass@` userinfo in the URL. That's why the app connects and shows signal strength.
- **The audio stream** is handed to `react-native-track-player` → `AVURLAsset`. iOS **AVFoundation silently strips userinfo** from the URL, so the request goes out unauthenticated → proxy returns 401 → playback never fills its buffer → stuck on "scanning the dial".

## Fix

Derive an `Authorization: Basic …` header from any URL-embedded credentials and hand a **credential-free** stream URL plus that header to the track player. RNTP maps track `headers` onto `AVURLAssetHTTPHeaderFieldsKey` (iOS) and the Android DataSource — the auth path AVPlayer actually honours.

The fetch/`<Image>` paths are left untouched (they keep using the credentialed base, which already works), so the blast radius is the audio path only.

- `lib/api.ts` — `splitCredentials()` + a small UTF-8-safe `base64()` (not the Hermes-version-dependent, latin1-only global `btoa`); `streamUrl()` now returns a credential-free URL; new `streamHeaders()` returns the auth header.
- `audio/player.ts` — `LiveTrackMeta.headers` threaded into `TrackPlayer.add`. Because the headless RemotePlay reload in `service.ts` re-loads from `getLastLiveMeta()`, lock-screen/CarPlay resume inherits the header for free.
- `hooks/usePlayer.ts` — pass `streamHeaders()` at both `loadAndPlay` call sites (tune + watchdog reconnect).

## Verification

- `tsc --noEmit` clean on the changed files (the only 2 errors are a missing `expo-blur` dep in the CI/local install — unrelated, pre-existing).
- Unit-checked `splitCredentials` + `base64` against Node's `Buffer` base64: the reporter's case (`user:password` → `Basic dXNlcjpwYXNzd29yZA==`), percent-encoded credentials (`p%40ss%3Aword` → `p@ss:word`), non-ASCII credentials, and the no-credentials passthrough (base comes back credential-free, header `undefined`) all pass.

Device QA against a basic-auth-protected station still recommended (per `app/docs/QA-CHECKLIST.md`) to confirm end-to-end on hardware.


perminder-klair:
Superseded — reopening directly against `develop` to clear the stale main-branch-guard check (this PR briefly targeted main on creation).

PR #778, opened 2026-07-03
feat(admin/tagger): add option to change batch size for LLM tagging

TomiEckert:
## What
A dropdown to change the batch for LLM tagging with weaker models. Preset to 5, 10, and 25. Can be changed to a textbox if that's preferable, but this one should be fine.

## Why
Lot's of cheaper/weaker models fail to tag 25 tracks in a batch, and the tagger then defaults to 1 track per request. This way if needed the operators can fine tune the tagging, and increase the tagging performance even on weaker models.

## How tested
Built the docker container, and tagged with tiny models. 25 fails, 10 works.

## Notes for reviewers
Could be changed to use textbox instead of a dropdown menu.

### DIFF
--- a/controller/src/music/tag-library.ts
+++ b/controller/src/music/tag-library.ts
@@ -79,7 +79,7 @@ function clamp01(n: number): number {
 
 interface CliFlags {
   limit: number;
-  batchSize: number;
+  batchSize: number | null;
   seedCount: number | null;
   maxRounds: number | null;
   noPropagate: boolean;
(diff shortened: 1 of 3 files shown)


perminder-klair:
Thanks for this Tomi. Good catch on the weak-model batch failures. GitHub marked this merged automatically because your commit landed on develop through #827, where I added the one missing piece: settings.update() wasn't persisting embedding.batchSize, so the dropdown wasn't actually saving. It's all live on develop now. Appreciate the contribution.

PR #779, opened 2026-07-03
feat(admin/shows): add blacklisted playlists to shows

TomiEckert:
## What
Added an option to exclude tracks from playlists. This works by adding songs you don't want to hear in a show to a playlist, ticking the box, and they will be removed from the pool.
This overrides the playlist anchor as well. So if you select a high-energy playlist for the anchor, but don't want techno in it, you can select the techno playlist as excluded, and the pool will only have non-techno high-energy tracks.

## Why
Even finer control on what can be played on a show

## How tested
Built the docker containers, and played around with some songs and playlists. Probably needs more testing (like literally more time, to see if excluded songs come up or not).

## Notes for reviewers
More testing, and probably another set of eyes on the logic, just to make sure I didn't mix stuff up.

image

### DIFF
--- a/controller/src/broadcast/dj-agent.ts
+++ b/controller/src/broadcast/dj-agent.ts
@@ -15,7 +15,7 @@ import { z } from 'zod';
 import * as settings from '../settings.js';
 import * as session from './session.js';
 import * as picker from '../music/picker.js';
-import { resolveShowPlaylistPool } from '../music/show-playlist.js';
+import { resolveShowPlaylistPool, resolveExcludedPlaylistIds } from '../music/show-playlist.js';
 import * as library from '../music/library.js';
 import * as mix from '../music/mix.js';
 import * as journey from '../music/journey.js';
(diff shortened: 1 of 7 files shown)


perminder-klair:
## Review + follow-up commit + tested ✅

Reviewed this end-to-end, pushed a small follow-up commit (`5995f0d`) with three cleanups, and tested the whole feature against a live stack. Nice work — the coverage across all three selection paths (pool picker, DJ-agent tools, and the LLM-free `scheduler.ts` coast) is the easy thing to miss and you got it.

### Follow-up commit `5995f0d`
- **`show-playlist.ts`** — `resolveExcludedPlaylistIds` now reuses the `playlist:${id}` memo key instead of `excl:${id}`. It's the identical `getPlaylist` fetch, so a playlist used as *both* anchor and blocklist is now fetched once, not twice.
- **`scheduler.ts`** — the exclusion block's comment said "no never-starve fallback", but the `if (allowed.length)` guard *is* one. That's actually correct here (this path is the last dead-air guard, mirroring the strict-playlist block above it) — so I fixed the comment to match rather than the behavior.
- **`picker-tools.ts`** — the "`seen` is never empty" invariant no longer holds once the blocklist can empty even the guaranteed `showPlaylistTracks` source; documented that, plus that the blocklist runs *after* the playlist lock so it overrides the anchor.

### Testing
Ran the dev stack from the PR branch and verified at three levels:
- **UI** — the EXCLUDED PLAYLISTS section renders, the checkbox works, and the show summary shows the `1 excluded` chip.
- **Persistence** — `excludedPlaylistIds` round-trips correctly into `schedule.json` (right id, `playlistIds` untouched).
- **Pick-time exclusion** — an integration test against real Navidrome (temp playlist, created + deleted) confirmed `resolveExcludedPlaylistIds` returns exactly the blocklisted ids, the pick-path filter drops all blocked / keeps all allowed tracks, and an empty list is a no-op.

Both lint gates (`controller` + `web`) pass.

One minor thing I intentionally left alone: the `10` cap literals in `ShowsPanel.tsx` — they match the existing `playlistIds` picker right above, so keeping them consistent beats extracting a shared constant here.

perminder-klair:
Thanks for the PR, Tomi. Really solid work, and the comments throughout made it easy to follow the logic. Appreciate you picking this one up.

PR #784, opened 2026-07-03
feat(broadcast): upgrade Liquidsoap 2.2.5 → 2.4.4

perminder-klair:
## What

Bumps the broadcast + AIO base image from `savonet/liquidsoap:v2.2.5` to `v2.4.4` (two major series), with the script migrations 2.4 requires and one behavioral fix the upgrade uncovered.

## Migration changes

- **`protocol.add` resolver** (`subhttp:`): 2.4 resolvers return `string?` instead of `[string]`.
- **`on_metadata`**: 2.3+ requires a `synchronous:` argument. `on_meta` does file I/O (`now-playing.json`), so it runs `synchronous=false` on the task queue.
- **`cross(persist_override=true)`** — the important one. 2.4's new default (`persist_override=false`) resets the `liq_cross_duration` override *before it ever sizes the stamped track's own end-of-track buffer*. Verified with `scripts/fx-render-test.sh xdur`: a track stamped 12s rendered an 86s output (static 4s buffer — stamp ignored) without persist, 78s (stamp honoured, 2.2.5 behavior) with it. Fades inside a too-small buffer are the truncated-fade/"no crossfade" failure mode, so without this every stamped duration (washout canvas, capped-exit washouts) silently breaks.
- **Controller counterpart**: persist means a stamp lingers over later *unstamped* tracks (a 12s washout canvas would become the new permanent crossfade). `getAnnotatedUri` now stamps **every** track, falling back to the operator's configured `crossfadeDuration`, so the lingering value is always the intended one. This covers the queue and `auto.m3u` (both flow through `getAnnotatedUri`).
- Cosmetic: `null()` → `null` deprecation, renamed locals shadowing the `output`/`json` stdlib modules.

## Verification

- `liquidsoap --check radio.liq` clean on v2.4.4.
- `fx-render-test.sh probe` **PASS** on 2.4.4 (filter.rc + comb instantiate inside the cross callback, audibly alter the render).
- All five transition renders (dry / sweep / washout / both / blend) on identical inputs: **0.00 dB max RMS delta** between 2.2.5 and 2.4.4 across the transition window — the DJ effect stack renders identically.
- Full boot test of the built image with all optional encoders force-enabled: `/stream.mp3`, `/stream.opus`, `/stream.flac`, `/stream.aac` all serve; `next.txt` queue pickup; a stamped 8s crossfade fired exactly 8s before track end; `say.txt` voice duck; `now-playing.json` written (one fire per track — no double-fire regression); hourly archive file written; telnet custom `restart` command works (clean shutdown → supervisor exit for the restart policy).
- `controller npm run lint`: 0 errors.

## Still broken upstream (workarounds stay)

Re-verified on 2.4.4: native `echo` remains a **bit-identical no-op** (dry-vs-wet md5 equal, using the dB-form feedback the 2.4 API demands), and `filter.rc mode="high"` still collapses to silence (40 Hz HPF on a 440 Hz tone → −inf RMS). The comb-based washout and subtract-HPF blend workarounds remain load-bearing; comments updated to pin both versions.

## Notes

- `scripts/fx-render-test.sh` default image bumped to v2.4.4 so future harness runs match prod.
- FLAC mount note: during the empty-library boot (silence), Icecast dropped the FLAC source with a socket timeout — FLAC compresses silence to ~no bytes, so Ogg pages outlive Icecast's timeout. With real audio flowing it's stable. Content-dependent, not a 2.4.4 regression; worth knowing for silent-station debugging.
- Deploy needs `up -d --build broadcast` (prod bakes the image); dev compose bind-mounts `radio.liq` but the base image comes from the rebuilt broadcast image.


perminder-klair:
Added `f2a63f8` — fix(dj): coach blend as the exceptional transition, not the same-lane default.

On-air data from the dev-stack soak showed ~80% of the model's effect asks were `blend` (13 armed + 9 variety-stripped vs 2 sweep / 2 dissolve / 2 chosen washout over 24h). Root cause: the FLOW criterion makes nearly every pick same-lane, and the coaching said blend is *for* same-lane picks — so it was always the correct answer, and session-history imitation locked it in. All three prompt surfaces now frame blend as reserved for an exceptionally locked pair (~1 in 5), and the per-event reminder no longer lists blend first (it also never mentioned dissolve — fixed).

PR #789, opened 2026-07-03
feat(dj): dissolve transition effect + Liquidsoap 2.4.4 + blend coaching rebalance

perminder-klair:
## What

Three stacked changes from the liquidsoap-244 worktree, soak-tested on the live dev stack:

1. **Liquidsoap 2.2.5 → 2.4.4** (`5aecc1e`) — broadcast image upgrade, radio.liq adjusted for the new API.
2. **`dissolve` — fourth transition effect** (`52c9e3e`) — a diffuse, beatless reverb wash that melts the outgoing track under the incoming pick: the SMOOTH way across a clash where the sweep is the dramatic way. Completes the 2×2 effect kit (blend = rhythmic/compatible, washout = rhythmic exit, sweep = dramatic/clash, dissolve = smooth/clash). Gated by `effectAllowedFor` (only fires when the pair measurably clashes, compat < 0.4) and yields to a washout already riding the previous track's exit. Design doc included.
3. **Blend coaching rebalance** (`00ea49e`) — on-air data showed ~80% of the model's effect asks were `blend` (13 armed + 9 variety-stripped vs 2 sweep / 2 dissolve / 2 chosen washout over 24h). Root cause: the FLOW criterion makes nearly every pick same-lane, and the coaching said blend is *for* same-lane picks — so it was always the correct answer, and session-history imitation locked it in. All three prompt surfaces now frame blend as reserved for an exceptionally locked pair (~1 in 5), and the per-event reminder no longer lists blend first (it also never mentioned dissolve — fixed).

## Testing

- Running live on the dev stack from this worktree (Liquidsoap 2.4.4, dissolve armed/dropped correctly per the `[mix]` log, new coaching confirmed in outgoing LLM requests).
- `scripts/fx-render-test.sh` extended to render the dissolve envelope offline.
- `controller && npm run lint`: 0 errors.


perminder-klair:
Duplicate of #784 — opened from a stale parallel branch. The only new content (the blend-coaching rebalance) has been cherry-picked onto worktree-liquidsoap-244 as f2a63f8 and now rides #784.

PR #824, opened 2026-07-04
feat(scrobble): configurable ListenBrainz API URL for self-hosted scrobbles

sosjalapeno:
## Summary
- Allow scrobbling to self-hosted ListenBrainz-compatible endpoints (e.g. Koito) via `scrobble.listenbrainz.baseUrl` in settings or `LISTENBRAINZ_API_URL` env override.
- Default behaviour unchanged: blank baseUrl still submits to api.listenbrainz.org.
- Admin token probe and scrobble submit both respect the configured base.

## Why
Homelab operators often run Koito or other LB-compatible scrobblers on their LAN. The existing ListenBrainz backend hardcoded the public API URL, so enabling scrobbling required forking.

## Test plan
- [ ] Blank baseUrl + enabled token → scrobbles still reach listenbrainz.org (or skip if no listeners)
- [ ] Set baseUrl to a local LB-compatible instance → submit-listens receives playing_now + single listens
- [ ] LISTENBRAINZ_API_URL env overrides settings baseUrl
- [ ] Admin Settings → Scrobble → Test works against custom base
- [ ] Listener gate unchanged: 0 listeners → `[scrobble] skip: 0 listener(s)`

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -92,6 +92,12 @@ SITE_URL=
 # chat key. OpenRouter embeddings also fall back to OPENROUTER_API_KEY above.
 # EMBEDDING_API_KEY=
 
+# ───────── Scrobbling (optional — also configurable in admin UI) ─────────
+# LISTENBRAINZ_USER_TOKEN=          # ListenBrainz or LB-compatible scrobbler token
+# LISTENBRAINZ_API_URL=             # optional full submit-listens URL override
+#                                   # e.g. http://koito:4110/apis/listenbrainz/1/submit-listens
+#                                   # otherwise set base URL in admin → Settings → Scrobbling
+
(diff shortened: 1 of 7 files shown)


perminder-klair:
Thanks for this, really clean addition. The env-wins-over-settings ordering and the URL normalization are exactly right, and I like that the default path to listenbrainz.org stays untouched.

I pushed one small refactor on top (12bc861). `listenbrainzSubmitUrl()` was doing the same env+settings read and normalization as `listenbrainzApiBase()`, and it worked out to be exactly `listenbrainzApiBase() + /submit-listens` in every case, so I collapsed it to that and dropped the now-unused `LISTENBRAINZ_DEFAULT` const. Keeps the two URL helpers from drifting apart later. Lint passes.

Good to merge from my side once CI is green.

PR #844, opened 2026-07-04
fix(broadcast): Liquidsoap 2.4.5 — un-stall DJ transition effects + boot log rotation

perminder-klair:
## Summary

Since the 2.2.5 → 2.4.4 engine upgrade (#784), every crossfade using the DJ-mode transition effects (#606 sweep/washout/dissolve) stalled the stream clock — 1,378 `Latency is too high` warnings in one day (zero in the prior two months) and two `Too much latency` source resets with audible dead air.

**Root cause** (proven by isolated full-graph sandbox + bisect, not just log correlation): 2.4.4's clock scans every source in the graph per pull, and pass-through layers pull their input more than once per frame — the three stacked `smooth_add` ducking layers were the dominant multiplier (removing them cut warnings ~6×). The per-transition effect chains add ~30 operators with fan-out (dissolve's `pure_tail` = 17 consumers of one `fade.out`), so the multiplication pegs a full core mid-crossfade. The same chains in a minimal graph, or rendered offline, cost nothing.

**Fix**: Liquidsoap **2.4.5** makes clock sync-source propagation push-based/O(1). Sandbox A/B on the identical graph: dissolve worst case went from ~76 warnings/crossfade at 100% CPU to **0 warnings at ~8%**; washout and sweep also clean. `scripts/fx-render-test.sh probe` passes on 2.4.5 (filter.rc + comb still audibly process — the reason that probe exists). Effects code untouched.

Second commit: boot-time rotation for `radio.log` (>50MB → `.old`) — liquidsoap has no size rotation and the file had grown to 218MB since May.

- `cb138be` fix(broadcast): upgrade Liquidsoap 2.4.4 → 2.4.5 — un-stall DJ transition effects
- `df19da2` fix(broadcast): rotate oversized radio.log at boot

**Live since 19:35 on the production station**: 7+ effect crossfades (washouts + sweeps), 0 latency warnings, idle CPU ~9% (down from ~18-20% on 2.4.4).

## Test plan
- [ ] `docker compose up -d --build broadcast` — container reports `Liquidsoap 2.4.5`, stream comes back on air
- [ ] With djMode on, effect crossfades (`DJ sweep:`/`DJ washout:` in radio.log) produce **no** `Latency is too high` lines
- [ ] `LIQ_IMAGE=savonet/liquidsoap:v2.4.5 scripts/fx-render-test.sh probe` → PASS
- [ ] Boot with a >50MB `state/logs/radio.log` → rotated to `radio.log.old`, fresh log starts

### DIFF
--- a/docker/Dockerfile.aio
+++ b/docker/Dockerfile.aio
@@ -49,7 +49,9 @@ RUN npm run build
 FROM caddy:2 AS caddybin
 
 # ---- final: everything under the liquidsoap (Debian bookworm) base ----------
-FROM savonet/liquidsoap:v2.4.4
+# v2.4.5 minimum — same reason as Dockerfile.broadcast: 2.4.4's O(n) clock
+# scan made the per-transition DJ effect chains stall the stream clock.
+FROM savonet/liquidsoap:v2.4.5
 
 USER root
 ENV DEBIAN_FRONTEND=noninteractive
--- a/docker/Dockerfile.broadcast
+++ b/docker/Dockerfile.broadcast
@@ -13,7 +13,14 @@
 #   - A tiny shell supervisor (broadcast-entrypoint.sh) resolves the shared
 #     password, renders icecast.xml, launches both, and exits if either dies
 #     so the container's restart policy bounces them together.
-FROM savonet/liquidsoap:v2.4.4
+# v2.4.5 minimum: 2.4.4's clock scanned every source in the graph per pull,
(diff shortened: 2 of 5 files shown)


perminder-klair:
2-hour production soak on liquidsoap 2.4.5 with djMode restored: **69 crossfades, 23 effect transitions (washouts + sweeps), 0 latency warnings, 0 source resets**. On 2.4.4 the same window averaged 4–5 stall bursts/hour. Marking ready for review.

PR #853, opened 2026-07-05
Feature: Make festivals user-configurable

InspectorCaracal:
## What
Changes the Festivals definitions to be user-configurable rather than hardcoded, allowing for people to add their own festivals or change the vibe of existing ones.

## Why
The code suggests this was an intended use-case and I know I'd like to be able to add my own personal events. Birthdays, for example!

## How tested
Ran lint, tested out adding and removing festivals in a development instance to make sure they load and save properly.

## Notes for reviewers
I debated between putting this as a settings configuration page and putting it as a primary setup page like DJ personas or shows, and ultimately opted for Settings as I felt it was slightly more along the vein of defining station timezone/locale.


perminder-klair:
Thanks for this, it's a genuinely nice addition and the Settings placement was the right call.

I pushed a couple of commits to your branch (maintainer edit) to sort out some edge cases we found in review:

- Deleting every festival now stays empty instead of quietly reseeding the 13 defaults on the next restart, so operators can actually turn the calendar off.
- The active window now spans month and year boundaries, so New Year's Day with a 3 day window fires from Dec 29 and Diwali's window reaches back into October like the UI copy promises.
- Impossible dates (Feb 31, Apr 31) are rejected server side, and the day dropdown clamps when you switch months.
- The mood list is pulled from the server's SHOW_MOODS instead of a hardcoded copy, so it can't drift.
- The description you added is now rendered in the list and threaded into the DJ's on-air context, so the DJ can actually use it.

All lint and typecheck green, and I tested it end to end against a running instance. Have a look when you get a chance and shout if anything doesn't sit right.

PR #856, opened 2026-07-05
fix(webhooks): gate track.play on listener count

sosjalapeno:
## Summary
- Gate outbound `track.play` webhooks on listener count > 0 (fail-closed when count is null, matching scrobble semantics)
- Add `listeners` to the webhook payload when fired
- Update admin payload docs / example JSON

## Why
Homelab HA ambient-light automation should not run when nobody is listening. Scrobble already gates this way; webhooks did not.

## Test plan
- [ ] 0 listeners → no POST, skip log line
- [ ] 1 listener → POST with listeners field
- [ ] dj.say / request.received unchanged
- [ ] npm test in controller

### DIFF
--- a/controller/src/broadcast/listeners.ts
+++ b/controller/src/broadcast/listeners.ts
@@ -181,6 +181,17 @@ export function getListenerCount() {
   return lastCount;
 }
 
+// Fail-closed presence check: the listener count when it is known and > 0,
+// else null. The single definition of "someone is tuned in" for outbound
+// side effects (scrobbles, gated track.play webhooks) — an unknown count
+// must never fire to an empty room. Contrast djCallsAllowed() below, which
+// deliberately fails OPEN so a stats outage can't take the DJ off the air.
+export function presentListeners(): number | null {
+  return typeof lastCount === 'number' && Number.isFinite(lastCount) && lastCount > 0
+    ? lastCount
+    : null;
+}
+
 // Last known stream status, from the same 15s poll. Returns offline + 0/0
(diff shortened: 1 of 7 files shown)


perminder-klair:
Reviewed — the controller-side gating logic checked out cleanly (fail-closed semantics match scrobble, defaults/persistence are safe for pre-existing settings.json files, and the ungated path is unchanged). Thanks for mirroring the scrobble pattern.

I pushed two small commits on top rather than round-tripping review comments:

- **cc1d3a9** — the gate toggle only persisted through the hook-list Save button, which any invalid draft row disables (add a blank hook → flip the toggle → it looks applied but can never be saved). The toggle now POSTs on click like the skills toggles, and the route accepts `webhooks` / `trackPlayListenerGated` independently so neither save path drags the other along.
- **91332b8** — extracted the fail-closed listener check into `presentListeners()` in `listeners.ts`, used by both scrobble and the new gate. The two inline copies had already drifted (only yours checked `Number.isFinite`), so one definition keeps them honest.

One thing considered and deliberately left as-is: the gate is station-wide rather than per-hook. Per-hook `events` filtering already exists, so a per-hook `listenerGated` flag would be the more general shape — but for the stated HA use case the global toggle is fine, and we can revisit if someone actually needs split policies.

Lint (`eslint` + `tsc --noEmit`) passes in both `controller/` and `web/`. LGTM once CI is green.

perminder-klair:
merged, thank you @sosjalapeno

PR #861, opened 2026-07-05
feat(broadcast): "chop" DJ transition — accelerating beat-synced crossfader cut

perminder-klair:
## What

A 5th DJ transition move alongside blend / washout / sweep / dissolve: the **chop** — the classic crossfader cut. The outgoing track is gated rhythmically on its own beat while the pick rises clean through the gaps: the PERCUSSIVE way across a clash (sweep = dramatic, dissolve = smooth), the move a DJ reaches for to jump energy up off a beat-driven track.

## How it sounds

The gesture runs on its own compressed clock — **dd = min(crossfade, 10s)** — and **accelerates like a transform cut**: beat-synced pumping (a 0.45 floor under the gate keeps the first holes reading as ducks, not dropouts; gone by 25% of dd) → true full-level cuts by ~2.5s → the gate rate doubles to eighth-notes at 40% → a sixteenth-note stutter from 72% throws into the incoming pick → silent by 0.92·dd, with the rest of the cross buffer left to the incoming ramp. Gate opens AT each beat start so the downbeat transient survives (p/2 and p/4 subdivide p, so acceleration never loses the downbeat); 12 ms smoothstep edges (square edges click); engage ramp makes e=0 bit-transparent.

The first cut of the envelope ran a constant one-cut-per-beat gate over the full crossfade with every-other-beat exit stabs — heard on-air it read as slow pumping that decelerated, the wrong direction for a cut. The accelerating compressed-clock version above is the result of a live A/B.

## How it's wired

Same plumbing as its four siblings:
- **radio.liq** — chop branch in `dj_transition`, pure `amplify` driven by a closure of `source.elapsed()` (the proven sweep_gain/wash_gain pattern — no new operator class, so no phase-0 probe needed). Rides the incoming pick's `liq_chop` like sweep/dissolve, gates the outgoing branch. Yields to a washout riding the outgoing exit.
- **mix.ts** — `effectAllowedFor('chop')` clash-gated like the sweep (compat < 0.6); `chopPeriodFor` = one beat of the OUTGOING track (60/bpm clamped 0.25–0.75 s, 0.5 s default).
- **queue.ts** — validate/strip/arm in `applyMixTransition` (washout precedence, anti-streak ledger, `chopPeriod` stamped from the predecessor's BPM — the pick carries it because the predecessor's annotation is already sent; the same reason the compressed clock lives in radio.liq rather than a shorter stamped canvas).
- **dj-agent.ts / prompts/picker.ts** — chop in both pick schemas (agent + pool paths), the shared `effectsGuidance`, and the per-pick event nudge.
- **subsonic.ts** — `liq_chop` / `liq_chop_period` annotate stamps.
- **fx-render-test.sh** — chop render mode mirroring the production envelope, per the harness's keep-in-lockstep rule.

## Verification

- `npm run lint` (eslint + tsc): 0 errors; `npm run test:llm` passes.
- `liquidsoap --check` clean on the production broadcast image (v2.4.5), both envelope versions.
- Offline render (`scripts/fx-render-test.sh render … chop`, v2.4.5, 220 Hz vs 880 Hz tones): isolating the outgoing tone shows the exact envelope.
- **Heard on-air with real music**: full path exercised live — controller arm (`chop armed: 0.5s gate`), annotation (`liq_chop`/`liq_chop_period`), Liquidsoap engage (`DJ chop: … p=0.5s, dd=10.0s, d=18.0s`), through both the natural track-end and operator-skip transitions; the DJ agent also picked the chop unprompted through the production pick schema. The envelope was then re-tuned from this listening (see commit 0100e5ec) and A/B'd live against the original.


perminder-klair:
On-air tuning after live A/B (0100e5ec): the original envelope read as slow pumping over a long operator crossfade, and the every-other-beat exit stabs read as decelerating. The chop now runs on a compressed clock (dd = min(d, 10s)) and **accelerates** like a transform cut — beat cuts → eighths at 40% → sixteenth stutter throwing into the pick, true cuts by ~2.5s. Envelope constants in the PR description are superseded by the commit message. fx-render-test.sh kept in lockstep; `liquidsoap --check` clean on the production image.

PR #878, opened 2026-07-06
fix(scheduler): stop auto.m3u repeats — telnet reload + key-based dedup (#874)

### LINKED ISSUE #874: Repeated songs: auto.m3u reload watch silently dies + duplicate library copies bypass id-keyed anti-repeat
## Summary

Listeners (operator + other users) hear the same songs daily despite an ~11k library. Log analysis (Jun 28 – Jul 5, staging) shows it's **not** the picker failing to explore, and **not** embeddings. Two independent faults, both in the **`auto.m3u` fallback path**, cause the repeats — and July 5 was the worst day by far because of fault #1.

## Fault 1 — Liquidsoap silently stops reloading `auto.m3u`

The playlist uses `reload_mode="watch"`, but the watch died: on July 5 the station aired **333 auto-sourced tracks cycling only 35 distinct titles** — the same ~30-track snapshot looped for 22 hours until the 22:02 rebuild recreated the broadcast container. Every other day that week was healthy (~95% distinct titles on auto):

```
2026-06-29: auto=154 distinct=139
2026-06-30: auto=106 distinct=100
2026-07-01: auto=162 distinct=146
2026-07-02: auto=133 distinct=127
2026-07-03: auto=57  distinct=56
2026-07-04: auto=82  distinct=78
2026-07-05: auto=333 distinct=35   ← dead watch, playlist looped ~11×
```

That loop is why "Eyes on Me" aired ~11 times that day, sometimes 43 minutes apart.

It's intermittent — prime suspect: the controller writes `auto.m3u` via atomic rename (`writeFileAtomic`, `controller/src/broadcast/scheduler.ts:301`), which creates a new inode each write; if Liquidsoap's inotify watch gets orphaned once, it never reloads again until restart. This is environment-independent fragility, which would also explain **other operators reporting the same symptom**.

## Fault 2 — duplicate library copies bypass id-keyed anti-repeat

Every anti-repeat guard keys on Subsonic track id, so a library that holds the same song more than once (a very common situation for operators) defeats them: playing copy #1 doesn't block copies #2–N, because each copy is a fresh id.

On the staging library this is severe: **11,182 media_file rows but only 8,723 distinct title+artist** — 2,459 duplicate copies. "Eyes on Me" and "Drama" each exist 6× (6 different Subsonic ids); Arjan Dhillon — the #1 most-aired artist (122 plays/week) — has 8 copies of each track.

The picker's tool layer already handles this: `collect()` filters recents by lowercased `title|artist` key as well as id (May fix). But `refreshAutoPlaylistInner` (`controller/src/broadcast/scheduler.ts:151`) still filters recents and dedupes the pool **by id only** — duplicate copies sail straight into the fallback playlist.

## Ruled out

- **Exploration is fine.** Over 7 days the picker's tools returned **10,941 distinct songs** — essentially the whole library. 2,065 distinct tracks and 874 distinct artists actually aired. The "head" of most-returned songs is skewed precisely *because* duplicated tracks get counted once per copy.
- **Embeddings aren't the issue** — `tracksThatSoundLikeThis` is only ~3% of tool usage; the repeats come from the fallback path, which doesn't use embeddings at all.
- **The 12h filter itself works**: a freshly written `auto.m3u` has zero overlap with the prior 12h of plays.

## Amplifiers

- `maxTrackSeconds=180` cuts nearly every track at 3:00 (median 168s between tracks, ~550 plays/day), so the station burns through rotation ~2× faster and any narrowness is twice as audible.
- DJ picks are listener-gated (`djCallsAllowed`), so overnight/no-listener stretches always coast on the ~30-track fallback — the exact path both faults live in.

## Recommended fixes (impact order)

1. **Stop trusting inotify**: after each auto-refresh write, have the controller telnet `auto.reload` to Liquidsoap (the telnet channel already exists via `broadcast/liquidsoap-control.js`). Deterministic, fixes it for every operator.
2. **Key-based filtering in the scheduler**: `refreshAutoPlaylistInner` should filter recents and dedupe the pool by lowercased `title|artist` key as well as id — `queue.recentlyPlayed(12)` already returns `.keys`, so this mirrors what `collect()` does.
3. Optional: bump `recentPlaysMax` (300 → ~600 in `controller/src/config.ts:205`) — at 550 plays/day, 300 entries barely cover the 12h window.

Both main fixes are small controller/scheduler changes that benefit upstream users too — fix 2 makes SUB/WAVE robust against duplicate-heavy libraries regardless of how the duplicates got there.

## Verification (after fixes, 24h of runtime)

```bash
jq -r 'select(.type=="track.play") | "\(.title)|\(.artist)"' state/logs/events-*.jsonl | sort | uniq -c | sort -rn | head -20
```

Success: no track >3×/24h, none twice within any 8h window; distinct/total ratio on `src=auto` plays stays ≳90% every day.

perminder-klair:
Fixes #874.

Two independent faults in the `auto.m3u` fallback path made listeners hear the same ~30 tracks on repeat despite a large library. Both are small controller/scheduler changes that also harden the station for upstream operators.

## Fault 1 — Liquidsoap silently stops reloading `auto.m3u`
`reload_mode="watch"` relies on inotify, but the watch can orphan itself: the controller rewrites `auto.m3u` via atomic rename (`writeFileAtomic`), which swaps the file's inode on every write, and a single missed watch leaves Liquidsoap looping the last-loaded snapshot until the container restarts (the July-5 case: 333 auto plays cycling only 35 distinct titles).

**Fix:** after each write, the controller telnets the playlist's built-in `auto.reload` command (`broadcast/liquidsoap-control.ts` → new `reloadAutoPlaylist()`). Deterministic; the inotify watch stays as a backstop. Best-effort — an unreachable mixer (dev / mid-restart) logs a note and never fails the refresh.

- Verified `auto.reload` is a real telnet command in Liquidsoap 2.4.5 (the pinned image): it appears in `help` and returns `OK` for an `id="auto"` playlist.

## Fault 2 — duplicate library copies bypass id-keyed anti-repeat
`refreshAutoPlaylistInner` filtered recents and deduped the pool **by Subsonic id only**. A library holding the same song more than once (N ids for one title) defeated the guard — playing copy #1 didn't block copies #2..N.

**Fix:** filter recents *and* dedupe the pool by lowercased `title|artist` key as well as id, mirroring `collect()` in the picker's tool layer. `queue.recentlyPlayed(12)` already returns `.keys`, so this reuses the existing key format.

## Fix 3 (optional, from the issue)
Bump `recentPlaysMax` 300 → 600 (`config.ts`). At ~550 plays/day (3-min max-track churn) 300 entries barely spanned the 12h anti-repeat window.

## Verification
- `npm run lint` (eslint + `tsc --noEmit`) green — 0 errors.
- `auto.reload` empirically confirmed against `savonet/liquidsoap:v2.4.5`.
- Post-deploy runtime check (from the issue): over 24h, no track >3×/24h, none twice within any 8h window, and distinct/total on `src=auto` plays stays ≳90%/day.

### DIFF
--- a/controller/package.json
+++ b/controller/package.json
@@ -11,7 +11,7 @@
     "lastfm-session": "tsx scripts/lastfm-session.ts",
     "test:embed": "tsx scripts/embedding-preflight-test.ts",
     "test:llm": "tsx scripts/llm-pure.test.ts",
-    "test": "tsx scripts/llm-pure.test.ts && tsx scripts/picker-recency-regression.ts && tsx scripts/recent-plays.test.ts && tsx scripts/listeners-status.test.ts && tsx scripts/lastfm-enrich.test.ts && tsx scripts/web-search-searxng.test.ts && tsx scripts/request-dedup.test.ts && tsx scripts/stale-link.test.ts && tsx scripts/embedding-heavy.test.ts && tsx scripts/tagger-perf.test.ts && tsx scripts/rescan-scope.test.ts && tsx scripts/show-playlist.test.ts && tsx scripts/embedding-dim-migrate.test.ts && tsx scripts/coverage-status.test.ts && tsx scripts/tag-propagator.test.ts && tsx scripts/embedding-prefix.test.ts && tsx scripts/propagation-eval.test.ts",
+    "test": "tsx scripts/llm-pure.test.ts && tsx scripts/picker-recency-regression.ts && tsx scripts/recent-plays.test.ts && tsx scripts/auto-pool.test.ts && tsx scripts/listeners-status.test.ts && tsx scripts/lastfm-enrich.test.ts && tsx scripts/web-search-searxng.test.ts && tsx scripts/request-dedup.test.ts && tsx scripts/stale-link.test.ts && tsx scripts/embedding-heavy.test.ts && tsx scripts/tagger-perf.test.ts && tsx scripts/rescan-scope.test.ts && tsx scripts/show-playlist.test.ts && tsx scripts/embedding-dim-migrate.test.ts && tsx scripts/coverage-status.test.ts && tsx scripts/tag-propagator.test.ts && tsx scripts/embedding-prefix.test.ts && tsx scripts/propagation-eval.test.ts",
     "lint": "eslint . && tsc --noEmit",
     "lint:fix": "eslint . --fix",
     "typecheck": "tsc --noEmit"
--- a/controller/scripts/auto-pool.test.ts
+++ b/controller/scripts/auto-pool.test.ts
@@ -1,0 +1,116 @@
+// Regression test for the auto.m3u fallback pool builder (broadcast/auto-pool.ts)
+// — the recency / dedup / artist-cap guards that keep the LLM-free coast from
+// re-airing just-played tracks and from stacking duplicate library copies (#874).
+// Run: `tsx scripts/auto-pool.test.ts`.
+//
+// node:assert-via-tsx style, matching scripts/recent-plays.test.ts.
+
+import assert from 'node:assert/strict';
+import { createPoolBuilder } from '../src/broadcast/auto-pool.js';
+
+const song = (id: string, title: string, artist: string) => ({ id, title, artist });
+
+// ── duplicate library copies are deduped by title|artist, not just id ────────
+
+// A library holding the same song 6× has 6 distinct Subsonic ids for it. An
+// id-only dedup would let all 6 into the pool and the fallback would stack the
+// same track. The builder must keep exactly ONE copy (#874).
+{
+  const b = createPoolBuilder({ recentIds: new Set(), recentKeys: new Set(), targetPool: 30, maxPerArtist: 99 });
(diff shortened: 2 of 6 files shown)


perminder-klair:
### Added: unit coverage for fault 2

Extracted the fallback pool's recency / dedup / artist-cap logic out of `refreshAutoPlaylistInner`'s `take()` closure into a pure `createPoolBuilder` (`broadcast/auto-pool.ts`) — scheduler behaviour unchanged — and added `scripts/auto-pool.test.ts` (wired into `npm test`). It asserts the guards that the old id-only path lacked:

- six duplicate copies of one song (six distinct ids) collapse to **one** pool entry
- the recency filter blocks a *different* copy of a just-played song by `title|artist` key, not just by id — the exact hole duplicate copies opened
- artist cap holds across `take()` calls; per-source cap + `targetPool` ceiling
- title-less rows dedupe by id only (no empty-key catalogue collapse)

Full `npm test` suite green; `npm run lint` (eslint + `tsc --noEmit`) 0 errors.

So coverage now: fault 1 (`auto.reload`) verified live against Liquidsoap 2.4.5; fault 2 (key dedup) covered by the new unit tests.

Issue #696, opened 2026-06-30
ElevenLabs voices come out flat, can we get expressiveness controls?


mksib:
Hey, first off huuuuge thanks for SUB/WAVE! It has genuinely brought a new twist to the way I enjoy my music.

I've been running the AIO with ElevenLabs as the DJ voice, and the thing I keep running into is that the same voice and text sound noticeably flatter out of the station than if I paste the same text and choose the same voice_id in ElevenLabs site. After poking around the code I think I understand why, and it seems like a couple of controls just aren't wired up yet.

As far as I can tell from `controller/src/llm/internal/speech/cloud-speech.ts`, the ElevenLabs call only passes model, voice, text and speed — there's no `voice_settings`, so stability/style/similarity/speaker_boost all just fall back to whatever the voice has saved on my account, and there's no way to touch them from inside the app.

The second thing: even with `eleven_v3` set, the audio tags it's built around, like `[excited]`, `[laughs]`, `[whispers]` and so on, are never actually make it through, because the DJ prompt says "no stage directions / no asterisks," so anything in brackets gets stripped before it reaches the API. So v3's whole reason for existing is unusable right now.

I would like to ask if it is possible to:

- Add a way to set the ElevenLabs voice_settings (stability, style, similarity_boost, speaker_boost) from admin → TTS, sent on each request.
- Add some way to let the DJ emit audio tags when the ElevenLabs model is v3, instead of stripping them. I noticed there's already a `CHATTERBOX_TAG_HINT` that conditionally nudges the DJ to use `[laugh]`/`[sigh]` tags for Chatterbox voices, so it feels like the same trick could work here.

The first one's the important one and the second is more of a nice-to-have.

One smaller related thing while I'm here: the persona "soul" only seems to get passed as a delivery instruction on the OpenAI gpt-4o-tts path, not for ElevenLabs, so the persona doesn't really shape the delivery there either. Not asking for that specifically, just flagging it as part of the same picture.

Happy to test any of this. Thanks!

geekylakshya:
Hey @perminder-klair, I dug into this properly, and the user's diagnosis is accurate across all three points. Before writing an implementation plan, I wanted to summarize what I found and get your thoughts on a couple of design decisions.

## What the code confirms

[...]

PR #915, opened 2026-07-07
feat(tts): expose ElevenLabs voice_settings globally under tts.cloud

geekylakshya:
The AI SDK always sends its defaults for stability, style, similarity_boost and use_speaker_boost, with no operator surface to tune them. That is issue #696: an ElevenLabs voice that reads too flat or too jittery on this station has no dial.

Add the four fields to settings.tts.cloud (defaults matching ElevenLabs' own baseline) with a 0..1 clamp in both the load normalizer and the update validator, spread them into providerOptions.elevenlabs.voiceSettings on the generateSpeech call only when the resolved provider is elevenlabs, and add a matching UI block in the Cloud TTS section of the Settings panel that appears only for the elevenlabs provider. The block reuses the exact label plus tabular readout plus range-slider pattern that TtsGainField and TtsSpeedField already use, so it blends with the surrounding form.

Per-persona overrides are left as a later step per the your suggestion on issue #696: persona.tts has no model or settings slots today, so the global path fixes the reported pain with the smallest surface.

### DIFF
--- a/controller/src/audio/tts.ts
+++ b/controller/src/audio/tts.ts
@@ -165,8 +165,15 @@ async function speakWith(engine: string, text: string, opts: any, personaTts: an
   }
   if (engine === 'cloud') {
     // Persona picks provider + voice; the shared tts.cloud holds key + model.
-    const cloudOverride = (personaTts && personaTts.engine === 'cloud')
+    // `opts.cloudVoiceSettings` (preview-only, from synthesizeSample) rides the
+    // same override so "Play sample" auditions UNSAVED ElevenLabs voice_settings
+    // sliders — cloud-speech spreads the override over the saved tts.cloud, so
+    // the live path (which never sets it) keeps reading saved values.
+    const personaOverride = (personaTts && personaTts.engine === 'cloud')
       ? { provider: personaTts.cloudProvider, voice: personaTts.voice }
+      : null;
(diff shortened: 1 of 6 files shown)


perminder-klair:
Pushed 97e78a2 with three follow-ups on top of your branch:

1. **Preview now auditions unsaved sliders** — the four voice_settings ride the `POST /settings/tts/preview` body → `synthesizeSample` → `cloudOverride`, same pattern `speed` already used. So you can drag Stability and hit *Play sample* without saving first. The live on-air path is untouched (it never sets the override, so it keeps reading saved settings).
2. **Client-side defaults deduped** — the `0.5 / 0 / 0.75 / true` literals in `SettingsPanel.tsx` (form hydration + dirty-check) now come from a single `ELEVENLABS_VS_DEFAULTS` const that mirrors the server's `DEFAULTS.tts.cloud`.
3. **`eleven_v3` caveat in the Stability hint** — the model field is free-form and v3 only accepts 0/0.5/1.0 stability, which would 400 every speak call with the failure hidden behind the piper fallback. A hint is cheap insurance.

Lint (`eslint` + `tsc --noEmit`) passes on both `controller/` and `web/`. Nice, careful PR overall — the SDK camelCase mapping, the clamp-don't-throw validation, and the provider isolation all checked out against the installed `@ai-sdk/elevenlabs`.

thank you @geekylakshya as always

PR #916, opened 2026-07-07
feat(dj): add ElevenLabs v3 audio-tag hint to the system prompt

geekylakshya:
ElevenLabs v3 renders bracketed audio tags as expressive cues instead of reading them aloud as words. Today's DJ prompt has no signal for that, so v3 voices never get to use their expressive range on this station. Issue #696.

Add ELEVENLABS_V3_TAG_HINT as a direct mirror of CHATTERBOX_TAG_HINT (same syntax, same list of cues) and gate it on the RESOLVED cloud model, not the raw global model and not the provider alone. That matters because a persona that overrides cloudProvider to elevenlabs while the global provider is OpenAI ends up voiced by eleven_flash_v2_5, not v3, so gating on the persona's provider would send the hint to a v2 voice that reads the brackets literally.

Add resolveCloudModelForPersona(persona) to cloud-speech.ts (owner of CLOUD_DEFAULT_MODELS) so the resolution rule stays a single source of truth with speak(), and thread it through djSystem via an optional second parameter that defaults to the resolver. Existing callers stay zero-touch; a future preview endpoint that already resolved the model can pass it in explicitly.

The base DJ prompt template forbids asterisks and quotes but says nothing about brackets, so no rule loosening is needed here or for the existing Chatterbox hint.

### DIFF
--- a/controller/scripts/llm-pure.test.ts
+++ b/controller/scripts/llm-pure.test.ts
@@ -18,6 +18,8 @@ import { embeddingBaseUrl } from '../src/llm/internal/provider/embedding.js';
 import { DEFAULT_LOCCA_EMBED_BASE_URL } from '../src/llm/internal/provider/registry.js';
 import { personaToneDirectives, normalizeDial, DIAL_NEUTRAL, validatePersonasStrict, clampTtsSpeed, TTS_SPEED_DEFAULT, clampMaxOutputTokens, resolveMaxOutputTokens, MAX_OUTPUT_TOKENS_MIN, MAX_OUTPUT_TOKENS_MAX } from '../src/settings.js';
 import { showMusicLean } from '../src/llm/internal/prompts/picker.js';
+import { resolveCloudModel } from '../src/llm/internal/speech/cloud-speech.js';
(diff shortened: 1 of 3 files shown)


perminder-klair:
Thanks for this, the resolved-model gating was exactly the right call. I pushed a follow-up commit (7fec2d85) on top rather than round-tripping through review comments:

- The hint now also fires when a persona has no engine of its own and the station defaultEngine is cloud. That's probably the most common ElevenLabs setup, and the persona-engine check missed it.
- The resolver returns nothing when the cloud engine has no usable key. In that case speak() falls back to Piper, which would read "[laughs]" out loud, the exact bug this PR fixes.
- Gave the v3 hint its own text with ElevenLabs' verb-form tags ([laughs], [sighs], [whispers]) instead of aliasing the Chatterbox constant, so tweaking one engine's cue list can't quietly change the other.
- Factored the resolution rule into a pure resolveCloudModel() and pinned it in llm-pure.test.ts, so the "mirrors speak()" claim is tested instead of comment-enforced.

Tests and lint are green. Left the Chatterbox gate as it was, since fixing it the same way needs an llm to audio import that doesn't exist anywhere yet. Happy to adjust if you see a problem with any of this.

PR #951, opened 2026-07-08
feat(shows): multi-value genre/mood/energy/era show filters (#929)

### LINKED ISSUE #929: multiple Genre leans , moods, energies, or decades in the shows
# Feature Request: Support Multiple Attribute Selections for Show Profiles

## Problem

Currently, show profiles only allow a single selection for attributes such as **Genre Lean**, **Mood**, **Energy**, and **Decade**. This makes it difficult to accurately describe shows that intentionally blend multiple styles or musical characteristics.

## Proposed Solution

Allow multiple selections for the following fields:

* Genre Lean
* Mood
* Energy
* Decade

This would enable show creators to better represent the overall sound and identity of a show without being restricted to a single value for each attribute.

## Considerations

If these attributes are used by the recommendation engine or scheduler, multiple selections could either be treated equally or support weighted values (e.g., Primary, Secondary, etc.). I'm open to whatever implementation best fits the existing architecture.

From a UI perspective, supporting multiple selections within the existing dropdown-style picker seems like a natural fit, though I'm open to other approaches.

## Example

Instead of being limited to:

* Genre Lean: Hard Rock

A show could be described as:

* Genre Lean: Hard Rock, Alternative Rock, Metal
* Mood: Aggressive, Energetic
* Energy: High, Medium
* Decade: 90s, 2000s, or Nothing after 2010's, etc.

This would provide a more accurate representation of a show's musical identity and improve filtering, recommendations, and station programming.

perminder-klair:
Closes #929.

A show's **Genre Lean, Mood, Energy and Era** each accept multiple values now. Semantics: **OR within an attribute** (a track matching any selected value qualifies), **AND across attributes**, every value weighted equally — no primary/secondary tiers, matching the issue's "treated equally" option.

## What changed

**Schema (`settings.ts`)** — `moods[]`, `genres[]`, `energies[]`, `eras[]` (a list of `{fromYear, toYear}` windows) are the canonical stored shape. Legacy singular fields migrate to one-element lists on load — the same pattern as `dj.soul → dj.souls` — and old clients writing singular fields still validate. `eras` as a window list is what makes non-adjacent decades ("90s + 2000s") and open-ended windows ("nothing after the 2000s" = `toYear`-only) expressible.

**Shared filters (`show-filter.ts`)** — all helpers take any-of lists; `inYearRange`/`preferEra` filter against the window union; new `eraSpan()` computes the coarse single-window envelope for APIs that only take one contiguous range (Subsonic `getRandomSongs`), with the exact union enforced by post-filter. Every never-starve contract is unchanged. New pinned test file `scripts/show-filter.test.ts` (12 tests).

**Pick paths** — the pool picker and the auto-playlist refresh call Subsonic per genre (size budget split evenly) against the era envelope, then tighten to the exact union; multi-mood shows pool **all** their moods into the mood-library source (autonomous hours keep the single dominantMood). The agent's strict locks (`picker-tools.ts` / `dj-agent.ts`) are any-of lists enforced in code, as before.

**Prompts / context** — `showMusicLean` renders every entry ("lean toward Hard Rock / Metal", "prefer tracks from 1990–1999 or 2010–2019"); single-value output is byte-for-byte the legacy line (pinned in `llm-pure.test.ts`). `dominantMood` and vocal delivery (`energyForDaypart`) lead with the *first* selected mood/energy, since both need a single value by contract — documented at both sites.

**API** — `/schedule` keeps a derived singular `mood` (= `moods[0]`) alongside `moods` so the native app and older clients keep working. `/generate/show` keeps the LLM draft schema singular (kinder to weak local models) and lifts it into lists in the route.

**Admin UI (`ShowsPanel`)** — moods/eras/energy become toggle-chip rows (same visual language as the LibraryPanel energy pills), genre lean becomes an add-to-chips input with the existing autocomplete + GenreSuggest wired to append. Custom era windows set via the API render as removable chips so they can't silently constrain picks. Caps mirror the controller's limit (6 per attribute).

## Verification

- `controller` + `web` lint (`eslint && tsc --noEmit`) clean; full controller suite passes (24 test files, including the new `show-filter.test.ts` and updated `llm-pure.test.ts` multi-value pins).
- Migration smoke-tested against a temp `STATE_DIR`: a legacy singular-field `schedule.json` loads as one-element lists; multi-value strict saves round-trip; a legacy singular write still validates; an invalid mood in `moods[]` is rejected with a clear error.

### DIFF
--- a/controller/scripts/llm-pure.test.ts
+++ b/controller/scripts/llm-pure.test.ts
@@ -705,14 +705,14 @@ async function main() {
     assert.equal(showMusicLean(null), '');
     assert.equal(showMusicLean(undefined), '');
   });
-  await test('soft genre-only show is byte-for-byte the legacy line', () => {
-    assert.equal(showMusicLean({ name: 'x', topic: 'y', genre: 'Jazz' }), SOFT_GENRE_LINE);
+  await test('soft single-genre show is byte-for-byte the legacy line', () => {
+    assert.equal(showMusicLean({ name: 'x', topic: 'y', genres: ['Jazz'] }), SOFT_GENRE_LINE);
   });
   await test('filtersStrict=false leaves the soft path unchanged', () => {
-    assert.equal(showMusicLean({ name: 'x', topic: 'y', genre: 'Jazz', filtersStrict: false }), SOFT_GENRE_LINE);
+    assert.equal(showMusicLean({ name: 'x', topic: 'y', genres: ['Jazz'], filtersStrict: false }), SOFT_GENRE_LINE);
   });
   await test('strict filters are a hard rule, not a soft lean', () => {
-    const out = showMusicLean({ name: 'x', topic: 'y', genre: 'Hip-Hop', filtersStrict: true });
+    const out = showMusicLean({ name: 'x', topic: 'y', genres: ['Hip-Hop'], filtersStrict: true });
     assert.match(out, /music filters are STRICT/);                  // the unified strict lock (#766)
     assert.match(out, /Hip-Hop tracks/);                            // genre carried into the lock
     assert.match(out, /Keep your talk inside them too/);            // stay-in-filter instruction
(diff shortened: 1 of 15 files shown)


perminder-klair:
## Runtime verification (worktree dev stack, real station data)

Booted the full dev stack from the branch against a copy of the live station's `state/` (real `schedule.json` with 16 shows, real `library.db`):

- **Migration**: every show migrated to plural arrays on load; the comma-crammed legacy genre fields (`"hip-hop, electronic, big beat"`) split into individual entries (added in the second commit — they never resolved as one tag before).
- **Admin UI** (Playwright): migrated values render as selected chips; toggled a second mood + two decades + a second energy + added a genre via the input; saved; reloaded; all persisted; reverted cleanly.
- **`/schedule`**: emits `moods` plus the derived singular `mood` for older clients.
- **Live pick path**: with a multi-genre show on air, the auto-playlist refresh resolved both library genres and the dedicated source dominated the pool:
  `Auto-playlist refreshed: 30 tracks (show-genre=14 mood=1 … show=Iron & Sweat genre=Hip-Hop,Electronic energy=high)` — the unresolvable "big beat" dropped out per the never-starve contract.

Lint (controller + web) and the full controller suite (24 files) pass after the follow-up commit.

PR #952, opened 2026-07-08
feat(admin): support multi-file jingle import

DMCK96:
## Summary
- The admin "Import jingle" modal now accepts multiple audio files in one go (`<input multiple>`), instead of one file per open of the modal.
- Files upload sequentially through the existing single-file `/jingles/upload` endpoint rather than one large multipart batch — keeps server-side memory bounded and means one bad/oversized file doesn't abort the rest of the batch.
- Modal shows live progress ("Importing 12/44…") and a pass/fail summary toast at the end.
- The optional "Label" field only applies when a single file is selected; each file in a multi-file import defaults to its own filename, same as the existing single-file default behavior.

No backend/API changes — reuses `POST /jingles/upload` as-is.

## Why
Came up importing a batch of 44 pre-recorded jingles by hand — doing that one file at a time through the modal was the friction point, not anything about the upload endpoint itself.

## Test plan
- [x] `npm run lint` (eslint + tsc --noEmit) passes in `web/` with no new errors
- [x] Manual: open admin → Settings → Jingles → Import, select multiple files, confirm sequential upload + progress + summary toast
- [x] Manual: single-file import still shows the Label field and old single-file toast copy

### DIFF
--- a/web/components/admin/SettingsPanel.tsx
+++ b/web/components/admin/SettingsPanel.tsx
@@ -697,21 +697,46 @@ export default function SettingsPanel() {
 
   // Multipart upload — adminFetch leaves Content-Type unset so the browser
   // sets the multipart boundary itself. The controller transcodes + levels.
-  const uploadJingle = async (file: File, label: string): Promise<boolean> => {
-    if (busy) return false;
+  // Files upload one request at a time (not one multipart batch) so a big
+  // import doesn't hold 40+ files in memory at once server-side and a single
+  // bad file doesn't sink the rest. `label` only applies when importing a
+  // single file — each file in a batch defaults to its own filename.
+  const uploadJingle = async (
+    files: File[],
+    label: string,
+    onProgress?: (done: number, total: number) => void,
+  ): Promise<boolean> => {
+    if (busy || !files.length) return false;
     setBusy(true);
+    const total = files.length;
+    let ok = 0;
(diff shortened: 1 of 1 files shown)


perminder-klair:
Pushed a follow-up commit (2510098d) on top of this — the sequential-upload approach was exactly right @DMCK96, so this just tightens the UX around it:

- **Failed files are now listed in the modal** (name + server reason) instead of only a count in the toast, and the modal stays open on partial failure. The selection still clears so a retry can't duplicate the files that already made it in.
- **A running batch can be stopped** — Cancel becomes **Stop** mid-import (AbortController, also cancels the in-flight request), and ESC/overlay close is blocked while uploads run so the modal can't detach from a running batch.
- Minor: the single-file failure toast no longer repeats the filename, `refresh()` is skipped when nothing imported, and a typed label clears when switching to a multi-file selection.

Verified end-to-end against a live dev stack (Playwright driving the admin UI): batch with a corrupt file → `2/3 jingles imported · 1 failed` with the per-file reason listed (`ffmpeg failed: Invalid data found when processing input`); Stop mid-batch of 8 → `Import stopped — 3/8 imported`, modal stays open; ESC during import correctly refuses to close; happy multi-file path and single-file + custom label behave as before. `npm run lint` (eslint + tsc) clean.

PR #957, opened 2026-07-09
feat(observatory): WebGL galaxy renderer replaces SVG + canvas maps

perminder-klair:
## What

The Observatory map is now a single WebGL renderer — a galaxy, not zoomable dots. `ConstellationMap` (SVG) and `ConstellationCanvas` (2D canvas) are deleted along with the `CANVAS_THRESHOLD` auto-switch and the `?renderer=` override; the admin Observatory and the landing-page showcase both ride the new `ConstellationGalaxy`.

## Why

The old map wasn't browsable: zooming just made the dots bigger. And the SVG/canvas split existed purely for performance, at the cost of keeping two renderers pixel-identical forever.

## How

- **Stars**: every track is a GPU point sprite (three.js) with a soft glow skirt, plus an UnrealBloom pass. Hollow (not-analysed / uncertain) nodes render as luminous rings. Colour-by modes, filter dimming, selection emphasis all match the old renderers' rules.
- **Nebula**: an additive gaussian-splat layer over the same positions — density itself reads as cloud. Gain auto-scales with library size; the cloud thins as you fly in.
- **Semantic zoom**: star screen-size grows ∝ k^0.45 (zoom reads as flying closer, not inflating), genre constellation names fade out past k≈4, grid-decluttered track titles fade in from k≈2 (≤64, priority to selection/neighbours/confidence). Labels live in one CSS-transformed layer with a single counter-scale var — pan/zoom never re-lays-out the DOM.
- **Always night**: bloom needs a dark ground, so the stage forces a dark palette locally in both site themes; node colours below a luminance floor get lifted toward warm paper (legend swatches pass through the same lift).
- **Cost control**: render-on-demand (frames draw only on change), IntersectionObserver skips offscreen draws — the idle landing embed costs no GPU. DPR capped at 2. WebGL-unavailable browsers get a plain notice.
- **Interaction parity**: same 1000×1000 meet-fit view model, spatial-grid hover picking, SVG highlight overlay (wiring/ripple/hover ring) carried over verbatim; wheel + drag plus new two-finger pinch; zoom max deepened 6→10.

## Verified

- `npm run lint` (eslint + tsc) clean; `npm run build` passes.
- Playwright against the landing showcase (mock 800-track library): entrance renders, zoom buttons work without stealing selection, track labels appear at depth, pan keeps rendering, hover tooltip works, no console/page errors from the renderer.
- Caught + fixed during testing: pointer-capture was retargeting zoom-button clicks to the map (deselect instead of zoom), and a StrictMode remount wedged the render-on-demand guard shut (stale rAF id).

## Notes

- Positions are unchanged (genre clusters + seeded jitter) — a future embedding projection (UMAP over the stored CLAP vectors) can feed straight into this renderer for real "sounds-alike" structure.
- `three` + `@types/three` added; the renderer chunk is dynamically imported (`ssr: false`) so it never server-renders and stays out of first-load JS.
- Accessibility trade-off: the old SVG renderer's DOM nodes are gone — stars are GPU points and the label layer is `aria-hidden`, so there's no keyboard or screen-reader path to individual nodes. The search/filter panel and dossier remain the accessible route into the library.


perminder-klair:
Added option B on top: **sound-map layout**. A controller-side UMAP child projects the CLAP audio vectors to 2D (persisted as `tracks.map_x/map_y`, migration v13, auto-run on boot when stale, forceable via `POST /library/observatory/project`), and the galaxy places nodes by those coordinates when ≥60% of the library is mapped — the rail shows `PLACED BY SOUND` vs `PLACED BY GENRE`. Verified end-to-end on a real 9,172-track library (113MB DB, 100% audio-embedding coverage): projection takes ~4.5min in the child, the map forms genuine acoustic continents, and genre constellation labels sit at their true sonic centroids (greedily decluttered). Unprojected libraries and the landing-page mock keep the genre-cluster fallback.

perminder-klair:
Stress-tested this at 200k tracks (some operators run libraries that size) — the renderer holds 60 fps with an O(1) DOM at every size, but the server bulk reads were blocking the controller event loop for ~2 minutes per load at 200k. Fixes + a permanent scale suite in #969 (targets this branch): `allTaggedSampled` 98 s → 0.6 s, lean observatory rows 15 s → 1.4 s, stats-cache + ETag ordering fixes. Full numbers and the remaining known walls (synapse-link stall at load, UMAP child memory/time at 200k) in that PR's description.

PR #965, opened 2026-07-09
fix(tts): normalize display text to spoken form; never air intro fragments

### LINKED ISSUE #962: DJ intro budget can truncate spoken lines into unfinished ellipsis fragments
## What

In DJ mode, a generated `say` line can be complete in the backend/agent output, but the spoken TTS output only includes the first portion of the line followed by an ellipsis.

Example generated line:

> Ever notice how a single chord can feel like a tiny pause in the day?

Actual spoken/output line:

> Ever notice how a single chord…

This makes the on-air break sound broken rather than intentionally brief.

## Why this matters

The current behavior appears to preserve the intro timing budget by hard-trimming the generated line and appending an ellipsis. That keeps the break short, but it can create an unfinished spoken sentence.

For radio playback, silence or a complete shorter line is preferable to airing a sentence fragment. A clipped thought sounds like a TTS or station failure.

## Expected behavior

When a DJ intro line exceeds the available intro budget, Subwave should avoid airing an unfinished fragment.

Preferred behavior would be one of:

- keep the longest complete sentence or clause that fits the budget
- drop the spoken line if no complete phrase fits
- optionally expose this as a setting later, such as trim/drop/advisory

The important part is avoiding spoken fragments like:

> Ever notice how a single chord…

## Suspected area

This looks related to the DJ intro budget enforcement path.

Likely files:

- `controller/src/broadcast/dj-agent.ts`
- `controller/src/llm/internal/prompts/intro-budget.ts`

From what I can tell, the LLM generates the full `say` value correctly, but the intro budget limiter shortens it before it reaches TTS / queue playback.

## Reproduction notes

1. Run a persona/show that generates short DJ links between tracks.
2. Use DJ mode with tracks that have a short measured intro.
3. Watch the backend/agent output show a complete `say` line.
4. Observe the aired/TTS line ending as a truncated fragment with an ellipsis.

## Suggested fix

Instead of hard-cutting over-budget text and appending `…`, the limiter could:

1. prefer a complete sentence if one fits
2. otherwise prefer a complete clause/phrase ending at safe punctuation
3. otherwise return an empty string / no spoken intro

That would make the station sound intentional rather than broken when the generated line is too long for the intro runway.

## Notes

This does not appear to be a Kokoro-specific TTS issue. Kokoro is speaking the shortened text it receives. The full generated `say` field exists earlier in the pipeline.

Image
Image

### LINKED ISSUE #963: TTS should normalize display text into spoken text before synthesis
## What

Some generated radio copy contains display-oriented symbols or markup that TTS engines may pronounce literally instead of naturally.

Examples:

- `76°F` may be spoken as “seventy-six F” instead of “seventy-six degrees Fahrenheit”
- `18°C` may be spoken as “eighteen C” instead of “eighteen degrees Celsius”
- `*quietly*` or other markdown-style emphasis may be spoken with “asterisk” or otherwise rendered awkwardly
- symbols like `%`, `$`, `mph`, `&`, brackets, or decorative punctuation may not be read the way a radio listener would expect

This is especially noticeable with Kokoro, but the issue is broader than one engine: the station is sending written/display text to a speech engine without enough spoken-text normalization.

## Why

The current prompt already tries to discourage markup/stage directions, but the model can still emit symbols, weather units, markdown emphasis, or display punctuation. Other generated sources may also produce this kind of text.

Prompt rules can reduce the problem, but they cannot guarantee clean TTS input. The safer fix is to normalize booth-bound text immediately before synthesis.

For radio output, the spoken line should be written the way it should be pronounced, not the way it might appear on a screen.

## Expected behavior

Before text reaches TTS, Subwave should convert common display forms into spoken forms and strip non-spoken markup.

Examples:

- `76°F` → `76 degrees Fahrenheit`
- `76 °F` → `76 degrees Fahrenheit`
- `18°C` → `18 degrees Celsius`
- `18 °C` → `18 degrees Celsius`
- `40%` → `40 percent`
- `$12` → `12 dollars`
- `35 mph` → `35 miles per hour`
- `A & B` → `A and B`
- `*quietly*` → `quietly`
- `**RadioMania**` → `RadioMania`

## Current behavior

The TTS engine may receive the original display text and pronounce symbols literally or awkwardly.

Example generated line:

> Clear night, 76°F — the sky refuses to dim.

Possible spoken result:

> Clear night, seventy-six F — the sky refuses to dim.

Preferred spoken result:

> Clear night, seventy-six degrees Fahrenheit — the sky refuses to dim.

## Suspected area

Likely area:

- `controller/src/audio/tts.ts`

There is already a `normalizeForSpeech(...)` hook before synthesis. That seems like the right place to expand normalization so all TTS engines benefit consistently.

## Suggested fix

Expand `normalizeForSpeech(...)` into a more complete spoken-text normalizer.

Initial scope could include:

- Fahrenheit/Celsius expansion:
  - `°F` → `degrees Fahrenheit`
  - `°C` → `degrees Celsius`
- common units/symbols:
  - `%` → `percent`
  - `$` → `dollars`
  - `mph` → `miles per hour`
  - `&` → `and`
- markdown cleanup:
  - remove `*`, `**`, `_`, `__`, backticks, and similar formatting while preserving the words
- leftover decorative-symbol cleanup for symbols that should not be spoken aloud
- tests for representative examples

This should be conservative enough not to rewrite normal song/artist/title text aggressively, but broad enough to prevent common radio-copy mistakes from reaching TTS.

## Notes

There is already a default system prompt rule intended to prevent this kind of output by asking the model to write spoken words only and avoid symbols, markdown, stage directions, and non-spoken formatting.

However, prompt guidance is not a complete guarantee. The model can still emit display-oriented text, and other generated sources may also send symbols or markup into the booth path.

This issue is about adding a defensive normalization layer before TTS so the final text sent to speech engines is consistently pronounceable, even when the prompt is not followed perfectly.

perminder-klair:
Fixes #963, fixes #962 — both reported TTS speech-quality issues, plus the gaps found while tracing them.

## #963 — spoken-text normalization before synthesis

- New pure module `controller/src/audio/speech-text.ts` (no imports, unit-testable), applied at the existing chokepoint every booth-bound string already converges on: `tts.speak()` and the admin `synthesizeSample()` preview.
- Rules, all keyed on an adjacent digit where it matters (conservative by design — real artist/title text rides the same lines):
  - `76°F` / `76 °F` → `76 degrees Fahrenheit`, same for °C, bare `45°` → `45 degrees`
  - `40%` → `40 percent`, `$12` → `12 dollars` (`$5 million` → `5 million dollars`), `35 mph` / `56 km/h` expand
  - `&` → `and` (the spoken form even inside names: Florence & the Machine, R&B)
  - markdown emphasis/backticks/headings stripped, words kept (`*quietly*` → `quietly`)
  - **untouched**: `Ke$ha`, `AC/DC`, snake_case, decimals, and Chatterbox's paralinguistic `[laugh]`/`[sigh]` tags (brackets deliberately preserved)
  - the existing SUB/WAVE → Subwave rule moves here unchanged
- **Source fix too**: the weather context line rendered `24°C` into the DJ prompt itself, so the model echoed display units. It now renders `24 degrees Celsius` — the normalizer stays as the backstop.

## #962 — intro budget never airs an ellipsis fragment

`enforceIntroBudget` used to hard-cut over-budget lines and append `…`, airing fragments like *"Ever notice how a single chord…"*. New cascade:

1. keep the last **complete sentence** that fits (however short — "Nice." sounds intentional)
2. else the last **clause boundary**, closed with a period
3. else **drop the line** (`''` — callers already treat it as "no spoken intro"; the track just plays)

Boundary detection now requires the punctuation to end a word, so decimals (`3.5`) and thousands commas (`1,200`) are never cut points.

### Gaps found while tracing, also fixed

- **Pool-picker links were never budget-enforced** — only the agent path was. The pool path now runs the same enforcement (same DJ-mode gate).
- **The 2.5 words/sec assumption ignored speech rate** — a persona at 0.8× overran even "fitting" lines. The word ceiling now scales by the new `tts.speechPaceScale()` (engine × persona × daypart — extracted from `speak()`'s existing pace computation, no behaviour change to playout).

## Tests

- New `controller/scripts/speech-text.test.ts` (19 cases, auto-discovered by run-tests) pinning every conversion **and** every must-survive case.
- `llm-pure.test.ts`: the old ellipsis-fallback pin replaced with the sentence/clause/drop cascade + pace-scaling pins.
- `npm run lint` clean (0 errors), full `npm test` suite: all 25 files pass.

### DIFF
--- a/controller/scripts/llm-pure.test.ts
+++ b/controller/scripts/llm-pure.test.ts
@@ -605,8 +605,29 @@ async function main() {
     assert.equal(enforceIntroBudget('text', 20000), 'text');                         // ≥18s
     const sentences = enforceIntroBudget('One. Two. Three. Four. Five. Six. Seven. Eight. Nine. Ten. Eleven. Twelve.', 4000);
     assert.ok(sentences.endsWith('.') && sentences.split(/\s+/).length <= 10);        // last full sentence
-    const hardcut = enforceIntroBudget('a b c d e f g h i j k l m n o p q r s t', 4000);
-    assert.ok(hardcut.endsWith('…') && hardcut.split(/\s+/).length <= 11);            // ellipsis fallback
+  });
+  await test('enforceIntroBudget never airs a fragment: clause, else drop (#962)', () => {
+    // No sentence or clause boundary anywhere — the line is dropped, not "…"-cut.
+    assert.equal(enforceIntroBudget('a b c d e f g h i j k l m n o p q r s t', 4000), '');
+    // A short complete sentence beats silence even when it's well under 40%.
+    const short = enforceIntroBudget('Nice. Then a very long thought that rambles on and on without ever stopping for breath at all', 4000);
+    assert.equal(short, 'Nice.');
+    // No sentence fits but a late clause does — cut at the LAST clause
+    // boundary that fits (the longest complete thought), closed with a period.
+    const clause = enforceIntroBudget('Ever notice how a single chord, held just long enough, can feel like a tiny pause in the day', 4000);
+    assert.equal(clause, 'Ever notice how a single chord, held just long enough.');
+    // A decimal point is not a sentence boundary.
+    const decimal = enforceIntroBudget('Running at 3.5 minutes this one just keeps going and going and going without a stop', 4000);
+    assert.equal(decimal, '');
+  });
+  await test('enforceIntroBudget word ceiling scales with speech pace', () => {
+    const line = 'One two three four five six seven eight nine ten eleven twelve thirteen fourteen fifteen sixteen seventeen eighteen nineteen twenty.';
+    // 4s × 2.5w/s = 10 words at pace 1 → over budget; at 2× the whole line fits.
+    assert.equal(enforceIntroBudget(line, 4000, 2), line);
+    // At half pace the ceiling halves — still no fragment, so it drops.
+    assert.equal(enforceIntroBudget('a b c d e f g h i j', 4000, 0.5), '');
+    // Garbage pace values fall back to the historical 1.0 assumption.
+    assert.equal(enforceIntroBudget('Short line.', 5000, NaN), 'Short line.');
   });
 
   // ---- persona tone dials (humour / local colour / warmth) ----
--- a/controller/scripts/llm-pure.test.ts
+++ b/controller/scripts/llm-pure.test.ts
@@ -619,6 +619,13 @@ async function main() {
     // A decimal point is not a sentence boundary.
     const decimal = enforceIntroBudget('Running at 3.5 minutes this one just keeps going and going and going without a stop', 4000);
     assert.equal(decimal, '');
+    // An abbreviation period is not a sentence end — the DJ never airs "Dr."
+    // alone; the cut falls through to the clause/drop steps instead.
+    assert.equal(enforceIntroBudget('Dr. Dre eases us in with a slow-building intro that takes its sweet time before the beat', 4000), '');
+    assert.equal(enforceIntroBudget('Here is a smooth cut feat. a guest who drifts in over a long patient intro tonight', 4000), '');
+    // …but a real stop later in the same line still wins, abbreviation and all.
+    const abbrev = enforceIntroBudget('St. Vincent is here. Now a very long ramble that goes on and on forever', 4000);
+    assert.equal(abbrev, 'St. Vincent is here.');
   });
   await test('enforceIntroBudget word ceiling scales with speech pace', () => {
     const line = 'One two three four five six seven eight nine ten eleven twelve thirteen fourteen fifteen sixteen seventeen eighteen nineteen twenty.';
--- a/controller/scripts/speech-text.test.ts
+++ b/controller/scripts/speech-text.test.ts
@@ -1,0 +1,102 @@
+// Unit tests for the pure spoken-text normalizer (audio/speech-text.ts) —
+// the defensive layer between generated radio copy and TTS (issue #963).
+// Run: `npm test -- speech-text` (tsx scripts/speech-text.test.ts).
+//
+// Two families of pins: display forms that MUST convert to spoken forms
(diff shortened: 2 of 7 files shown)


perminder-klair:
Pushed review fixes in 912178e (the three pre-merge items from the review):

**1. Abbreviation periods are no longer sentence stops** — `enforceIntroBudget` vetoes a `.` that follows an abbreviation or lone initial (Dr., St., Mr., feat., E. …), so an over-budget "Dr. Dre eases us in…" now falls through to the clause/drop steps instead of airing **"Dr."** alone. A real stop later in the line still wins ("St. Vincent is here." trims correctly).

**2. One enforcement chokepoint, on the spoken text** — new `trimLinkToIntro()` applied inside `enqueuePick`, so every pick path (agent, pool, future producers) is covered and the duplicated gate blocks are gone. It enforces on `normalizeForSpeech(stripThinking(...))` — the raw text under-counted what actually airs (a 19-word line with `76°F`/`40%`/`$5 million`/`R&B` speaks as 25 words, +31%). The agent path still trims eagerly so its session turn records the line as aired.

**3. Currency + entity tightening** — `&amp;`/`&#39;`/`&quot;`/`&nbsp;` decode before the `&` rule (no more "and amp;"), the `&` rule skips undecoded entity shapes (`&lt;` untouched), `$100k`/`$5M`/`$2bn` expand properly (no "100 dollarsk"), "$5 million dollars" no longer doubles, magnitude words are \b-anchored ("$5 millionaire" no longer reads "dollarsaire"), and unknown glued suffixes (`$100x`) are left alone rather than mangled.

All cases pinned in `speech-text.test.ts` / `llm-pure.test.ts` — lint 0 errors, all 25 test files pass.

Still open from the review (follow-up material, not blocking): Unicode `…`/en-dash boundary glyphs in the budget cascade, glyph-less temperature forms (`76F`) + the weather segment tool's display-unit instruction, markdown links reaching TTS, and pick-time vs air-time pace staleness.

Issue #576, opened 2026-06-24
Bug(UI/UX): UI layout breaks when DJ announcement text is long


geekylakshya:
When the DJ announcement contains a large amount of text, it can overlap with the visualizer area. The content remains readable, but the layout no longer maintains proper separation between the announcement section and the visualizer.

## Expected behavior

* Announcement text should not overlap the visualizer
* Layout should adapt to longer announcements
* Consistent spacing should be maintained between content sections

## Actual behavior

* Long announcements extend into the visualizer area
* Visual elements appear crowded and less polished

## Screenshot

<img width="2671" height="1129" alt="Image" src="https://github.com/user-attachments/assets/a3caf845-a658-4ed5-a731-a4af3de99705" />

perminder-klair:
Nice catch, and thanks for the clear report and screenshot. Fixed in #589: the now-playing block now keeps to its own space above the visualizer, so long announcements get clamped instead of spilling over the bars.

Issue #723, opened 2026-07-01
Post-Audio Analysis UI slowdown


WhatKateDoes:
I encountered this before, but didnt clock the cause, and to be honest I'm still not sure I have but:

After audio analysis of library (30k tracks) is complete, the UI becomes very sluggish.  When I previously enountered this problem I started a clean install and attempted to restore backup - but at the time was unable to do, so just rebuilt cleanly.  All good, everything fine, didnt think any more of it, until last night, when I ran the audio analysis process overnight, and subsequently this morning find the UI unresponsive.  It does work, but its very laggy, multiple seconds for some pages to update.

perminder-klair:
Found the cause. The analysis run writes several large JSON blobs (beat grid, song structure, vocal ranges) directly onto each track row in the library database. A few hot paths were reading whole rows and parsing all of that on every request, the worst one being the now-playing poll every open player makes every 5 seconds. The database driver is synchronous, so on a 30k track library those reads block the whole server and everything gets laggy.

Fix is up in #771. The hot paths now read only the small fields they actually need and skip the analysis blobs entirely, plus the stats queries get a short cache. Good news is nothing in your database is corrupted, this was purely read overhead, so you won't need to rebuild once the fix lands.

PR #968, opened 2026-07-09
feat(web): now-playing UX — up-next tease, clock toggle, off-air state, cover placeholder

perminder-klair:
## What

Seven UX/UI improvements to the player's now-playing area, all frontend-only against data the player already polls:

1. **Up next tease** — the queue head (`state.upcoming[0]`) fades in under the DJ line for the last 30 s of the current track, like a real radio board. Tapping it opens the Timeline drawer. Needs a known duration to time the window.
2. **Clock toggle** — tapping the caption time readout flips elapsed (`2:31 / 4:05`) ↔ remaining (`-1:34 / 4:05`); persisted in `localStorage` (`subwave:remaining`), same pattern as the ticker pref.
3. **Off-air state** — the stage now reacts to `streamOnline === false`: "Off air" caption, muted `off air` headline, DJ line/meta/stale art hidden. Previously only the TransportBar knew; the stage kept showing a frozen "Now playing".
4. **Cover placeholder** — the art slot always renders, with a disc-mark placeholder (the `.player-mark` vocabulary at cover scale) when the track has no artwork, the image fails to load, or the stage is off air. Kills the text-column layout jump.
5. **Long-title clamp** — `line-clamp-2` + `title` attr on the h1, so a paragraph-length title can't eat the bounded stage region (the h1 flavour of #576).
6. **DJ line animation cost + reduced motion** — the type-on is now word-by-word instead of per-character (~1/6th the animated spans on long scripts), and skips the stagger entirely under OS reduced-motion or lite mode — the JS-driven tween was previously reachable by neither (`MotionConfig reducedMotion="user"` only drops transform animations; `html.lite`'s CSS `animation: none` can't touch WAAPI tweens). The artwork ripple rings (`sw-ripple-circle`) are now also hidden in both modes — they weren't in the reduced-motion block, and lite froze them as a static ring stack.
7. **Phone width** — the DotRail slims to 76 px below `sm` (tighter label tracking so SCHEDULE/TIMELINE still fit), the stage's right reserve follows (80 px), and the DJ line drops its 82 % cap on phones.

## Screenshots (stubbed controller)

| Scenario | Result |
|---|---|
| Live, long title | clamps at 2 lines, art + meta + DJ line render |
| Clock toggled | caption reads `-4:18 / 4:20` |
| Up next | `↦ UP NEXT · …` line under the DJ line |
| No artwork | disc-mark placeholder, stable layout |
| Off air | full off-air stage, placeholder art |
| Mobile 375px | slimmed rail, full-width DJ line |

## Testing

- `npm run lint` (eslint + tsc) clean — the 3 warnings are pre-existing in admin files this PR doesn't touch.
- Visual smoke test: Next dev server against a stub controller serving fixture `/now-playing` / `/state` / `/session` payloads for each scenario above (incl. the 4-poll offline confirmation path), desktop + 375 px viewports, via headless Playwright.


perminder-klair:
## Live-stack test round (real dev stack, Chrome)

Tested against the actual broadcast+controller stack from this worktree, driven in Chrome. Everything from the stubbed round re-verified live: clock toggle (`-2:16 / 3:10`, persisted across reload), DJ line showing a real spoken segment + request intro (tap opens Booth with matching transcript), cover art/placeholder, mobile layout, no console errors. The up-next tease rendered live during the last 30s of a queue-fed track (`↦ UP NEXT · PATA CHALGEA — IMRAN KHAN`).

Live testing also exposed two data gaps the stub couldn't see — fixed in 11511b0:

1. **`/now-playing` never carried `duration`** (Liquidsoap annotate metadata has none), so the clock toggle and tease would have been dark in production. The controller now recovers it from `queue.current` (full Subsonic song for requests + DJ picks), falling back to the library DB's `duration_sec` (added to the lean `TrackLite` read, still no heavy blobs per #723). Unknown tracks degrade to elapsed-only. Bonus: the waveform progress split works again for annotate-only tracks.

2. **Client clock drift**: `trackStartedAt` was stamped at "first seen by this client", so a tab hidden at the transition — or a poll catching a torn `now-playing.json` read — dragged the clock minutes behind the broadcast (observed live: `-2:59` shown with 11s actually remaining). `useStationFeed` now anchors to `/state` `current.startedAt` for the matching track (guarded against future timestamps from clock skew). Verified: a tab loaded 51s into a track read `-2:18 / 3:07` immediately.

PR #969, opened 2026-07-09
perf+test(observatory): survive 200k-track libraries — stress suite + hot-path fixes

perminder-klair:
## What

Stress-tests #957 at the scale some operators actually run (200k+ tracks), and fixes the three server-side walls the tests found. Targets the PR branch so it folds into the galaxy-renderer PR.

## Findings at 200k tracks (before)

All of these ran **synchronously on the controller's event loop** — every listener-facing endpoint (/now-playing, /request) frozen meanwhile:

| call | before | after |
|---|---:|---:|
| `allTaggedSampled(25k)` — the default cap a 200k library serves | **98.3 s** | 0.59 s |
| `allTaggedSampled(100k)` — UI hardMax | **105 s** | 1.1 s |
| `allTagged()` full scan | **14.7 s** | 1.4 s |
| `stats()` repeat call | recomputed every time | 0 ms (cache works) |
| ETag 304 revalidation | paid full `stats()` | O(1) |

## Fixes

- **Thin-window stratified sample**: the window functions ran over `t.*`, pushing every fat acoustic blob (beats/bars/structure JSON) through SQLite's partition sorter. Now windows over `(id, genre)` only and joins the full rows back. ×166.
- **Lean bulk reads**: `allTagged`/`allTaggedSampled` returned full `TrackRecord`s, JSON-parsing every acoustic blob per row for a payload that needs a pace *mean* and a vocal *presence* flag. Both now return a lean `ObservatoryTrackRow` — the fat blobs never leave SQLite. Same lesson as `getTrackLite` (#723), applied to the bulk path.
- **stats() TTL stamp**: the 5 s cache was stamped *before* the compute, so once compute > TTL every call was a miss (15 s each, and the admin panel polls it). Stamped after.
- **ETag before stats()**: a 304 no longer pays the tally scans.

## Tests

- `controller/scripts/observatory-scale.test.ts` (folds into `npm test`, `OBS_SCALE_N` tunable, default 200k): seeds a real better-sqlite3 DB with realistically fat analysed rows; budget-asserts every route call + sample proportionality/stability + payload stringify/gzip + `setMapCoordsBulk` + the client's pure layout pipeline imported from web `data.ts`.
- `web/scripts/observatory-stress.mjs` (standalone Playwright harness, not a repo dep): drives the real `/observatory` page against a mocked API at arbitrary N — load/parse/layout, long tasks, frame cadence, hover/select/filter latency, DOM-bound checks, heap.

## Browser results on the PR branch (real-GPU headless Chromium, AMD 890M, DPR 1)

| N | payload | load | zoom/pan | select | filter | DOM nodes | heap |
|---|---:|---:|---:|---:|---:|---:|---:|
| 25k | 9.3 MB (1.3 gz) | 0.6 s | 60 fps | 26 ms | 17 ms | 667 | 46 MB |
| 100k | 37 MB (5.2 gz) | 1.0 s | 60 fps | 46 ms | 73 ms | 660 | 76 MB |
| 200k | 75 MB (10.3 gz) | 1.8 s | 60 fps (worst 50 ms) | 169 ms | 239 ms | 659 | 208 MB |

The galaxy renderer itself scales beautifully — full 60 fps at 200k GPU points, and the label layer + SVG highlight overlay stay **O(1)** (~660 DOM nodes, 9 SVG elements at every size), which is the PR's design contract holding.

## Known remaining walls (reported, not fixed here)

- **`buildSynapseLinks` at load**: ~4.3 s main-thread stall at 200k (~1.2 s at 100k, negligible at the 25k default) during geometry build. Candidates: skip links above a node threshold (they're nearly invisible under bloom at that density), chunk into idle callbacks, or numeric grid keys.
- **UMAP projection child at 200k**: ~390 MB Float32 → ~1.2 GB as `number[][]` before umap-js builds its KNN graph; extrapolating the in-repo 9k≈4 min data point, hours of CPU. Worth capping projection input (stratified sample, e.g. 50k) and letting unmapped tracks fall to the genre placement the client already handles.
- **`computeStats()`** is still a multi-second scan at 200k when cold; the cache fix makes it survivable, but incremental tallies would be the real fix.

## Verified

- `npm test` — all 25 controller test files pass (scale test included, N=200k).
- `npm run lint` clean in controller + web.
- Browser sweep at 25k/50k/100k/200k: no page errors, screenshots verified.

### DIFF
--- a/controller/scripts/observatory-scale.test.ts
+++ b/controller/scripts/observatory-scale.test.ts
@@ -48,6 +48,7 @@ import {
 const N = Math.max(1000, Number(process.env.OBS_SCALE_N) || 200_000);
 const AUDIO_N = Math.min(N, 50_000); // audio vectors are 2 KB each — cap the seed, extrapolate
 const AUDIO_DIM = 512;
+const CELL_DIAG = 64 * Math.SQRT2; // buildSynapseLinks grid cell diagonal
 
 let failures = 0;
 function test(name: string, fn: () => void | Promise<void>) {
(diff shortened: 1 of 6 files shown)


perminder-klair:
Pushed the suite to **400k tracks** (and fixed what it found):

**Server** (all fine after the earlier lean-read fixes): sampled 25k read 1.1 s, 100k 1.8 s, full lean scan 2.9 s, stats 1.6 s cold/0 ms warm, coord bulk write 8 s. Full-payload-at-400k would be 152 MB raw / 20.6 MB gzipped (only reachable by raising `OBSERVATORY_HARD_MAX`).

**New wall found + fixed — `buildSynapseLinks` went super-linear**: the grid keeps the scan local but not small; dense cells degenerate toward O(n·clusterSize). 18 s main-thread stall at 400k sound-mapped, ~65 s on the genre-cluster layout. Fixed with a 96-candidate per-track scan budget, own-cell-first probe order (corner-first burned the budget one cell over and stretched links from p95 8u → 76u). Links stay local (p95 8.0 → 8.8u at 400k, identical ≤200k), sparse cells never hit the budget. Result: 400k in-browser load stall 19 s → 4.4 s, 200k 4.5 s → 2.1 s. The scale test now pins both the dense-worst-case timing and a p95 link-length locality bound.

**Browser at 400k** (149 MB payload, real-GPU headless, AMD 890M): interactive header at 1.9 s, 60 fps zoom/pan (worst frame 83 ms during the label recompute), select 215 ms, filter 346 ms, DOM constant at 657 nodes / 9 SVG elements, heap 295 MB, no crash, no page errors. (Harness note: at this size the payload must ride real HTTP — a 150 MB `route.fulfill` body kills the tab via the CDP protocol; the harness now serves it from a side server via redirect.)

**UMAP projection at 400k remains out of reach**: ~781 MB Float32 → ~2.3 GB as `number[][]` before umap-js builds its KNN graph — the cap-the-projection-input recommendation stands, stronger.

PR #977, opened 2026-07-10
feat(broadcast): configurable Icecast listener cap (ICECAST_MAX_CLIENTS)

perminder-klair:
Templates the Icecast `<limits><clients>` value (was a hard-coded 100) behind a new `ICECAST_MAX_CLIENTS` env var.

## Why
- Small hosts want to bound bandwidth by capping concurrent listeners.
- The hosted SUB/WAVE offering positions stations as private (personal, ~15 listeners); this makes that cap enforceable at the Icecast level.

## Behaviour
- Empty/unset → `100` (the historical default), so existing installs are unchanged.
- Rendered by the broadcast entrypoint via the same `sed` pass as the Icecast passwords (numeric value, no escaping risk).

## Changes
- `docker/icecast.xml.template` — `<clients>${ICECAST_MAX_CLIENTS}</clients>`
- `docker/broadcast-entrypoint.sh` — default + sed substitution
- `docker-compose.yml` — passes `ICECAST_MAX_CLIENTS` through to the broadcast service
- `.env.example` — documents the var

Verified the render defaults to 100 when unset and honours an explicit value (e.g. 15).


perminder-klair:
Reopening against develop.

PR #1018, opened 2026-07-12
fix(shows): wire excluded playlists into pick paths + make strict filters actually strict

perminder-klair:
## What

Fixes three related field reports from Discord about show music constraints not being honoured:

1. **"Excluded playlist song just played on my show"** — the playlist blocklist (#779) was a **global no-op**: `settings.resolveActiveShow()` never copied `excludedPlaylistIds` onto the resolved show object, so `resolveExcludedPlaylistIds()` always received `undefined` on every pick path (DJ agent, pool picker, auto.m3u fallback) and returned null. Broken since the feature shipped. Fixed in the resolver, pinned by the new `resolve-show.test.ts`.

2. **"Era set to 2020s with strict on, majority of tracks not from 2020s"** — the never-starve fallback in strict mode ran **per source / per tool call**: any discovery source containing zero in-filter tracks passed its *entire unfiltered result* through (`preferEra`/`preferGenre`/`preferMood` contract). The similarity tools cluster on the current track, so once one off-era track aired they returned 0 in-era matches and flooded the candidate set — "strict" was held only by prompt compliance. Now:
   - **Agent tools**: strict locks are HARD per tool (new `onlyGenre`/`inYearRange`/`onlyMood`/`onlyEnergy`), mirroring how `playlistLock` already worked. A tool with no matches returns an empty result steering the model to another tool; a fully-starved run fails into the pool fallback.
   - **Pool picker**: strict filters are additionally enforced on the FINAL merged pool, which is where never-starve now lives — a genuinely starved constraint still degrades to the full pool rather than dead air (same scoping as the existing strict-playlist block).
   - The auto.m3u fallback already hard-filtered genre/era; unchanged.

3. **"Playlist only (strict) show played 13/13 off-playlist songs"** — the enforcement code is genuinely hard on all paths, but `resolveShowPlaylistPool()` swallows every fetch failure with bare `catch {}`. A stale playlist id (playlist deleted/recreated in Navidrome) or a Navidrome error silently un-anchors the show: no lock, no `showPlaylistTracks` tool, zero log output — while the system prompt still told the agent to call the unregistered tool. Now all three pick paths log an operator-facing warning when a pinned anchor resolves to nothing, `show-playlist.ts` warns per failed fetch, and the prompt's playlist steer is gated on the pool actually resolving.

## Verification

- `npm test` — all 31 test files pass, including the new `resolve-show.test.ts` (fails without the settings fix) and extended `show-filter.test.ts` hard-filter pins.
- `npm run lint` / `tsc --noEmit` — clean.

## Notes for the reporters

- The excluded-playlist and strict-era reports are fully explained by (1) and (2).
- The 13/13 strict-playlist report is most likely (3) — worth asking Pezzles to re-select the playlists in the show editor (stale ids) and check the DJ log for the new warning after upgrading.

### DIFF
--- a/controller/scripts/resolve-show.test.ts
+++ b/controller/scripts/resolve-show.test.ts
@@ -1,0 +1,91 @@
+// Regression pins for settings.resolveActiveShow — the resolved-show object is
+// what EVERY pick path consumes (dj-agent, pool picker, auto-playlist refresh),
+// so a field the resolver drops is a feature silently disabled station-wide.
+// The canonical case: excludedPlaylistIds was validated + persisted but never
+// copied into the resolved show, so resolveExcludedPlaylistIds() always saw
+// undefined and the #779 playlist blocklist was a global no-op.
+//
+// Run: npm test -- resolve-show
+
+import assert from 'node:assert/strict';
+import { resolveActiveShow } from '../src/settings.js';
+import { setStationTimezone } from '../src/time.js';
+
+let failures = 0;
+async function test(name: string, fn: () => void | Promise<void>) {
+  try {
+    await fn();
+    console.log(`  ✓ ${name}`);
+  } catch (err) {
+    failures++;
+    console.error(`  ✗ ${name}\n    ${(err as Error).message}`);
+  }
+}
+
+// Deterministic wall clock: pin the station zone and resolve at a fixed
+// instant — Saturday 2026-01-03 19:00 UTC (dow 6, hour 19).
+setStationTimezone('UTC');
+const at = new Date(Date.UTC(2026, 0, 3, 19, 0, 0));
+
+const schedule: (string | null)[][] = Array.from({ length: 7 }, () => Array(24).fill(null));
+schedule[6][19] = 's1';
+
+const settings = {
+  schedule,
+  shows: [
+    {
+      id: 's1',
+      name: 'Jazz Hour',
+      topic: 'smooth jazz for the weekend evening',
+      personaId: 'p1',
+      moods: ['calm'],
+      genres: ['Jazz'],
+      eras: [{ fromYear: 1960, toYear: 1979 }],
+      energies: ['low'],
+      filtersStrict: true,
+      playlistIds: ['pl-anchor-1', 'pl-anchor-2'],
+      playlistStrict: true,
+      excludedPlaylistIds: ['pl-blocked'],
+    },
+  ],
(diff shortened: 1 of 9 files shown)


perminder-klair:
Follow-up commit `faeea67` — hardens the strict-filter work after a deep review of the branch turned up a cluster of bugs the `prefer*` → `only*` switch introduced (the hard filters removed a never-starve net several paths still relied on, and it wasn't applied uniformly).

**Unified onto one shared helper** — `applyStrictLocks(tracks, locks, {starve})` in `show-filter.ts` is now the single source of truth for "make this pool strict", used by both pick paths and the auto-playlist coast so they can't drift.

Fixes:
- **Pool picker all-or-nothing revert** — one zero-coverage tag class (e.g. a mood on an un-tagged library) discarded an otherwise genre/era-pure pool and leaked off-filter tracks back in. Now a per-dimension cascade: a starved dimension is skipped, the rest stay pure.
- **Agent path starved the whole show** on an unresolvable genre or an un-analysed library — every tool emptied, tripping the breaker with a misleading "model can't handle tools" log. Genres now resolve to library tags (drop-out on no-match); mood/energy locks are coverage-gated on the library's `byMood`/`byEnergy` vocab. Resolved once in `pickViaAgent` and threaded through `run()`.
- **`showPlaylistTracks` returned a bare `[]`** when strict locks emptied it while the prompt mandated that tool — the documented fabrication trigger. Now returns an `emptyResult` note with the "never invent an id" rule.
- **Auto-playlist coast leaked strict mood/energy** (LLM-free fallback had no final-pool pass). Added.
- **`repickFromSeen`** now gets the real `playlistResolved` + `showAt`, so the tool-less salvage prompt doesn't reinstate the playlist lean when the anchor never resolved.
- **`inYearRange` `Number(null)===0` trap** — un-yeared tracks (and `TYER=0000`) passed an open-lower-bound era window ("1989 and earlier"). Now treats null/''/non-positive year as unknown.

Tests pin the open-lower-bound era hole and the cascade in both starve modes. Full suite (31 files) green, lint 0 errors.

**Deferred (not in this commit) — needs a product call:** listener requests still bypass the excluded-playlists blocklist (`requestAgent` passes no `excludedIds`). The request path is deliberately exempt from the strict *genre* lock ("explicit listener ask wins"), but a blocklist of explicit tracks arguably should apply even to requests. Left out so it's a conscious decision rather than bundled into the refactor.

Issue #749, opened 2026-07-02
DJ-mode adaptive blend stamps liq_cross_duration on the wrong transition (off-by-one)


perminder-klair:
## What

`applyMixTransition` (controller/src/broadcast/queue.ts) computes the feature-1 adaptive blend for the transition **prev → item** — `crossSecondsFor(cur, next)` weighted by their tempo/key compatibility, capped to *item's* instrumental intro — and stamps the result on `item.track.crossSec` → `liq_cross_duration` on **item**.

But a track's `liq_cross_duration` governs the crossfade at its **own end**: Liquidsoap's `cross` sizes its buffer from the *outgoing* track's override, and `dj_transition` reads `a.metadata` for the fades (the fade == buffer invariant). So the value computed for prev→item actually rules **item→next**.

## Evidence

Verified empirically against the real `savonet/liquidsoap:v2.2.5` image — `scripts/fx-render-test.sh xdur` (added in #606's rework): a track stamped `liq_cross_duration=12` against a cross default of 4s produced a **12s buffer at its own end** (output duration 78s = 50 − 12 + 40; the transition callback logged d=12 for a→b and d=4 for the unstamped b→end).

## Impact

Every DJ-mode blend length is keyed to the wrong pair: the compatibility maths for prev→item lands on item→next. Magnitudes are similar on average, so it's subtle — but the intro cap (sized to *item's* intro so the fade-in completes before its vocals) protects the wrong transition entirely.

## Why it isn't a one-line fix

The correct value for item→next isn't known when item is annotated (the next pick doesn't exist yet). Fixing it needs a different signalling channel — e.g. a small file the controller writes when the pair is known and `cross`'s duration getter reads at buffer time, or accepting one-transition-late semantics explicitly.

Found during the #606 transition-FX rework; see `docs/superpowers/specs/2026-07-02-dj-transition-fx-rework-design.md` (cross-duration physics section). The washout canvas in #606 is unaffected — it deliberately stamps the flagged track itself, which is exactly the transition the wash fires on.

PR #1057, opened 2026-07-16
feat(analysis): on-pick "a la carte" track analysis (discussion #1032)

perminder-klair:
Implements the request in discussion #1032: instead of requiring bulk analysis runs upfront, analyse a track **at the moment the DJ queues it for air**, so its first spin already mixes with fresh data and the library backfills organically with what actually gets airplay.

## How it works

- New setting **`audio.analyzeOnPick`** (default **off**) with an Enable/Pause row in the Library panel's Acoustic analysis section.
- In `queue.drainToLiquidsoap()`, right before `applyMixTransition` reads the library record, the item's track is analysed if it still needs anything: stale/missing bpm/key (`analysis_version`), a missing CLAP "sounds-like" vector, or missing vocal ranges — each opt-in dimension only when it's wanted (env/admin toggle) *and* the backend can produce it, mirroring the bulk pass's gating.
- **Bounded by a 60s deadline**: past it, the hand-off proceeds with whatever data exists (exactly today's behaviour) while the analysis keeps running in the background and still caches its result for the track's next spin. Awaiting in the drain loop matches the existing awaited TTS render — it's the queue's one slow-work seam, a full track away from when the item airs.
- Skips cleanly when: the track isn't in `library.db` yet (`upsertTrackAnalysis` is an UPDATE), a bulk tagger/analyzer run owns the backend (pidfile check via `tagger-lock`, avoiding the `tagger ↔ queue` import cycle), or no analysis backend is up.
- Single-flight per track id — a listener request and a pick racing to the same song share one analysis.
- A stored CLAP vector triggers a best-effort zero-shot audio-mood scoring pass, same as the bulk path.

## Refactor

The per-track analyse+store core (analyze → `upsertTrackAnalysis` → vector + provenance stamp) is extracted from `runAnalysisPass`'s loop into a shared `analyzeAndStore()`, so the bulk pass and the on-pick path share one implementation. The provenance stamp guard moves from per-run to per-process — idempotent row, no behaviour change.

## Testing

- `npm run lint` clean in both `controller/` and `web/` (eslint + tsc).
- Feature is off by default; when off, zero new code runs in the queue path beyond one settings read.

### DIFF
--- a/controller/src/broadcast/queue.ts
+++ b/controller/src/broadcast/queue.ts
@@ -11,6 +11,7 @@ import * as subsonic from '../music/subsonic.js';
 import * as mix from '../music/mix.js';
 import * as library from '../music/library.js';
 import * as blocklist from '../music/blocklist.js';
+import { analyzeOnPick, analyzeOnPickEnabled } from '../music/analyze.js';
 import { speak, voiceGainDb } from '../audio/tts.js';
 import * as djAgent from './dj-agent.js';
 import * as programme from './programme.js';
@@ -956,6 +957,27 @@ class Queue {
         // while we were awaiting the TTS render above — don't hand a removed
         // track to Liquidsoap.
         if (!this.upcoming.includes(item)) continue;
+
+        // On-pick "a la carte" analysis (discussion #1032): when enabled and
+        // this track still needs analysis, analyse it NOW — before the mix
+        // transition below reads the library record — so this very hand-off
+        // gets the fresh bpm/key/outro/vocal/loudness data. Bounded by the
+        // deadline inside analyzeOnPick: past it the item drains with whatever
+        // data exists (today's behaviour) and the analysis keeps running in
+        // the background, caching its result for the next spin. Awaiting here
+        // matches the TTS render above — the drain loop is the queue's one
+        // slow-work seam, a full track away from when this item airs.
+        if (analyzeOnPickEnabled() && item.track?.id) {
+          const got = await analyzeOnPick(item.track.id);
+          if (got === 'analyzed') {
+            this.log('mix', `on-pick analysis cached → ${item.track.title}`);
+          } else if (got === 'pending') {
+            this.log('mix', `on-pick analysis still running — ${item.track.title} airs with existing data; result caches for next time`);
+          }
+          // Same guard as after the TTS await: an operator cancel may have
+          // spliced this item out while we were analysing.
+          if (!this.upcoming.includes(item)) continue;
+        }
 
         // DJ-mode mixing (features 1 & 2): shape the transition INTO this track
         // from its tempo/harmonic compatibility with the track it follows. The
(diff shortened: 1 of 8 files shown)


perminder-klair:
Closing this rather than rebasing it. It's 402 commits behind `develop` and the drain path it hooks into has changed shape three times since July, each time in a direction that works against what this branch does.

**Why a rebase wouldn't land it:**

- **#749 pair-aware drains** hold a pick unsent until its successor is queued. `drainAction` returns `hold` and breaks out of the loop *before* the point this branch inserts its analysis, so every item that reaches the new code has an unsent successor — which sends it down the `othersWaiting` branch and backgrounds the analysis. The awaited path, the one that actually made the current hand-off use fresh data, is unreachable under today's defaults.
- **#1409** added a render budget to the awaited TTS render in that exact spot (`introRenderBudgetSec(remainingUntilItemAirs(item))`), on the rule that music commitment must not sit behind optional slow work. This branch cites that same await as its precedent for a fixed 60s block — the precedent is now the counter-example.
- **#1099** established that analysis stands down while anyone is listening. On-pick analysis runs, by construction, at a listener-facing hand-off, and consults nothing. An operator with `analyzeQuietOnly` and `analyzeOnPick` both on gets two modules making opposite calls on the same single-threaded backend.

**Four of the eight files are already superseded on `develop`:**

- `library.ts` — the `keyRanges` projection fix landed independently, and the projection is now pinned by `scripts/silence-trim-library.test.ts`.
- `library-db.ts` — `hasAudioVector` landed independently in `library-db/vectors.ts` (with a `LIMIT 1` this version lacks); the file itself is now a barrel over 14 modules.
- `LibraryPanel.tsx` + `LibraryTaggingPanel.tsx` — the `saveAudioToggle` refactor is already done as `library/useTaggerControls.ts` + `useAdminMutation`.

The `analyzeAndStore` extraction is also based on a loop body that no longer exists: `runAnalysisPass` has since grown lyric-first vocal ranges (#1125), stem-cache budgeting (#1257) and embedding-only analysis, none of which the extraction knows about.

**Discussion #1032 is still open and unanswered.** Nothing on `develop` prioritises analysis by what actually airs, so the request stands — this is a close on the implementation, not on the idea.

**What the answer probably wants instead:** a low-priority background analyser that takes its scope from the `plays` table / `music/airing.ts` history, so the library backfills toward what actually gets airplay, and that runs behind the existing quiet gate rather than around it. Nothing on the pick path should block, and nothing should sit inside `drainToLiquidsoap` — it's the single writer of `next.txt`, and it holds `senderBusy` for every hand-off including listener requests.

**One fix salvaged:** `36447b8a` — `downloadCapped` was leaving its partial `.audio` in `analyze-tmp` on every failure path past `createWriteStream`. That was real and independent of the rest of this branch, so it's on its own branch with a test.

perminder-klair:
Salvaged as #1547.

Issue #660, opened 2026-06-27
blank.skip causes 100% CPU spin when no music is in queue


naranco66:
Environment: Podman, LXC container, fresh install

Problem: On a fresh install (pre-onboarding, no music queue), 
liquidsoap enters a 100% CPU busy-loop and stops feeding 
Icecast → /stream.mp3 returns 404.

Root cause: blank.skip(max_blank=5.0, radio) in radio.liq 
(line ~484) spins when the music chain has no available source.
The emergency fallback audio is NOT silent (-16dB mean), 
so the blank detector isn't the trigger — it's the skip 
loop when music is unavailable.

Bisection evidence:
- single(emergency) → icecast: OK (no spin)
- + smooth_add: OK
- + cross: OK  
- full production graph: SPIN
- full graph minus blank.skip: OK ← culprit isolated

Workaround: comment out blank.skip in radio.liq

The spin also recurs whenever the queue empties at runtime 
(Navidrome down, playlist exhausted).

perminder-klair:
nice find! thank you, will fix this in next release.

perminder-klair:
Fix is up in #665. Turns out blank.skip busy-loops at ~100% CPU any time the music chain is empty, and it does it in any position in the graph, so reordering didn't help and I just pulled it out. The emergency fallback already covers a dead source, which was really the only thing it was protecting against. Verified against an empty library and CPU drops from ~100% to around 6%. Thanks for the clean bisection, made this easy to pin down.

PR #1066, opened 2026-07-16
feat(broadcast): idle-pause the programme when nobody is listening

perminder-klair:
## What

Option B of the empty-room discussion: when nobody is listening, pause the *programme* instead of the LLM alone — while keeping the Icecast mounts up so any client can tune back in and wake it.

- **radio.liq idle gate** — an `idle_mode` ref + `switch` to `blank()` inserted after the emergency fallback. When idle, the whole chain above (dj_queue, auto playlist, cross buffer, subhttp downloads from Navidrome) stops being pulled and freezes mid-track. New telnet commands `idle_on` / `idle_off` / `idle_status`, idempotent, logging only on state change. A plain `blank()` branch is safe — the #660 CPU spin was specific to `blank.skip`.
- **`broadcast/stream-idle.ts` monitor** — 5s tick. Live: reads the 15s listener monitor's cached count and arms an empty-clock; after `idleAfterMinutes` of continuous zero it pauses. Idle: forces a fresh Icecast poll each tick, so the first listener (VLC, Sonos, web player — anything that connects to the still-up mounts) resumes the programme within ~5s, exactly where it froze. Fail-open: an unknown count never holds the station silent. Adopts the gate state from Liquidsoap on controller boot and re-asserts idle after a mixer restart (which always boots live).
- **DJ silenced while idle** — `djCallsAllowed()` returns false during the pause regardless of `llm.pauseWhenEmpty`, so no say.txt/intro.txt WAVs pile up against the un-pulled voice queues and dump on resume.
- **Settings** — `stream.idleWhenEmpty` (default **off**) + `stream.idleAfterMinutes` (default 10, 1–1440). Applied live over telnet; no mixer restart, no liquidsoap_*.txt boot file. Admin → Settings gets an “Idle pause” card next to Broadcast; `GET /state` exposes `streamIdle` so clients can tell “empty room” from “broken”.
- Pure transition logic in `stream-idle-pure.ts` (fail-open / re-assert / empty-clock-reset branches), pinned by `scripts/stream-idle.test.ts` — same split as `programme-pure.ts`.

## Not covered

- `POST /stream-stop` (`stream_off`) is untouched — that remains the operator's hard off-air switch that tears the mounts down.
- Manual operator actions (`/dj/*` runners) intentionally stay un-gated, matching the budget-gate convention; a WAV spoken mid-idle airs on resume.
- now-playing.json is left as-is while idle: the frozen track is what resumes, so the stale title is correct the moment audio returns.

## Verification

- `npm run lint` clean in `controller/` and `web/`; `npm test` — all 34 controller test files pass (incl. the 11 new transition cases).
- `liquidsoap --check` passes on the edited radio.liq (broadcast image).
- Runtime smoke test in a throwaway broadcast container: `idle_status→off`, `idle_on→OK`, `idle_status→on`, `idle_off→OK`; log shows `idle_gate` switching to `idle_silence` and back with transitions.

Prod note: the radio.liq change ships with the broadcast image — needs `up -d --build broadcast` (or a pulled image) like any mixer change.

### DIFF
--- a/controller/scripts/stream-idle.test.ts
+++ b/controller/scripts/stream-idle.test.ts
@@ -1,0 +1,110 @@
+// Unit tests for the pure idle-pause transition helper in
+// broadcast/stream-idle-pure.ts. Run: `tsx scripts/stream-idle.test.ts`
+// (folded into `npm run test`).
+//
+// nextIdleState decides when the programme idle-pauses (zero listeners for
+// the configured window) and when it resumes. The branching is
+// regression-critical: too eager and a zero-blip between listeners silences
+// a live station; too lax and the wake-on-connect never fires, leaving a
+// tuned-in listener staring at silence. node:assert-via-tsx style, matching
+// listeners-status.test.ts.
+
+import assert from 'node:assert/strict';
+import { nextIdleState, type IdleState } from '../src/broadcast/stream-idle-pure.js';
+
+let failures = 0;
+function test(name: string, fn: () => void | Promise<void>) {
+  return Promise.resolve()
+    .then(fn)
+    .then(() => console.log(`  ✓ ${name}`))
+    .catch((err) => { failures++; console.error(`  ✗ ${name}\n      ${err?.message || err}`); });
+}
+
+const LIVE: IdleState = { idle: false, zeroSince: null };
+const IDLE: IdleState = { idle: true, zeroSince: null };
+const AFTER = 10 * 60_000; // 10 min window
+const T0 = 1_000_000;
+
+async function main() {
+  console.log('nextIdleState (idle-pause transitions):');
+
+  await test('disabled toggle is inert while live', () => {
+    const r = nextIdleState(LIVE, { enabled: false, count: 0, now: T0, idleAfterMs: AFTER });
+    assert.equal(r.action, null);
+    assert.deepEqual(r.state, LIVE);
+  });
+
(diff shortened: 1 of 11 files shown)


perminder-klair:
**End-to-end verification** (throwaway broadcast container from this branch's radio.liq + the worktree controller against it, `idleAfterMinutes: 1`, tracks fed through the real `next.txt` → `dj_queue` path):

1. Track playing, room empty → controller idle-paused mid-track right on schedule (djLog: *programme idle-paused — no listeners for 1 min*), `idle_status` → `on`, radio.log shows `idle_gate` switching to `idle_silence`, `GET /state` → `streamIdle: true`.
2. Connected a plain HTTP client with a VLC user-agent to `/stream.mp3` **while paused** — connected instantly (mount never went away), and the programme resumed **2 s** later (djLog: *programme resumed — listener connected*).
3. Captured audio across the resume: music → 15 s silence (the pause) → music again at −3 dB peaks, resuming mid-track (ffmpeg silencedetect/volumedetect).
4. Repeated pause/resume cycles in one session (5 pauses logged) — no stuck states.

One rig artifact worth noting for anyone reproducing: a **fresh empty-library boot broadcasts silence rather than the emergency loop even on develop** (verified with a stock radio.liq control container) — the never-fed music chain reports ready-but-blank so the emergency fallback doesn't engage. Pre-existing, unrelated to this PR.

PR #1071, opened 2026-07-16
fix(scheduler): stamp loudness gain on auto.m3u fallback tracks

mksib:
## What
Makes the `auto.m3u` fallback apply the same per-track loudness correction the normal queue does, so tracks played from it aren't louder than everything else.

## Why
I noticed on my station that a track, that was playing when i connected, was way too loud, and then the next one was gain corrected. Turns out the loudness/ReplayGain gain only gets stamped in the queue drain path (`applyLoudnessGain`). The `auto.m3u` fallback writes tracks straight through `getAnnotatedUri` and never calls it, so anything that plays off the fallback (queue's empty because the LLM is off when no one listens (my case) or down or slow, budget cap hit, or right after a restart) just plays at unity gain.

## How tested
Ran `npm run lint` in `controller/` and it passes. Confirmed the fallback pool now goes through the same `applyLoudnessGain` as the drain path before the M3U gets written, so both honor the `settings.loudness.source` setting.

### DIFF
--- a/controller/src/broadcast/scheduler.ts
+++ b/controller/src/broadcast/scheduler.ts
@@ -357,6 +357,18 @@ async function refreshAutoPlaylistInner() {
     if (allowed.length) { pool.length = 0; pool.push(...allowed); }
   }
 
+  // Loudness normalisation: the queue drain stamps liq_amplify per track, but
+  // this fallback bypasses the drain entirely — without a stamp here every
+  // coast track (queue empty: LLM down/slow, budget-hard, restart) played at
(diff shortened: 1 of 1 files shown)


perminder-klair:
Thanks for this, good catch. I ran it in an isolated harness against a stubbed Subsonic backend: on develop the fallback lines carry no liq_amplify at all, with your patch all 30 come out stamped, including the getSong recovery for tracks that don't ship the tag in the API response. The gains match what the drain path produces, so both routes level the same now. Will merge once the lint run goes green.

Issue #997, opened 2026-07-12
Bug: station ID talks over the DJ & no way to disable


joehand:
Got a station-ID announcement ("you are listening to subwave...") that talked over the actual DJ's actual into. Need to guard against overlapping the DJ's own between-track talk.

Is there a way we can disable this? It throws the flow quite a bit and voice doesn't match with any others.

perminder-klair:
Thanks for the report, and good catch. Dug into this and the voice you heard wasn't actually the DJ's station ID (those are serialized so they can't talk over each other). It was the built-in jingle stinger, which plays from inside the audio mixer where the DJ's voice scheduler can't see it. That's why it landed on top of the intro, and why the voice sounded like nobody else on the station: it's a pre-rendered clip, not your DJ's voice.

Fix is up in #1004. Two parts:

1. The mixer now tells the controller when a jingle is playing, and the DJ holds any spoken segment until it's finished.
2. You can turn jingles off completely by setting the jingle ratio to 0 in admin settings (needs a mixer restart to apply).

Will land in the next release.

PR #1072, opened 2026-07-16
feat(dj): instrumental beds under between-track links

perminder-klair:
Closes the "Whacky idea: Beds" thread in Discord #suggestions (K8-Bit, skinny_dennis, Jaz666). **Off by default.**

`Song A ──cross──▶ bed (DJ talks) ──ramp──▶ Song B`

Today every link is talked *over* the incoming song, so every second of DJ content costs a second of the song it's introducing. skinny_dennis named the real cost: *"I'd be more inclined to work on skills with interesting content if it didn't mean that the DJ would talk over a big chunk of the music."* Skills, programmes and banter all add talk, so the roadmap keeps making links more expensive. This decouples the two.

Design doc: `docs/superpowers/specs/2026-07-16-dj-beds-design.md` (first commit — reviewable on its own).

## The bed is a track, not new mixer plumbing

Jaz666 called this "a fundamental change to the mixer". It isn't. The music has to actually stop for a bed to exist, and only the music chain can do that — so a bed is just a track in `dj_queue`. Everything needed already shipped:

- `cue_cut(music)` is already applied → `liq_cue_out` trims the bed to the script.
- `liq_cross_duration` already sizes a track's own exit → that *is* the ramp into the next song.
- The link's WAV is rendered at `queue.ts:947`, **before** the track URI is written at `:982` — so `wavDurationMs()` gives the bed's length at the exact moment it's decided.

**Mixer diff is one `on_meta` branch.** A bed carries no title/artist, so the existing `title != ""` gate would drop it silently — and that silence is the problem, because the controller learns "bed on air, air the link" only from the marker. It writes `bed-playing.json` instead, mirroring `jingle-playing.json` (#997), and never touches `now-playing.json` or the ICY title.

No `liquidsoap_*.txt`, so toggling beds costs **no mixer restart**.

## The trigger

skinny_dennis proposed "bed if TTS.wav > scanned Vocal Inactivity" — right instinct, wrong column. `intro_ms` is only a vocal onset when Demucs ran (heavy image + `ANALYZE_VOCAL_ACTIVITY=1`, defaults empty). Otherwise it's an energy heuristic its own docstring calls *"a soft budget, never a gate"* — and it measures the wrong thing here: a track opening full-band with vocals at 0:15 reads `intro_ms ≈ 0`, firing a bed exactly where the ramp is longest.

So the policy reads the **three-state** vocal semantics already in `library-db.ts`, which fall out neatly:

| `vocalRanges` | ramp budget | result |
|---|---|---|
| non-empty | `introMs` (the real onset) | bed iff the DJ would outlast the vocal |
| `[]` instrumental | `Infinity` | **never** bed — nothing to trample |
| `null` not computed | — | duration threshold (default 12s) |

Threshold by default, vocal onset as an automatic upgrade wherever the data exists.

## Withdrawn during design

I floated "loop Song A's outro as the bed" in the thread. It doesn't survive inspection: the exit-loop effect is built *inside* the cross callback, so the loop and Song B's fade-in share one canvas — Song B would be buried under the loop for its full duration. Same flaw rules out doing the bed in `dj_transition` at all. Documented under approaches-considered so it doesn't get re-proposed.

## Overrode the spec on one call

The spec said *don't* reuse `sounds/bed.mp3` (it's the asset the studio bed was disabled over). Reversed after measuring it: **71s**, already committed, already baked into both images, long enough for any script. The studio bed failed as *continuous* playback — a drone under everything, forever. As a link bed it plays alone for one link and stops. Not worth a new asset before hearing whether the existing one works in the new context.

It ships as the one bundled default and is **deletable** — bed defaults deliberately don't use sfx's undeletable-`builtin` semantics, because a bed the operator finds annoying is the entire risk of the feature. `beds.json.retired` makes the deletion stick across restarts.

## Verified

- `scripts/bed-policy.test.ts` — new, pins all three vocal states, threshold fallback, the ramp arithmetic, pool filtering. **Full suite: 36/36 files pass.**
- `npm run lint` — **0 errors**, controller + web.
- **`radio.liq` checked against the real broadcast image** (`liquidsoap --check`, exit 0) — and confirmed the harness catches a deliberate break (exit 1, precise line), so the pass means something.
- **The bed URI resolved in the real image**, which is what the whole feature rests on:
  ```
  subwave_kind=bed  liq_cue_out=24.50  liq_cross_duration=6.00  title=[] artist=[]
  ```
  The custom annotation survives to metadata (so the `on_meta` branch fires), the cue/cross land, and the empty title confirms the branch is genuinely necessary.

Not verified on-air — no listening test, and taste is the actual risk here (see below).

## Scope

**In:** between-track links only. **Out:** request intros (ride the heavy duck by design; bedding them means reworking duck routing) and programme/banter segments (fire mid-track, so bedding them means stopping a song mid-play). Both documented as follow-ups.

## Main risk

Taste, not feasibility. `radio.liq:1193` already has `bed_enabled = false` — *"the looping ambient bed was audible/annoying under the DJ's voice during links."* Different mechanism so the failure doesn't transfer, but if every link gets a bed, this is morning-zoo radio. Hence default-off, a 12s threshold rather than 5s, an audition button in the admin UI, and deletable defaults.


perminder-klair:
**6386bb9** addresses all 15 findings from the multi-agent review:

| # | Finding | Fix |
|---|---|---|
| 1 | Stale `bed-playing.json` re-fires after restart, airing a recovered link over the wrong song | `onBedStarted` only treats a marker <10s old as an edge (`BED_MARKER_FRESH_MS`); also skips the disk read when no bedded item is pending |
| 2 | Entry-side FX (sweep/blend/dissolve/chop) armed for the A→B seam get applied to the bed | stripped in `maybePushBed` after the bed commits, plus the armed transition stinger |
| 3 | `rampBudgetMs` trusts `introMs` that a `keepVocal` lean re-analysis desyncs from the ranges | onset now read off `vocalRanges` itself (`min startMs`) |
| 4 | Jingle rotate can insert a stinger between bed and song | new `bed_on_air` ref gates jingle availability in radio.liq (`--check` exit 0 in the broadcast image, negative control verified) |
| 5 | Operator cancel orphans the bed in `dj_queue` | `resolveDjQueueRidWithBed` finds the bed by `subwave_kind` in request metadata; `removeUpcoming` cancels it too |
| 6 | Marker fires at cross-FEED time → DJ talks over Song A's fade | bed sized for the predecessor's exit canvas (`bedLengthFor` entryCrossSec) and the link held until it elapses |
| 7 | `ensureDefaults` installs a null-duration bed that's listed but never selectable | skipped with a warning (same gate as `importAudio`), retried next boot |
| 8 | `shouldDropStaleLink` can drop a link whose bed already committed → naked bed | the same check runs at bed-push time against the drain-time predecessor |
| 9 | Intro WAV reaped (>1h predecessor) → committed bed airs naked | `airIntro` re-renders from the script when the WAV vanished |
| 10 | No `bedded` guard → double bed on crash-recovery re-drain | `if (item.bedded) return` |
| 11 | Uncontrolled threshold input can display never-persisted values | controlled inputs with snap-back on invalid entries |
| 12 | `BED_HEAD_SEC` cancelled out of every test assertion | constants pinned literally, entry-cross case added |
| 13 | `beds.json` written non-atomically while read on the drain path | `writeFileAtomic` (sfx.json too — same inherited flaw) |
| 14 | `crossSec` plumbed through five layers but reachable only by raw API | ramp control added to the Beds section |
| 15 | Byte-identical `slugify` triplicated | extracted to `util/slug.ts`, shared by beds + sfx |

Deliberately **not** done here: the full asset-store/router/upload-modal unification across jingles/sfx/beds (finding 15's larger ask) — it would triple this diff's blast radius; only the byte-identical primitives were extracted.

Verified: bed-policy test + full 36-file suite pass, controller & web lint clean (`tsc` web errors seen locally are a stale `node_modules`, not this branch), `liquidsoap --check` exit 0 against the real broadcast image.

*Still not verified on-air — the timing changes (entry-cross hold, jingle deferral) deserve a listening test before enabling beds on the live station.*

perminder-klair:
Review fixes landed in 537becd5:

- **0s crossfade preserved in bed entry sizing** — `maybePushBed`'s `|| 10` turned a legitimate `crossfadeDuration: 0` (hard-cut station) into 10s of phantom entry cross, i.e. ~10s of bare bed before the DJ speaks. Now guarded with `Number.isFinite`, matching `getAnnotatedUri`'s `??`-chain semantics; the 10s fallback applies only when the value is genuinely absent.
- **`bed_on_air` latch closed** — it was only reset by a titled track, so an untitled non-bed item after a bed (emergency single, untagged file) left it stuck true, starving the jingle rotate until the next tagged song. `on_meta` now resets it in an `else` branch. `liquidsoap --check` against the real broadcast image: exit 0.
- Documented the FIFO-order assumption behind `orderedRids` and the persist round-trip that makes the `bedded` re-drain guard survive restarts.

One item stays on the on-air verification checklist rather than in code: the "never pushes an ICY title" claim. `on_meta` only calls `radio.insert_metadata` for titled tracks, so the *explicit* ICY push is confirmed absent for beds — but the encoder may still forward the bed's empty-title source metadata natively, which would blank hardware-radio displays for the bed's duration instead of leaving Song A showing. Jingles have carried the same empty-title shape all along without complaints, so this is likely fine, but worth an ear during the listening test.

Verified: `bed-policy.test.ts` passes, controller lint 0 errors, radio.liq type-checks in the broadcast image.

perminder-klair:
**On-air verification done** — ran the branch's dev stack from a worktree and watched a real bed air end to end:

```
[beds] bed "ambient-room" 28.35s (10s entry cross) → 6s ramp into "while you were sleeping" (14s link, no vocal onset, over 0s)
[beds] bed on air → airing the link for "while you were sleeping" in 9.2s (entry cross)
```

- **Sizing**: 10s entry (the predecessor's real washout `crossSec` stamp, not the fallback) + 2.5 head + ~13.85 voice + 2 tail = 28.35s. ✓
- **Mixer**: `radio.log` shows `on_meta: bed started` at 11:23:47 and the next song's metadata at 11:24:10 — 23s later, against a predicted cross-feed of 22.35s (28.35 − 6s ramp). ✓
- **Marker**: `bed-playing.json` written with filename + startedAt; controller caught the edge on the next tick and held the link 9.2s (10s entry − detection latency). `introAired` flipped exactly once. ✓
- **UI**: `now-playing.json` stayed on the outgoing song for the whole bed — never leaked. ✓
- **ICY (the review's open question, now closed)**: polled Icecast `status-json.xsl` every 2s through the window — the title held the outgoing song for the full bed (t+2s…t+20s) and flipped to the incoming song at t+22s. The bed's empty-title metadata never reaches Icecast; only the explicit `insert_metadata` in the titled branch writes ICY. No blanking on hardware radios. ✓
- **Persistence**: `bedded: true` / `bedEntrySec: 10` present in `queue.json`, so the restart re-drain guard holds. ✓

Not exercised: the jingle gate (no jingle fell due during the window) and the actual listening/taste test.

PR #1080, opened 2026-07-17
Feat/embedding api key field

### LINKED ISSUE #1079: feat : Add API key field to Embedding settings
Currently the embedding settings panel has no input for an API key. When using an openai-compatible embedding server that requires authentication (e.g. a self-hosted proxy like CLIProxyAPI), there is no way to configure the Bearer token from the UI  the field exists in settings.json and the code reads it (apiKey: s.apiKey || ...), but there is no
form field to set it.

geekylakshya:
close #1079

## What

Adds a **Bearer Token** password input to the **Embedding Settings** panel when the embedding provider is set to `openai-compatible`. This mirrors the existing Bearer Token field already available in the LLM settings.

## Why

The `openai-compatible` embedding provider already supports authentication via `settings.embedding.apiKey`, and the backend correctly reads and redacts this value. However, the admin UI did not expose a field to configure it.

As a result, operators using authenticated OpenAI-compatible proxies (such as **CLIProxyAPI** or **LiteLLM**) had to manually edit `state/settings.json` and restart the controller to provide a bearer token.

## How Tested

- Switched the embedding provider to `openai-compatible`, entered a bearer token, and saved the settings.
  - Verified that `settings.json` was updated.
  - Confirmed the UI displays the `(on file)` placeholder after reloading.
- Used **Test Embeddings** with an unsaved bearer token.
  - Verified that the key is forwarded to the probe endpoint.
  - Confirmed the test request authenticates successfully using the supplied token.
- Verified that the Bearer Token field is **not** displayed for cloud providers or `ollama`.

## Notes for Reviewers

The backend change is a single line: `apiKey` was added to the probe endpoint's allowed query parameter whitelist, enabling users to test an unsaved bearer token before saving it.

All remaining changes are UI-only.

The `locca` provider intentionally does **not** expose this field, as it is designed for local, unauthenticated use.

### DIFF
--- a/controller/src/routes/settings.ts
+++ b/controller/src/routes/settings.ts
@@ -801,18 +801,21 @@ router.get('/settings/llm/models', requireAdmin, async (req, res) => {
 });
 
 // ---------------------------------------------------------------------------
-// GET /settings/embedding/probe — test whether the configured (or supplied)
+// POST /settings/embedding/probe — test whether the configured (or supplied)
 // embedding endpoint can actually produce embeddings, surfacing the result
 // in the admin UI BEFORE a long tagging run instead of failing mid-job.
-// Optional query overrides (provider/model/baseUrl/ollamaUrl) test the unsaved
-// form values; omitted fields fall back to saved settings.embedding → llm.
+// Optional body overrides (provider/model/baseUrl/ollamaUrl/apiKey) test the
+// unsaved form values; omitted fields fall back to saved settings.embedding →
+// llm. POST body rather than query params so the bearer token never rides a
+// URL that reverse-proxy access logs capture — same shape as
+// /settings/llm/probe-compat.
 // Always 200s with { ok, dim, code, message } — a chat-model / unreachable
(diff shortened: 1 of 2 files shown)


perminder-klair:
Thanks for the contribution — the backend plumbing was exactly right (`settings.update()` already handled the `'set'` sentinel, and `resolveEmbeddingConfig` picks up the override). I pushed `aac7a09` with two review fixes on top:

1. **Probe is now `POST` with a JSON body** instead of `GET` query params — an unsaved bearer token in a query string ends up in reverse-proxy access logs (the BYO compose supports operator Traefik/nginx). This mirrors the existing `POST /settings/llm/probe-compat` shape. Both UI callers (`runProbe` and the locca `detect`) were updated.
2. **Dropped the extra "Bearer token saved" toast + `refresh()`** after `saveSettings` — `saveSettings` swallows errors internally and already toasts/refreshes for the whole patch, so the added ok-toast fired even when the save failed. Now it just clears the input, same as the LlmSection compat-key flow.

PR #1083, opened 2026-07-17
fix: per-provider base URLs for locca and openai-compatible (#1082)

### LINKED ISSUE #1082: bug: locca and openai-compatible share baseUrl — changing one overwrites the other
## What

Fixes provider-specific configuration for base URLs and bearer tokens in the LLM, fallback LLM, and embedding settings.

## Why

Changing the base URL for `locca` currently overwrites the saved base URL for `openai-compatible`, and vice versa. The same issue exists for both the fallback LLM configuration and the embedding provider.

### Root Cause

`llm.baseUrl`, `llm.fallback.baseUrl`, and `embedding.baseUrl` are each stored as a single shared string, regardless of the selected provider. As a result, both the `locca` and `openai-compatible` UI inputs are bound to the same underlying value, with no provider-specific namespace in either the schema or persisted settings.

### Additional Issues

- The **"(on file)"** placeholder for the Bearer Token checks `keys['openai-compatible']` even when the selected provider is `locca`, so a saved `locca` token is never indicated in the UI.
- The `locca` bearer token is silently omitted from the save payload because the save logic only includes `apiKey` when `provider === 'openai-compatible'`.

## How Tested

- Configured different base URLs for both `openai-compatible` and `locca`.
  - Verified that each provider retains its own URL after saving and reloading.
- Repeated the same validation for the fallback LLM and embedding provider settings.
- Saved bearer tokens for both `openai-compatible` and `locca`.
  - Verified that the correct `(on file)` placeholder is displayed for the active provider.
  - Confirmed that both providers' tokens are persisted correctly and no longer dropped from the save payload.
- Verified migration from the legacy `baseUrl` field preserves existing configurations.

## Notes for Reviewers

This change introduces provider-specific base URL storage by replacing:

```ts
baseUrl: string
```

with:

```ts
providerBaseUrls: Record<string, string>
```

in the following configuration sections:

- `llm`
- `llm.fallback`
- `embedding`

A migration copies the legacy `baseUrl` value into `providerBaseUrls[currentProvider]` on first load to maintain backward compatibility.

The UI now binds each provider's Base URL input to its corresponding entry in `providerBaseUrls`, preventing configuration from being overwritten when switching providers.

Additionally, the Bearer Token placeholder and save logic have been updated to correctly support both `openai-compatible` and `locca`.

geekylakshya:
Replace the shared single `baseUrl` field in `llm`, `llm.fallback`, and `embedding` settings with a `providerBaseUrls` map keyed by provider id. Switching between locca and openai-compatible no longer overwrites each other's saved server URL.

Migration: on first load, any existing `baseUrl` value is seeded into `providerBaseUrls[currentProvider]` so no URL is lost on upgrade. The legacy `baseUrl` field continues to be derived at runtime from the map so all downstream consumers (registry, legs, embedding) are unaffected.

Also fixes two secondary bugs in the same area:
- Bearer token "on file" hint was checking `keys['openai-compatible']` even when provider was `locca`; now checks `keys[activeProvider]`.
- locca bearer token was silently dropped from the save payload because the condition only checked for `openai-compatible`; now covers both inline-key providers.

Closes #1082


perminder-klair:
Pushed a follow-up commit (cf09420) addressing review findings:

- **Embedding + inherited provider (data loss):** `update()` derived `embedding.baseUrl` from `embedding.provider` alone, but the UI keys the map by the *effective* provider (own, else the chat provider). With the default inherit setting a dedicated embedding URL (#405) derived to `''` — embeddings silently fell back to the chat server, and the saved URL was wiped on the first Library-section save. Migration, seeding and derivation now all use the effective provider, and `embedding.baseUrl` re-derives on every `update()` so an `llm.provider`-only change moves the live slot too.
- **Stale URL on provider-only patches:** `applyLlmLegPatch` only re-derived `baseUrl` when the patch contained URLs. A raw-API/MCP `{llm:{provider:'locca'}}` kept the openai-compatible URL — the same wrong-server class as #1082. The flat field is now unconditionally re-derived from the map at the end of every leg patch.
- **locca bearer token:** the widened save condition was live but no bearer input ever rendered for locca. Added the input for locca on both legs, keyed the '(on file)' hint by the active provider, and the Test connection button defaults to the locca host URL when the field is blank. (The input already clears on provider switch, so no cross-provider key routing.)
- **Robustness/style:** `providerBaseUrls: null/[]/'x'` now throws a clean validation error instead of a TypeError; `||` instead of `??` in LibrarySection so a cleared field falls through to the chat URL; `load()` computes each leg's map once.

Verified with an 18-assertion load/update round-trip against a throwaway STATE_DIR (legacy migration, provider switches, inherit derivation, bad shapes) — all pass; `tsc` + `eslint` clean on controller and web.

PR #1089, opened 2026-07-18
fix(analyzer): pre-decode files libsndfile can't open once via ffmpeg

### LINKED ISSUE #1073: Heavy analyzer falls back to slow audioread decoder for MP3 files
**Describe the bug**
When running the heavy analyzer (subwave-analyzer-heavy / subwave-aio-heavy) against an MP3-based library, every track triggers a librosa/PySoundFile fallback warning and decodes via the much slower `audioread` backend instead of the fast `libsndfile` path:

/app/scripts/analyze_worker.py:439: UserWarning: PySoundFile failed. Trying audioread instead.
  y48, _sr48 = librosa.load(
/opt/analyzer/venv/lib/python3.13/site-packages/librosa/core/audio.py:184: FutureWarning: librosa.core.audio.__audioread_load
Deprecated as of librosa version 0.10.0. It will be removed in librosa version 1.0.
  y, sr_native = __audioread_load(path, offset, duration, dtype)

This appears on essentially every track during a sounds-like / vocal-activity backfill, significantly slowing down analysis on libraries that are predominantly MP3 — a very common case for Navidrome/Subsonic libraries.

**Expected behavior**
MP3 decoding should go through the fast libsndfile path rather than falling back to audioread. This is very likely due to the bundled libsndfile version being older than 1.1.0 MP3 support was only added to libsndfile in 1.1.0. Bumping the libsndfile version in the heavy analyzer image (or its base image) should resolve this for the majority of users with MP3 libraries.

**Steps to reproduce**
1. Run subwave-aio-heavy (or subwave-analyzer-heavy) against a Navidrome library that's mostly/entirely MP3.
2. Enable "Audio fingerprint (sounds-like)" and/or "Vocal activity" on the Library page.
3. Run a backfill.
4. Watch the container logs, the PySoundFile→audioread fallback warning appears for nearly every track.

**Environment**
- Install type: Unraid one-click (AIO)
- Image: ghcr.io/perminder-klair/subwave-aio-heavy:latest
- Library: mostly MP3, ~2,900 tracks
- Host: amd64, Unraid

**Additional context**
Not a functional bug. Analysis completes correctly, just noticeably slower than expected on MP3-heavy libraries due to the decoder fallback. Flagging since it's likely to affect most operators, as MP3 is probably the most common format in Navidrome libraries.

perminder-klair:
## Summary

Fixes #1073 — but not the way the issue proposed. Investigation showed the published heavy images already bundle **libsndfile 1.2.2** (via the `soundfile==0.14.0` wheel) with MP3 support enabled, so a version bump changes nothing. The real problem is what happens whenever libsndfile can't open a *specific* file:

- **m4a/AAC** — never supported by libsndfile at all (this is most of the reference station's library);
- **quirky MP3s** — e.g. junk bytes before the first frame header (old rips/stream rips), which reproduce the reporter's `Format not recognised` → audioread fallback exactly.

For those files, **every** `librosa.load` in an analyze request fell back to the deprecated audioread path: a fresh full-file ffmpeg subprocess decode *per load* (duration probe + body + up to 3 CLAP windows + outro + vocal ≈ 5-7 decodes per track), warning spam in the logs, and a timebomb — audioread support is removed in librosa 1.0.

## The fix

`ensure_fast_decode()` in `analyze_worker.py`: probe the local file with soundfile once; if it can't be opened, decode it **once** to a temp WAV via ffmpeg (already present in every analyzer-capable image; native rate/channels so the stereo BS.1770 loudness sum stays correct) and run the whole request off that through the fast libsndfile path.

- Files soundfile already handles: untouched, zero overhead beyond the one open-probe.
- ffmpeg absent (bare local venv) or decode failure: falls back to exactly the old behaviour.
- Outro safety: a pre-decoded WAV defeats `analyze_outro`'s short-decode truncation check, so with *unknown* completeness the outro is skipped rather than risking mid-song audio measured as an ending.

## Verification

Ran the patched worker over its stdio protocol on three file classes, both in a local venv with the exact pinned wheels and **inside the published `subwave-aio-heavy:latest` image** (python 3.13):

| file | before | after |
|---|---|---|
| MP3 with junk prefix (reporter's symptom) | audioread fallback per load | pre-decoded once, analyzes clean |
| real m4a (raw from Navidrome) | audioread fallback per load | pre-decoded once, bpm/key/outro all measured |
| normal tagged MP3 | fast path | fast path, untouched |

`PySoundFile failed. Trying audioread instead.` warnings: **0** across all runs (previously several per unsupported track).

### DIFF
--- a/controller/scripts/analyze_worker.py
+++ b/controller/scripts/analyze_worker.py
@@ -39,6 +39,8 @@ the stored audio vectors possible.
 import importlib.util
 import json
 import os
+import shutil
+import subprocess
 import sys
 import tempfile
 import urllib.request
@@ -57,6 +59,11 @@ import urllib.request
 ANALYZE_SECONDS = float(os.environ.get("ANALYZE_SECONDS", "").strip() or "40")
 ANALYZE_SR = int(os.environ.get("ANALYZE_SR", "").strip() or "22050")
 FETCH_TIMEOUT_S = float(os.environ.get("ANALYZE_FETCH_TIMEOUT_S", "").strip() or "60")
+# Ceiling for the one-shot ffmpeg pre-decode (ensure_fast_decode) — the input
+# is a byte-capped download, seconds of decode work even on a weak NAS CPU.
+FFMPEG_DECODE_TIMEOUT_S = float(
+    os.environ.get("ANALYZE_FFMPEG_TIMEOUT_S", "").strip() or "120"
+)
 
 # --- CLAP audio embedding (optional, opt-in) -------------------------------
 # Off unless ANALYZE_AUDIO_EMBEDDING is truthy. CLAP wants 48 kHz mono; the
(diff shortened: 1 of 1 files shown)


perminder-klair:
End-to-end verification from the worktree (isolated STATE_DIR, real Navidrome, local venv backend with the exact pinned wheels):

Ran the controller's actual analysis pass (`analyze-library.ts --limit 3 --audio` — the same child `POST /library/analyze` spawns) over the full pipeline: `downloadCapped` prefetch → `.audio` temp on the state dir → patched worker → library.db.

- 3/3 m4a tracks analyzed (`ok=3 fail=0`), each logging exactly one `pre-decoded once via ffmpeg`
- `PySoundFile failed` warnings: **0** (previously every m4a load fell back)
- DB rows written with real bpm/key/LUFS and measured `outro_json`
- No leftover `swanalyze_dec_*.wav` temps; `analyze-tmp` cleaned on exit
- ~6s wall for 3 tracks including downloads

Unrelated pre-existing find while testing: `spawnChild` in `broadcast/tagger.ts` hardcodes `cwd: '/app'` and doesn't handle the child `error` event — on a bare-host dev run, clicking any tagger/analyze button crashes the whole controller (`spawn npx ENOENT` → unhandled 'error' → process exit). Harmless in containers where /app exists; worth a follow-up.

Issue #864, opened 2026-07-05
Non-user instigated time call-outs off


WhatKateDoes:
My slovenly DJ's are doing time call-outs, but about 10 mins late:

```
18:23:59
[play]
"Afraid of Time" by Hans Zimmer
18:20:16
[link]
18:10. This particular piece of Hans Zimmer feels like the exact moment the air leaves the room.
18:20:08
[pick]
Same album, maintains the same cinematic scale and tension.
18:20:00
[station-id]
18:15. Whatever coordinate you've drifted from, you've landed in the right frequency [confirmation-en]. This is The Ninth House, broadcasting from the Bone Marrow, and I'm Bubblina. Just settle in.
18:19:58
[pick]
Now playing "Cornfield Chase" by Hans Zimmer [id: yxWIfIfCt6FQmk0q2IDgNO] (after "Light of the Seven" by Ramin Djawadi). Pick the track to play next. Stay silent — no link this time. Set "transition" by what THIS moment needs — "washout" to dissolve out as it ends, "sweep" for a gear-change entry, "dissolve" to melt a clash into ambience, "blend" ONLY for an exceptionally locked pair (a plain crossfade already handles ordinary same-lane picks), "normal" for a plain hand-off. Vary your craft: never the same transition three picks running, and if your last pick used an effect, lean "normal" now unless the moment clearly calls again. Your recent transition choices, oldest first: normal, normal, normal, washout — the station strips a third repeat, so vary deliberately.
18:19:58
[play]
"Cornfield Chase" by Hans Zimmer
18:10:27
[pick]
Maintains cinematic weight, shifts from piano-led to organ-driven tension.
18:10:21
[link]
18:02. The air in the Bone Marrow is getting thick, and this piece of piano-led tension from Ramin Djawadi fits the humidity perfectly. It starts with a deceptive stillness that feels like a held breath. Just stay very still for a moment. I'll be watching the monitors.
```

WhatKateDoes:
And yet, in another shout-out, was correct, but it was followed immediately by another one that was out:

```
18:31:18
[link]
18:24. If you're currently staring at a flickering monitor or a long stretch of empty void, let's just lean into the stillness for a moment. There is a certain dignity in the quiet parts of the shift. Hans Zimmer returns to ground us. This one feels like a long exhale.
18:31:17
[pick]
Stays in cinematic lane, shifts from epic to reflective.
18:31:06
[station-id]
[...]

PR #1090, opened 2026-07-18
fix(dj): air show handoffs at a track boundary, in the right voice

perminder-klair:
Closes the two Discord reports about show changeovers. Spec + implementation.

## The reports

- **Gurthyy [NSTV]** — station IDs have a deferred path that waits for the next track boundary; handoffs should too, instead of firing at the top of the hour over the middle of a song.
- **Jaz666** — a morning → mid-morning changeover where the handoff sprawled across ~6 minutes, and a link generated for the incoming DJ was **spoken in the outgoing DJ's voice**.

## Root cause

Gurthyy is right that the deferred path exists (`queue.announceAtNextTrack`, one caller: station IDs). But the handoff's problem wasn't a missing deferral.

`queue.onTrackStarted` **already** called `runPersonaHandoff` before the pick, and **already** computed a look-ahead (`showAt = now + duration + 120s`) to pick under the incoming show's brief. It just rolled the session on the *live* clock:

```ts
// The session roll and handoff above stay on the live clock: only the pick looks ahead.
```

At 09:58 the grid still says "morning show", so that roll never fired and the `:00` cron won it mid-song. Net effect: the changeover track — already picked *for the new show* — aired **before anyone had handed over**. That is Jaz's timeline exactly.

## What changed

**One date drives the whole boundary sequence.** Roll, episode plan, mic-pass, episode hook and pick all key off `showAt`, so the mic-pass lands in front of the changeover track and the pick's brief can't disagree with the on-air persona — there's no second date to disagree with. This needed *less* machinery than deferring: `session.pendingHandoff()` was already a "wait for a boundary" mechanism, so `announceAtNextTrack` is untouched.

**The hourly cron rolls but no longer airs** (`rollSessionNow({ airHandoff })`). Takeover routes still air immediately — an operator action is explicit. A pending mic-pass expires after 20m, and supersedes a pending station ident on the same boundary.

**The persona is carried on the queue item.** `speak()` at drain and `voiceGainDb()` at air each re-resolved it, so generation, render and air could name three different DJs. `voiceGainDb` in `airIntro` was the last call site in the codebase still resolving from the clock. Speakers now resolve once via `session.onAirPersona()`.

## Two things worth reviewer attention

**1. A prerequisite that inverts the design if missed.** `getFullContext()` didn't return the date it was built for, so `session.start()` resolved the persona off the wall clock. An anticipated roll would build a session with the incoming *show* but the outgoing *persona* — `stampRolledFrom` then sees no persona change, sets `rolledFrom = null`, and the mic-pass is **suppressed entirely**. Worse than the bug being fixed. Contexts now carry `at`.

I verified this both ways against the real modules with a seeded state dir: a look-ahead roll arms the mic-pass with the incoming persona, and **stripping the `at` stamp reproduces the suppression on demand** (show rolls to "Mid Morning", persona stays on the outgoing DJ, `pendingHandoff()` null).

**2. Identity looks ahead; the clock does not.** Caught while implementing: `SCRIPT_CONTEXT_FIELDS` includes `clock`, so driving the handoff prompts off `showAt` would have handed them a time up to a track-length + 2min ahead of when the mic-pass actually airs — the DJ misstating the time on air, the failure #864 fixed for links. The handoff now takes `date`/`clock`/`time` from the live moment and `show`/`mood`/`festival` from the look-ahead.

## Known edge, documented

The session now leads the schedule grid by up to ~2 minutes, so `getEffectivePersona()` (wall clock) disagrees with the session in that window. `session.onAirPersona()` is the fix on the paths that write or voice lines. The three no-arg `pickOnAirSpeaker()` sites in `scheduler.ts` were **deliberately left alone** — idents fire at `:15/:30/:45` and the hourly at `:00`, none inside the window; only the manual link route can disagree, and its fallback is the outgoing DJ speaking once more.

## Verification

- `npm test` — 37/37 files pass, including new `scripts/handoff-boundary.test.ts` (14 assertions pinning `contextDate` and `handoffIsStale`, both pure).
- `npm run lint` — 0 errors (550 pre-existing repo-wide `no-explicit-any` warnings; this branch adds none and removes one).
- Integration check of the Change 0 chain against real modules, positive and counter-proof, as above.

Not yet run on the live stack — the on-air check (handoff timestamp landing next to an `onTrackStarted` rather than `:00`) is listed in the spec's manual verification section.

### DIFF
--- a/controller/scripts/handoff-boundary.test.ts
+++ b/controller/scripts/handoff-boundary.test.ts
@@ -8,7 +8,7 @@
 // pick off that one date — so the sign-off/greeting land in front of the
 // changeover track, and the pick's brief can't disagree with the on-air persona.
 //
-// Two pure pieces hold that up, pinned here:
+// Three pure pieces hold that up, pinned here:
 //
 //  - contextDate    — reads the `at` stamp getFullContext puts on every context.
 //                     This is the load-bearing bit: session.start() resolves the
@@ -22,11 +22,18 @@
 //                     hourly cron rolls without airing, so with nobody listening
 //                     (or across one very long track) a pending handoff can
 //                     outlive the moment it describes.
+//  - rollIsBackward — refuses to roll the session to an OLDER moment than the
+//                     one it was resolved for. Inside the look-ahead window a
+//                     live-clock maybeRoll caller (routes/request.ts) still
+//                     resolves the outgoing show; unguarded, a listener request
+//                     there would archive the just-rolled session, arm a
+//                     mirrored reverse mic-pass, and voice the request intro as
+//                     the DJ who just signed off.
 //
 // node:assert-via-tsx style, matching scripts/stale-link.test.ts.
 
 import assert from 'node:assert/strict';
-import { contextDate, handoffIsStale } from '../src/broadcast/session.js';
+import { contextDate, handoffIsStale, rollIsBackward } from '../src/broadcast/session.js';
 
 const MAX_AGE = 20 * 60_000;   // mirrors HANDOFF_MAX_AGE_MS in dj-agent.ts
 
@@ -120,6 +127,34 @@ function main() {
   test('non-numeric stamp → treated as unstamped, not stale', () => {
     for (const bad of ['2026-07-18T10:00:00.000Z', NaN, Infinity, {}] as any[]) {
       assert.equal(handoffIsStale(bad, now, MAX_AGE), false, `at=${String(bad)}`);
+    }
+  });
+
+  console.log('\nbackward-roll guard (rollIsBackward):');
+  // The session was look-ahead-rolled for 10:02 at a real time of ~09:58.
+  const sessionAt = '2026-07-18T10:02:00.000Z';
+
+  test('live-clock caller inside the look-ahead window → backward, skip', () => {
+    // A listener request at 09:58:30 still resolves the outgoing show.
+    assert.equal(rollIsBackward(new Date('2026-07-18T09:58:30.000Z'), sessionAt), true);
+  });
+
+  test('caller past the described moment → forward, roll allowed', () => {
+    assert.equal(rollIsBackward(new Date('2026-07-18T10:02:00.001Z'), sessionAt), false);
+  });
+
+  test('exactly the described moment → not backward (strict <)', () => {
+    assert.equal(rollIsBackward(new Date(sessionAt), sessionAt), false);
+  });
+
+  test('session persisted before ctxAt existed → never blocks a roll', () => {
+    assert.equal(rollIsBackward(new Date('2020-01-01T00:00:00.000Z'), undefined), false);
+    assert.equal(rollIsBackward(new Date('2020-01-01T00:00:00.000Z'), null), false);
+  });
+
+  test('garbage stamp → never blocks a roll', () => {
+    for (const bad of ['not-a-date', '', 12345, {}] as any[]) {
+      assert.equal(rollIsBackward(new Date(), bad), false, `ctxAt=${JSON.stringify(bad)}`);
     }
   });
 
--- a/controller/scripts/handoff-boundary.test.ts
+++ b/controller/scripts/handoff-boundary.test.ts
@@ -1,0 +1,130 @@
+// Regression tests for the show-handoff boundary logic.
+// Run: `tsx scripts/handoff-boundary.test.ts`.
+//
+// A persona handoff used to fire from the `0 * * * *` cron and duck the middle
+// of whatever song was playing. It now rides the TRACK grid: queue.onTrackStarted
+// resolves the show for the moment the next pick will air (`showAt = now +
+// duration + PICK_SHOW_LOOKAHEAD_SEC`) and drives the roll, the mic-pass AND the
+// pick off that one date — so the sign-off/greeting land in front of the
+// changeover track, and the pick's brief can't disagree with the on-air persona.
+//
+// Two pure pieces hold that up, pinned here:
+//
+//  - contextDate    — reads the `at` stamp getFullContext puts on every context.
+//                     This is the load-bearing bit: session.start() resolves the
+//                     persona from it, and if it silently fell back to the wall
+//                     clock a look-ahead roll would stamp the OUTGOING persona
+//                     onto the INCOMING show's session. stampRolledFrom would
+//                     then compare that persona against itself, see no change,
+//                     and suppress the mic-pass ENTIRELY — worse than the bug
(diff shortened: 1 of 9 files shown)


perminder-klair:
## On-air verification (dev stack from this worktree)

Ran the dev stack against this branch with the operator's real library, personas and schedule. Boundaries were forced with a timed takeover (its expiry is date-aware, so the look-ahead crosses it — the same code path a weekly-grid boundary takes).

### ✅ Timing — the core fix, verified

Takeover pinned Sona's Golden Hour (p_ember), **expiring 15:22:37**. Grid returns Deep Work (p_vesper) at expiry — a real persona change.

```
15:18:48.077  show              key=show:s_focus     ← rolled to the INCOMING show
15:18:48.077  handoff  trace.start  from=Sona to=Heer
15:18:48.364  generateSignoff
15:18:48.674  generateHandoffGreeting
15:18:48.675  track-event                            ← TRACK BOUNDARY
15:18:48.840  djAgentPick                            ← pick came AFTER the mic-pass
```

- Mic-pass fired at **15:18:48, on a track boundary** (track starts confirmed at 15:20:37, 15:22:36, 15:24:25… — `track-event` marks each).
- That is **3m49s ahead of the 15:22:37 boundary** — the look-ahead anticipating it, as designed.
- **At 15:22:36–37, when the boundary actually passed: nothing was spoken.** It passed silently, mid-track. That is precisely the reported bug not happening.
- The pick followed the mic-pass, not the other way round.

Control: the takeover's *own* handoff at 15:07:37 fired immediately mid-track (nearest track start 25s later) — correct and unchanged for an explicit operator action.

### ✅ Greeting attribution (Change 6) — verified against a real before/after

Archived sessions from before this branch:
```
22:01:23  handoff  speaker=Heer      ← sign-off had attribution
22:01:41  handoff  speaker=(none)    ← greeting had NONE   ⟵ the bug
```
This build:
```
15:43:32  handoff  speaker=Heer      ← sign-off (outgoing) ✓
15:43:51  handoff  speaker=Sona      ← greeting (incoming) ✓
```
Both turns now carry the correct, distinct speaker — the same metadata that selects the TTS voice.

### ✅ Also confirmed live

- `rolledFrom.at` stamped on every roll (`{"personaId":"p_vesper","personaName":"Heer","showName":"Deep Work","at":1784456001108}`).
- `handoffAired` lifecycle: `false` while pending → `true` once aired.
- A roll left the mic-pass **pending** rather than airing it mid-track, and a later boundary took it — the Change 2 behaviour.

### ⚠️ Not verified on air: `introPersona` on a track-tied link (Change 5)

Honest gap. The station's cloud model hit its **weekly usage limit** mid-session (`ollama.com` quota), and the local fallbacks don't clear the bar on this box: `qwen3.5:9b`/`4b` blow the 90s agent deadline, `llama3.2:1b` fails structured-output validation. With the pool picker unable to complete, `enqueuePick` never runs, so no track-tied link is ever generated.

Change 5 is covered by typecheck and by inspection (`voiceGainDb` in `airIntro` was the last call site resolving persona from the clock), and the mechanism it shares with the handoff path is proven above — but the link case itself has **not** been observed on air. Worth a re-run once model quota is back.

Environment restored afterwards: takeover cleared, `controller/.env` reverted, stack torn down, ports free. The main checkout was never touched.

perminder-klair:
### Code review

Found 3 issues:

1. Deferring the hourly handoff breaks the programme-intro guard. The cron now rolls with `airHandoff: false`, so `runPersonaHandoff` never runs and `handoffAired` stays `false` — but `maybeRunIntro`'s skip-guard requires `rolledFrom && handoffAired`. On a persona-change boundary into a programme show, the standalone episode intro now airs immediately at wall-clock :00, mid-song (the exact bug class this PR fixes), and the episode is introduced a second time when the deferred mic-pass greeting airs carrying `episodeAngle`.

https://github.com/perminder-klair/subwave/blob/938ec5bb9ccbabc1b0ff136d7700b530c30ddc36/controller/src/broadcast/programme.ts#L195-L199

https://github.com/perminder-klair/subwave/blob/938ec5bb9ccbabc1b0ff136d7700b530c30ddc36/controller/src/broadcast/scheduler.ts#L514-L516

2. A listener request can roll the look-ahead session backward. `resolveRequest()` still fetches a live-clock context and calls `session.maybeRoll(ctx)`, and `maybeRoll` compares session keys with no direction guard. In the look-ahead window (up to track duration + 120s before every show boundary, after `onTrackStarted` has rolled to the incoming show), a request hard-rolls the session back to the outgoing show and stamps a fresh reverse `rolledFrom` with `handoffAired: false` — the request intro is then voiced by the outgoing DJ right after the incoming DJ's greeting, and a bogus reverse mic-pass is armed for the next boundary.

https://github.com/perminder-klair/subwave/blob/938ec5bb9ccbabc1b0ff136d7700b530c30ddc36/controller/src/routes/request.ts#L210-L224

3. Programme planning silently no-ops during the look-ahead window. `programme.ensurePlan(ctx)` and `programme.onSessionSettled(this, ctx)` omit the `now` argument, so `activeEpisode()` defaults to the live clock: `settings.resolveActiveShow(now)` resolves the outgoing show while the session key is already the incoming show, the key comparison fails, and `ensurePlan` returns early without building the plan. The handoff greeting into a programme show therefore airs with `episodeAngle: null`, defeating the stated purpose of the call ("Plan a programme episode BEFORE the mic-pass so a handoff into a programme show can weave the episode angle into its greeting").

https://github.com/perminder-klair/subwave/blob/938ec5bb9ccbabc1b0ff136d7700b530c30ddc36/controller/src/broadcast/queue.ts#L1467-L1471

https://github.com/perminder-klair/subwave/blob/938ec5bb9ccbabc1b0ff136d7700b530c30ddc36/controller/src/broadcast/programme.ts#L73-L79

perminder-klair:
All three review findings fixed in 8ab5788f:

1. `maybeRunIntro` now stays pending while `session.pendingHandoff()` is armed — the boundary tick resolves the intro after `runPersonaHandoff` settles, so no mid-song intro at :00 and no double introduction.
2. Sessions carry `ctxAt` and `session.maybeRoll` refuses to roll to an older moment (`rollIsBackward`, pure + unit-pinned) — a live-clock listener request inside the look-ahead window can no longer roll the session backward.
3. `programme.ensurePlan`/`onSessionSettled` default `now` to `session.contextDate(ctx)`, so the episode plan builds at the roll tick and the handoff greeting carries the episode angle.

All 37 test files pass (5 new `rollIsBackward` assertions), lint 0 errors. Spec updated with a Review fixes section.

PR #1102, opened 2026-07-19
feat: CUDA analyzer flavour + quiet-times analysis gate (#1099)

### LINKED ISSUE #1099: [BUG / Feature Request] Support Native GPU / CUDA Acceleration for subwave-analyzer-heavy (CLAP & Demucs)
I am deploying SUB/WAVE with a massive music collection and have enabled the split stack architecture with the heavy analyzer (ANALYZER_HEAVY=1). 
However, the initial library scan and background analysis for CLAP embeddings (laion/clap-htsat-unfused) and Vocal Activity Demucs processing execute entirely on the host CPU.
Even when passing NVIDIA hardware configurations via `docker-compose.yml (deploy.resources.reservations)`, configuring host runtimes, and enabling internal app flags `(FORCE_CUDA=true)`, nvidia-smi shows 0% GPU utilization and no active VRAM allocation.

**Describe the solution you'd like**

I would love to see native GPU/CUDA acceleration implemented inside the subwave-analyzer-heavy service stack. This would allow hosts with dedicated graphics cards to rapidly unburden their CPU threads and accelerate deep library ingestion.
To figure out why the GPU wasn't being targeted, I ran a container diagnostic check via:
`bashdocker exec -it sub-wave-analyzer python3 -c "import torch; print(torch.cuda.is_available())"`

This fails with a `ModuleNotFoundError: No module named 'torch'.`

It appears that while subwave-analyzer-heavy successfully pulls a different image target, the app's current extraction backend (whether utilizing an ONNX Runtime build, specific Node/Bun machine learning wrappers, or a frozen binary) is executing inside an environment missing a CUDA-bound framework or missing the explicit hardware execution providers (like CUDAExecutionProvider for ONNX) required to push the tensor graphs down to the GPU.

---------------------
OK, the above was written by Gemini, but only after aiding me in debugging.

If you do manage to add this, I was also suggest adding a `Run at Quiet Times` tick box, for users like me who already run a Local LLM and Local TTS, and don't have the overhead to run another GPU process when the station is actively broadcasting to listeners.
This would wait until listener count drops to 0 for x amount of minutes before starting the library scan.

Cheers!

perminder-klair:
Closes #1099.

Two independent deliverables from the issue, both behind explicit opt-ins — CPU installs see zero behaviour change.

## Part A — `subwave-analyzer-cuda` (GPU CLAP + Demucs)

The heavy analyzer has always shipped **CPU-only torch wheels**, so `deploy.resources` GPU reservations could never light up the GPU (and `FORCE_CUDA` was never a real flag). This adds a third published flavour and the plumbing to use it:

- **`docker/Dockerfile.analyzer`**: new `WITH_CUDA=1` build arg switches the torch index to **cu124** (torch/torchaudio stay pinned at 2.6.0 — one torch *version* repo-wide, in lockstep with chatterbox). Base image unchanged: cu124 wheels vendor the CUDA runtime, so the host needs only the NVIDIA driver + Container Toolkit.
- **`docker-compose.analyzer-gpu.yml`** (new, mirrors the `tts-heavy-gpu` overlay): swaps the `analyzer` service to `subwave-analyzer-cuda` and reserves the GPU. `docker compose -f docker-compose.yml -f docker-compose.analyzer-gpu.yml up -d`. `ANALYZER_HEAVY` is irrelevant while applied. Embedded into the CLI assets so `subwave init` materialises it.
- **`analyze_worker.py`**: `resolve_device()` (`ANALYZE_DEVICE=auto|cpu|cuda`, auto default) — CLAP transformers model + text tower `.to(device)`, Demucs `apply_model(device=…)`. A cuda request with no visible GPU **warns and falls back to cpu**; the ONNX CLAP path deliberately stays CPU; the AIO stays CPU-only.
- **CI**: new `subwave-analyzer-cuda` matrix entry in `publish-images.yml` (amd64, ~9.9GB — the nvidia-cu12 libs).

## Part B — quiet-times gate ("run at quiet times")

Off by default. When enabled (`settings.audio.analyzeQuietOnly` + `analyzeQuietMinutes`, default 10 min; `ANALYZE_QUIET_ONLY=1` env wins on), the analysis pass **pauses between tracks while anyone is listening** and resumes once the stream has been listener-free for the window.

- Enforced inside `runAnalysisPass`, so it covers the server-spawned tagger child AND `npm run analyze`, and reacts within one 30s poll — start-time gating would do nothing for a pass that runs for hours. Manual "Analyse now" runs are gated too (the bypass is the toggle, not the button).
- Decision logic is a pure helper (`music/analyze-quiet-pure.ts`) pinned by `scripts/analyze-quiet.test.ts`. It **fails open** on an unknown listener count — the opposite direction from `djCallsAllowed()`: a stats outage must never stall a library scan; outage time accrues quiet so a recovery at 0 listeners doesn't pause a running pass in an empty room.
- Count comes from a new `listeners.probeListenerCount()` — same fetch + Safari dedupe as the monitor, minus the history append (a second writer in the child process would stripe duplicate sparkline rows).
- Admin Library panel grows a "Quiet times" row (toggle + idle-minutes input) next to the sounds-like/vocal controls; a waiting run shows "Waiting for quiet (N listening)".

Design spec: `docs/superpowers/specs/2026-07-19-issue-1099-analyzer-cuda-quiet-design.md`.

## Verification

- `npm run lint` clean in controller (0 errors) + web; CLI `tsc --noEmit` clean; `npm test` 37/37 controller test files incl. the new gate suite.
- Compose merges validated over both `docker-compose.yml` and `docker-compose.byo.yml` (rendered service shows the cuda image, build args, and device reservation); CLI embedded assets regenerated (verify job parity).
- **Built the cuda image locally** (amd64, no GPU): torch reports `2.6.0+cu124`, all heavy deps import, `resolve_device()` → cpu with the documented warning both on `auto` and explicit `cuda`. So the fallback path is proven; **actual GPU execution needs validation on NVIDIA hardware, which this machine doesn't have**.
- Quiet gate: `probeListenerCount()` verified live — `null` against an unreachable Icecast (fail-open) and a correct count against a stubbed status endpoint; pause/resume decision branches are unit-pinned.

## Ask for @Jaz666

Could you validate the CUDA flavour from this branch on your NVIDIA host?

```bash
git fetch && git checkout feat/1099-analyzer-cuda-quiet-times
# rebuild ALL services — the quiet-times gate lives in the controller image
# and its toggle in the web image, not just the analyzer
docker compose -f docker-compose.yml -f docker-compose.analyzer-gpu.yml up -d --build
docker exec sub-wave-analyzer /opt/analyzer/venv/bin/python -c "import torch; print(torch.cuda.is_available())"
```

(Note the venv path — the earlier `python3 -c "import torch"` probe failed because torch lives in `/opt/analyzer/venv`, not system python.) Then run a rescan and watch `nvidia-smi` — the analyzer logs `analysis device: cuda (<GPU name>)` on first model load.

## Field-test follow-ups (2nd push)

Overnight testing on real NVIDIA hardware surfaced two problems, both fixed:

- **Quiet gate now reacts mid-scan.** The toggle was read once at pass start; a multi-hour scan never saw a mid-run flip. The gate now re-reads `settings.json` on every check, so flipping it lands within one track / one 30s poll. (The original test instructions above also only rebuilt the analyzer — the gate ships in the controller image; corrected.)
- **Idle GPU release.** The cuda worker dropped nothing from VRAM between passes, starving co-resident TTS/LLM. It now unloads CLAP + Demucs after 5 idle minutes (`ANALYZE_IDLE_UNLOAD_S`, `0` disables) and lazy-reloads on the next request. ~300–500MB of CUDA context remains until the container stops (torch limitation).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -122,6 +122,9 @@ SITE_URL=
 #   docker compose -f docker-compose.yml -f docker-compose.analyzer-gpu.yml up -d
 # ANALYZE_DEVICE=    # auto (default) / cpu / cuda — torch device for CLAP/Demucs;
 #                    # only meaningful on the cuda analyzer flavour
+# ANALYZE_IDLE_UNLOAD_S=  # cuda flavour: seconds of no analysis before models are
+#                         # dropped from VRAM (default 300; 0 = keep resident).
+#                         # Frees the GPU for co-resident TTS/LLM between passes.
 #
 # Runtime flags — env wins ON over the admin toggles (settings.audio.*), never
 # off. A flag with no matching backend in the image is a clean no-op.
--- a/.env.example
+++ b/.env.example
@@ -117,10 +117,19 @@ SITE_URL=
 #                    # set DOCKER_DEFAULT_PLATFORM=linux/amd64 (runs emulated).
 #                    # Unraid one-click (AIO) users instead pull subwave-aio-heavy.
 #
+# NVIDIA GPU? The heavy stack can run on CUDA instead — not an .env toggle (a
+# GPU device reservation can't live here); layer the analyzer-gpu overlay:
+#   docker compose -f docker-compose.yml -f docker-compose.analyzer-gpu.yml up -d
+# ANALYZE_DEVICE=    # auto (default) / cpu / cuda — torch device for CLAP/Demucs;
+#                    # only meaningful on the cuda analyzer flavour
+#
 # Runtime flags — env wins ON over the admin toggles (settings.audio.*), never
 # off. A flag with no matching backend in the image is a clean no-op.
 # ANALYZE_AUDIO_EMBEDDING=   # 1/true = fill CLAP audio vectors (needs ANALYZER_HEAVY)
 # ANALYZE_VOCAL_ACTIVITY=    # 1/true = fill Demucs vocal ranges (needs ANALYZER_HEAVY)
+# ANALYZE_QUIET_ONLY=        # 1/true = pause analysis while anyone is listening;
+#                            # resumes after settings.audio.analyzeQuietMinutes
+#                            # (default 10) with no listeners (#1099)
 #
 # Building from source instead of pulling? `docker compose build analyzer` with
 # ANALYZER_HEAVY=1 bakes the stack (or pass WITH_CLAP=1 / WITH_DEMUCS=1 directly).
--- a/.github/workflows/publish-images.yml
+++ b/.github/workflows/publish-images.yml
@@ -106,6 +106,18 @@ jobs:
             build_args: |
               WITH_CLAP=1
               WITH_DEMUCS=1
+          # CUDA analyzer — the heavy stack on cu124 torch wheels for NVIDIA
+          # hosts (#1099). amd64-only like heavy. Selected by the
+          # docker-compose.analyzer-gpu.yml overlay (which also adds the GPU
+          # device reservation), not by an .env var. onnxruntime stays the CPU
+          # build — only the transformers CLAP path and Demucs run on the GPU.
+          - image: subwave-analyzer-cuda
+            dockerfile: docker/Dockerfile.analyzer
+            platforms: linux/amd64
+            build_args: |
+              WITH_CLAP=1
+              WITH_DEMUCS=1
+              WITH_CUDA=1
     steps:
       # fetch-depth: 0 pulls full history + tags so `git describe` resolves the
       # version on a workflow_dispatch from a branch (a tag push already has the
--- a/README.md
+++ b/README.md
@@ -166,6 +166,15 @@ That repoints the `analyzer` service at `subwave-analyzer-heavy` (CLAP + Demucs,
 also offers it, and Unraid one-click users pull the `subwave-aio-heavy` image
 instead. Only the expressive *voices* above need the separate `tts-heavy` sidecar.
 
+Hosts with an NVIDIA GPU can run the heavy stack on CUDA instead — layer the
+`docker-compose.analyzer-gpu.yml` overlay, which swaps the service to the
+`subwave-analyzer-cuda` image and reserves the GPU (needs the NVIDIA driver +
+Container Toolkit, nothing else):
+
+```bash
+docker compose -f docker-compose.yml -f docker-compose.analyzer-gpu.yml up -d
+```
+
 ### Local dev (contributors)
 
 ```bash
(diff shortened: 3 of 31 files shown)


Jaz666:
@perminder-klair thank you so much.
I can confirm Audio Fingerprinting backfill is speeding though, the GPU cycles to 100% for 1 second, then 2 seconds off, and the process is using 1560MiB of VRAM.

Demucs, also working, a bit lighter on GPU demand, but analyzed 25 tracks in the time it used to take to do 1 on the CPU.

I shall pop ANALYZE_QUIET_ONLY into the .env and give it a whirl (currently pulling the update you've just released).

Jaz666:
@perminder-klair ~Can confirm starting a scan whilst the station was on air with a listener delayed the scan from starting.~
There needs to be an indication that the scan has been paused in the Library screen, and the timeout definitely needs to be user defined (I have Ollama currently set to a 30 minute timeout, so would match that in the analyzer).

Question for you, would there be any benefit from binning the audio fingerprinting and demucs done on CPU so far (30% of my library), or would the data returned be the same whatever the underlying architecture?

Jaz666:
If you read what was here before, disregard - I think testing this in the middle of the V0.44 release has complicated things
Fully upgraded to 0.44, then repulled, back in business.

Let me know if you need more testing tomorrow.

Jaz666:
OK, longer test when I woke up this morning.
Library scan was nicely running overnight, but tuning in to the station did not pause the scan, even after waiting a few songs. This resulted in OOM for my TTS, and Piper Fallback.

Manually stopped the scan, and everything ran smoothly.

A couple of hours later and I got another Piper fallback with no library scan running, due to TTS memory again.
Checked nvidia-smi, and the analyzer was still loaded into VRAM.

perminder-klair:
@Jaz666 thank you for this! really appriciate

perminder-klair:
Thanks for the overnight test @Jaz666 — both reports were real, and both are fixed on the branch as of `d7582a66`.

**Scan not pausing.** Two things bit you at once. My test instructions only rebuilt the `analyzer`, but the quiet-times gate lives in the **controller** image (and its toggle in the **web** admin) — so unless you rebuilt those too, the gate wasn't even running. On top of that there was a real gap: the toggle was only read at the *start* of a pass, so enabling it while a scan was already hours in did nothing. Fixed — the running pass now re-reads the toggle (and the idle-minutes value) from disk on every track, so a mid-scan flip pauses within ~30 seconds of the current track finishing.

**Analyzer squatting in VRAM.** Correct read. The worker kept CLAP + Demucs resident forever — fine on CPU, rude on a shared GPU. It now drops the models out of VRAM after 5 idle minutes and lazy-reloads them on the next request (`ANALYZE_IDLE_UNLOAD_S` in `.env` tunes the window, `0` disables). One caveat: ~300–500MB of CUDA *context* stays until the container restarts (a torch limitation), but the multi-GB model memory is returned. Nice side effect: when the quiet gate pauses a scan, the GPU frees up for your TTS about 5 minutes later.

To re-test, rebuild **all** services this time (sorry for the earlier bum steer):

```bash
git pull
docker compose -f docker-compose.yml -f docker-compose.analyzer-gpu.yml up -d --build
```

Then enable **Quiet times** on `/admin/library` (off by default), start a scan, and tune in: the run view should flip to "Waiting for quiet" within ~30s of the current track finishing, and `nvidia-smi` should show the analyzer's memory drop about 5 minutes after that (look for `idle 300s — released CLAP + Demucs from GPU memory` in `docker logs sub-wave-analyzer`).

Issue #629, opened 2026-06-26
fix(scheduler): respect active show genre on fallback playlist when pause-when-empty


MMagTech:
> **Context:** Filed by @MMagTech from [Millennial FM](https://millennialfm.mmagtech.com) — a real operator scenario, not a code review transplant. Related to #628 (picker diversity should respect show genre) but this is a **different code path**: the Liquidsoap fallback playlist, not the AI picker pool.

## User story

**As an operator** running a **genre-themed scheduled show** (e.g. hip hop with **genre strict** on) with **Pause when empty** enabled, I expect the stream to stay on-brand even when nobody is listening — same as when the AI is picking.

**As a listener** who tunes in mid-show, I don't want to hear several off-genre tracks (e.g. country during a hip hop hour) before programming recovers.

## Problem

When listener count is **0** and `settings.llm.pauseWhenEmpty` is on:

- AI picks and DJ speech stop (intended).
- Music continues from **`auto.m3u`**, rebuilt by `refreshAutoPlaylist()` in `broadcast/scheduler.ts`.

That fallback pool uses **dominant mood**, Navidrome playlists matching mood name, recently-added / frequent / starred / **random** — it does **not** apply the active show's:

- `genre` / `genreStrict`
- `fromYear` / `toYear`
- `energy`
- show `topic` brief

So during a **scheduled strict hip hop hour with zero listeners**, the station can air **country or other off-genre** music. When listeners return, picks only resume on the **next track transition** (and listener count polls ~every 15s), so **1–3 more fallback tracks** often play before the first AI pick — making the show feel "off schedule" even though the grid slot is still correct.

**Genre strict** on the show only affects the **AI picker** (`music/picker.ts` / agent) when picks are running; it does not filter `auto.m3u`.

## Observed behaviour (Millennial FM)

1. Hip hop show scheduled with genre strict.
2. Listeners drop to 0 → pause-when-empty silences AI.
3. Fallback music includes non–hip-hop (e.g. country).
4. Operator tunes in during the same show hour → several tracks still off-genre before hip hop programming stabilizes.

## Proposed solution

When `resolveActiveShow()` returns a show for the current hour, **`refreshAutoPlaylist()` should build the fallback pool using the same music steers as the picker**, at minimum:

- **Genre strict:** fallback tracks from that Navidrome genre only (mirror `preferGenre` / `getSongsByGenre` logic); no unfiltered random filler for strict shows.
[...]

PR #1116, opened 2026-07-21
fix(shows): one-directional, boundary-aware genre matching for strict shows

perminder-klair:
Reported on Discord: a strict emo/pop-punk show kept drifting into broader genres.

## The bug

`genreMatches` (`controller/src/music/show-filter.ts`) matched a substring in **both** directions:

```js
gn === target || gn.includes(target) || target.includes(gn)
```

That last clause means a track tagged merely `Pop` satisfies a strict `Pop Punk` show, and `Rock` satisfies `Alternative Rock`. Verified by running the shipped function:

| track tag | show genre | before | after |
|---|---|---|---|
| `Pop` | Pop Punk | ✅ leak | ❌ |
| `Rock` | Alternative Rock | ✅ leak | ❌ |
| `Emo` | Emo Pop | ✅ leak | ❌ |
| `Alternative` | Alternative Rock | ✅ leak | ❌ |
| `Trap` | Rap | ✅ leak | ❌ |
| `Pop Punk` | Pop | ✅ | ✅ |
| `Punk Rock` | Punk | ✅ | ✅ |
| `Contemporary R&B` | R&B | ✅ | ✅ |

The reporter caught the reverse direction. The `Trap`/`Rap` row is a second false positive they didn't spot: the direction we *keep* was doing raw substring with no word boundaries, so `Rockabilly` also matched a `Rock` show.

## The fix

Matching is now one-directional and boundary-aligned: a track's tag may **refine** the show's genre, never broaden it, and the show's genre must land on word boundaries of the original tag. Separators are still normalised away, so `Pop-Punk` still matches `Pop Punk` and `Hip-Hop/Rap` still matches `Rap`.

All strict paths funnel through `genreMatches` — `preferGenre`, `onlyGenre`, `applyStrictLocks`, and the scheduler's auto-playlist `enforce()` — so this is a single-point fix. The soft path tightens too; it never-starves, so it can only get more accurate.

## Deliberately not fixed here

`subsonic.resolveGenreName` (`music/subsonic.ts:276`) carries the **same** reverse-substring, one layer upstream. A show configured `Pop Punk` on a library with no such tag resolves to the tag `Pop`, and then every Pop track matches *exactly* — so this PR alone doesn't help that operator.

It needs a fetch-hint/match-target split rather than a one-line tightening: the resolved tag is also the genre passed to `getRandomSongs`, and an unresolved strict genre currently **disables the filter outright** (`scheduler.ts:392`, acceptance criteria #629) precisely because `enforce()` hard-drops per source with no final never-starve — writing an empty `auto.m3u` would drop the station onto the emergency single. Worth its own PR. Loose resolution is also correct on the listener-request path (`routes/request.ts:141`), where someone asking for "punk rock" should still get punk.

## Verification

- `npm test` — all 41 test files pass
- `npm run lint` — exit 0, no new warnings
- 3 new test cases in `scripts/show-filter.test.ts` pinning refine / no-broaden / word-boundary; the old test asserting "substring both ways" is replaced


perminder-klair:
## Validated against a real 10,416-track library

Replayed both the old and new predicate over a real `library.db` (read-only copy). Per-show matched-track counts:

| show genre | before | after | Δ | what was dropped |
|---|---|---|---|---|
| `Pop Punk` | 887 | **0** | -887 | all 887 were plain `Pop` |
| `Indian Pop` | 1944 | 928 | -1016 | `Pop`, `Indian` |
| `Trap` | 134 | 27 | -107 | `Rap`, `West Coast Rap` |
| `Rap` | 230 | 203 | -27 | `Trap` |
| `R&B` | 352 | 346 | -6 | `Urbano latino` |
| `Punjabi` | 1212 | 1212 | 0 | — |
| `Hip-Hop` | 3092 | 3092 | 0 | — |
| `Pop` | 2509 | 2509 | 0 | — |

Two things this confirms:

- **Broad shows are untouched** (`Punjabi`, `Hip-Hop`, `Pop` all unchanged), and they keep their refinements — `Punjabi` still includes all 618 `Punjabi Pop` tracks, `Rap` still includes `West Coast Rap`/`Alternative Rap`/`Pop Rap`. The fix only tightens narrow shows, which is the intent.
- **`R&B` → `Urbano latino`** was a live false positive nobody had noticed: `"urbanolatino"` contains `"rb"`. That's the forward direction, and it's the word-boundary half of the fix earning its keep. Now pinned as a regression test.

## ⚠️ This also proves layer 1 alone does NOT fix the reported scenario

Simulating `resolveGenreName` against the same library's real tag list:

```
show "Pop Punk"         -> resolves to library tag: "Pop"   ← BROADENED
show "Emo"              -> resolves to library tag: NULL (filter disabled)
show "Alternative Rock" -> resolves to library tag: "Alternative Rock"
```

Resolution runs **before** matching. So end-to-end on this library, a `Pop Punk` show still has its target rewritten to `Pop`, and those 887 tracks then match *exactly* — this PR changes nothing for it. `Emo` resolves to NULL and the strict filter is switched off entirely.

For the reporter's emo / pop-punk / alternative show that means only `Alternative Rock` is actually repaired by this PR. **Layer 2 is the bigger half of their bug, not a follow-up nicety.**

The design is now clear and the dead-air objection is tractable: match on the operator's raw text (boundary-aware) and demote the resolved tag to a *fetch hint* for `getRandomSongs`, then add a final never-starve for genre in `refreshAutoPlaylist` — the same guard the strict-playlist block and the blocklist block already use a few lines away, so it's an established pattern in that function rather than a novel risk.

Keeping that out of this PR so the predicate fix stays reviewable on its own, but it should land right behind it.

perminder-klair:
## Layer 2 landed here too — as a diagnosability fix, not the rewrite

### First, the `getGenres` unknown is closed

Checked Navidrome directly (salt+token auth, read-only) against `library.db` as ground truth:

```
getGenres returned 111 tags (status ok)
library.db distinct tags: 86 (primary 86, secondary-only 0)
tags present in library.db but MISSING from getGenres: 0
secondary-only tags missing from getGenres: 0  (getGenres DOES return secondaries)
```

So multi-value genres resolve fine. **That was the one scenario that would have made layer 2 a correctness bug, and it's disproven.**

### Which means layer 2 is smaller than I claimed

`resolveGenreName` tries **exact match first** and only falls back to substring on a miss. Against the real library:

```
"Alternative Rock" -> "Alternative Rock"   (exact — no broadening)
"Rap"              -> "Rap"                (exact)
"Indian Pop"       -> "Indian Pop"         (exact)
"Pop Punk"         -> "Pop"                ← broadened (no such tag)
"Emo"              -> NULL                 ← (no such tag)
```

Broadening happens **only when the genre isn't in the library at all** — in which case the show is unsatisfiable and no matching rule fixes it. I withdraw "layer 2 is the bigger half of the reported bug"; for any show whose genres are real tags, PR #1116's layer 1 is the complete fix.

**So I did not do the fetch-hint/match-target split.** It would have rewritten the dead-air fallback path to change *which* wrong thing happens to an unsatisfiable show — and dumping to the whole library (3090 Hip-Hop tracks here) is arguably worse than dumping to `Pop`. Not worth the risk.

### What actually shipped

The real defect is that the station **silently** airs a different show than configured. Now it says so:

- `genreResolutionWarning()` / `...Once()` in `show-filter.ts` — describes the gap in operator words, `null` when the resolution was faithful (cosmetic differences like `hip hop` → `Hip-Hop` don't warn), memoised so a standing misconfiguration doesn't reprint on every pick and hourly refresh. 4 unit tests.
- Wired into all three show-genre resolution sites — scheduler, pool picker, dj-agent. The picker routes it through the existing `strictInfo` diagnostics channel rather than a second log call, because `queue` is injected into `pickViaPool` and isn't in scope inside `buildCandidates`.
- **Show editor warns inline** when a configured genre matches no library tag, reusing the genre list already fetched for autocomplete. This kills the class before it ever reaches the station. Never warns when that fetch failed.

Untouched on purpose: `resolveGenreName` itself, and the listener-request / `songsByGenre` lookup paths, where loose matching is correct.

`controller` lint 0 · `web` lint 0 · 41/41 test files pass.

Issue #577, opened 2026-06-24
Bug: Curiosity segments repeat the same facts after controller restart and when Wikipedia pool is exhausted


geekylakshya:
# Problem

Listeners hear the same curiosity facts repeated, sometimes within the same day.

Confirmed from event logs on the test LXC (192.168.0.171):

* Kenneth Arnold / flying saucer aired at 04:00 and 08:55 on June 24 (same day, 5h gap; controller restarted between them)
* Nintendo 64 launch aired at 09:30 and 10:45 on June 23 (same day, reworded but same source event)

# Root causes

## 1. `seenCuriosity` is in-memory only and is wiped on restart

`skills/_agent.ts:199` holds the dedup set as a plain `new Set<string>()`.

Any controller restart (deploy, crash, Docker restart) wipes it. Wikipedia returns the same "On This Day" items for the same date, so identical events can be re-aired after a restart.

## 2. The Wikipedia pool is small and exhausts quickly

`curiosity.ts:fetchOnThisDay()` filters to roughly 8 allowed items per day (cultural/scientific events since 1850).

The `getCuriosityItem` tool consumes 3 items per call (`segment-tools.ts:84`), so the pool is exhausted after roughly 3 curiosity segments.

After that, all calls return:

```json
{
  "available": false
}
```

## 3. The LLM fallback path has no deduplication

When the tool returns `available: false`, the agent falls through to pure generation under `cap.desc`.

There is no history of previously generated curiosity content, so the same persona and time-of-day context can lead the model to generate the same facts repeatedly.

# Suggested fixes

## 1. Persist `seenCuriosity` to disk

Store the dedup history in `state/seen-curiosity.json`, similar to how `_recentPlays` is persisted in `recent-plays.json`.

On startup:

* Load the persisted set
* Prune entries older than 7 days
* Save updates whenever new curiosity items are aired

This prevents repeats across controller restarts.

## 2. Feed recent curiosity history into fallback generation

When `available: false`, append the last 5-10 aired curiosity texts to the agent context.

This gives the model explicit examples of recently used content and helps it avoid generating the same facts again.

perminder-klair:
thank you @geekylakshya <3

Issue #310, opened 2026-06-05
DJ sometimes talks over themselves


WhatKateDoes:
These two speech segments intersected.  Happens sometimes, particularly if DJ is feeling chatty:

22:00:44
segment·station-id
The static hums low across the hull tonight, a soft reminder that we are broadcasting from somewhere very far out. This is The Ninth House, where the shadows stretch long and the blues linger in the dead air of the void. Tune in if you can find the signal among the wreckage.
22:01:05
segment·hourly-check
It's nearly the witching hour out here in the black velvet, darling. The deep silence of this void stretches out long and heavy as we drift along. Time is just a slow, dragging shadow tonight.

perminder-klair:
Thanks for the detailed report, the timestamps made this easy to pin down.

Fix is up in #312. Verified that:
- the station ID can no longer land on :00 alongside the hourly check, and
- with the old code a link aired 0ms after a 2s station ID started (the overlap), while the new code makes it wait the full duration first.

Will close this once #312 merges. Thanks again for flagging it!

PR #1148, opened 2026-07-23
feat: pause-then-talk breaks for long DJ segments (#551)

### LINKED ISSUE #551: DJ Talking over
So DJ likes to talk over current song playing, it's really annoying.

Would be nice to have a setting to force DJ to wait until AFTER a song has finished before doing news / weather / curiosity stuff

uhhuhyeah:
Closes #551 (or, at least my take on 551). Design approved in principle in [this comment](https://github.com/perminder-klair/subwave/issues/551#issuecomment-4990332332).

## What changed on air

Ducking stays the default for short asides. When a show opts into **pause-and-talk** and a spoken skill segment's rendered clip runs longer than a threshold, the segment no longer ducks over the music. The current song plays to its natural end, the DJ speaks **in the clear**, and the next track ramps in briefly under the sign-off. The clip rides the music timeline as its own annotated `subwave_kind="talk"` item through the `next.txt` seam, mirroring beds.

Per-show, **off by default**; the length cut-off (`pauseTalkMinSeconds`, default 20s) is station-wide.

## The two hazards you flagged, and where each is neutralized

1. **The crossfade would treat a speech clip like a song.** Handled by padding silence into the clip (`audio/wav-pad.ts`) sized to the predecessor's exit canvas, plus the clip's own short `liq_cross_duration` for the ramp into the next song. The house rule (fade == buffer, vary the buffer) is kept intact. The predecessor's exit canvas is already sent when a segment fires mid-song, so padding is the deterministic, testable option.
2. **A jingle stinger could land next to it.** `radio.liq` gains a `talk_on_air` ref, and the jingle rotate gates on it (`{not bed_on_air() and not talk_on_air()}`). No stinger between the break and the song it leads into. A stinger *before* the break stays possible (rare, idiomatic; accepted for v1).

## Shape

- **Pure + tested** (`broadcast/pause-talk.ts`, unit-pinned in `scripts/pause-talk.test.ts`): gap-vs-duck decision, pad math, sanity guard, the talk URI builder. `audio/wav-pad.ts` writes a *new* padded copy (original kept for a ducked fallback); best-effort, PCM-only. **Cloud (mp3) voices always duck**, logged.
- **Queue** (`broadcast/queue.ts`): `announce()` gains `gapEligible`; a long clip is held in one `_pendingPauseTalk` slot (newer supersedes older, air-time bookkeeping deferred like `_pendingVoice`); `maybePushPauseTalk()` injects it in `drainToLiquidsoap` *before* `maybePushBed`, strips the following track's entry effects + link, and falls back to ducked air on any failure; a coasting flush covers the auto-playlist-only case.
- **Liquidsoap** (`radio.liq`): an `on_meta` branch before the title gate writes `talk-playing.json` and, like a bed, never touches now-playing / ICY / history / scrobbles.
- **Opt-in call sites**: only the two skill-segment `announce()` calls pass `gapEligible: true`, so operator skills qualify automatically while links/banter/programmes/hourly/station-IDs/requests always duck.
- **Admin UI**: per-show toggle + the shared threshold input. Docs in `docs/pause-talk.md`; CLAUDE.md updated.

## Notes for review

- `radio.liq` changed, so testing needs a **broadcast rebuild** (`up -d --build broadcast`; dev: `restart broadcast`).
- Settings: `pauseTalk` is a plain per-show boolean and `pauseTalkMinSeconds` a clamped `[5,90]` number. **No mixer restart** (nothing in `radio.liq` reads the threshold).
- `npm run lint` (controller + web) is clean; `npm run test:pausetalk` (and the full `npm test`) pass.

## Smoke-tested on air

Ran the dev stack against a real library and confirmed the full path end to end: a 31s segment on an active pause-talk show cleared the threshold → held → padded to 43s → injected into the timeline → aired in a real gap with the next track ramping in under it, `talk-playing.json` written, now-playing untouched.

## Screenshots
Screenshot 2026-07-23 at 13 00 38

Screenshot 2026-07-23 at 13 01 31


SkinnyDennis:
My hack for accomplishing this was to record a three minute silent bed for the DJ to talk over and leave it as the only bed. It works great and is giving me the radio vibe I want for my djs. I'd love to be able to assign different beds to different shows (including silent beds like this), but I don't have the skills to add that to the codebase cleanly.

perminder-klair:
Closing this one, but not the idea — #551 stays open and the core of what you built here is what I'd want the re-land to start from. Below is what changed underneath the branch and what the rebuild looks like, so you're not reverse-engineering it from a diff.

## Why closing rather than rebasing

The branch is 256 commits behind `develop`. A test merge leaves ~4,300 conflicted lines across 9 files, but the line count isn't the real problem — two of the files you touch were split into directories while this sat open, so the biggest hunks are re-authoring, not conflict resolution.

| | at your merge base | on `develop` today |
|---|---|---|
| `controller/src/settings.ts` | 4,928 lines, monolith | 2,163 lines, split into `settings/{defaults,normalize,validate,patch-registry,store,…}`, plus a zod show schema in `schemas/show.ts` mirrored to `web/lib/schemas.generated.ts` (CI diffs the mirror) |
| `controller/src/broadcast/queue.ts` | monolith | split into `queue/{pure,types,voice-io,voice-marker,intro-render,kinds}` |
| `announce()` | inline bookkeeping | rebuilt around `voiceChannelFor` + `onQueued`/`onSpoken`, with real air-time stamps off `voice-playing.json` (#1382) |
| `runCapability()` | returns `string` | returns `{ aired, text, reason }`, consumed by `programme.ts` |
| talk scheduling | four independent crons | one `* * * * *` tick over `talk-scheduler.ts`'s slot table |
| IPC docs | root `CLAUDE.md` prose | `docs/internals/broadcast.md` + terse root invariants |

`settings.ts` alone is 2,508 conflicted lines in 3 hunks (1470/722/316) — the functions you patched (`normalizeShows`, `validateShowsStrict`, `resolveShowShape`, the `update()` branch) all live in different files now. `ShowsPanel.tsx` is another 1,139. Resolving that by hand is slower and more error-prone than re-applying ~200 lines of logic to the new layout.

## The design change, which matters more than the conflicts

`e2d49d4a` (#1505) landed one arbitrated talk-slot scheduler. Every scheduled spoken segment is now a row in one table driven by a single per-minute tick, and one of the three rules holding it together is **in-flight talk counts**: `scheduler.ts:960` feeds the planner `pendingTalk: pendingTalkOf(queue.pendingVoiceKind())`, because `getLastTalkBreakAt()` reports what has *aired* and a boundary-deferred ident is talk that hasn't. That blind spot was #1419's root cause.

A pause-talk hold is exactly that kind of talk, on a second pending slot — `_pendingPauseTalk` — that `pendingVoiceKind()` doesn't read. So:

1. `:12` — the segment director (the table's `fill` row) produces a 35s news read. `holdForPauseTalk` takes it. Nothing aired, so `getLastTalkBreakAt()` is unchanged; `pendingVoiceKind()` is still `null`.
2. `:15` — the `station-id` row plans, sees a clear 180s quiet gap and no pending talk, fires with `air: 'next-track'`. The ident lands in `_pendingVoice` for the next track boundary.
3. That's the same boundary the drain injects the break at.

Listener hears: song ends → 35s news in the clear → next track starts → ident ducked over its opening. That's #310 again, on the longest segments — the ones the gap rule most exists to protect.

The conclusion I'd draw: pause-and-talk has to be a **delivery mode the scheduler knows about**, not a path around it. Concretely that means `announce()` returns an air/hold status instead of `void`, `runCapability` propagates it into the `{aired}` contract develop already has, and the hold is visible to the planner alongside `_pendingVoice`.

## The voice chain, and a mechanism worth considering

`radio.liq:1211-1212`:

```
voice = mic_chain(amplify(override="liq_amplify", 1., edge_fade(voice_queue)))
intro = mic_chain(amplify(override="liq_amplify", 1., edge_fade(intro_queue)))
```

`mic_chain` (compress → makeup → 40ms slap echo) and `edge_fade` apply to those two queues only. A clip riding `dj_queue` reaches the broadcast bus with nothing but the brick-wall limiter, which sits at ~0 dB on this material. So the break — the one moment the feature exists for, the DJ alone in the clear — is the only spoken segment on the station with no compression, no makeup gain, no slap echo and no edge fade. It will not match the ducked break the same DJ did ninety seconds earlier. `voiceGainDb` is a per-engine/persona dB trim; it doesn't stand in for the chain.

**Recommended direction, not a requirement:** invert what goes into `dj_queue`. Instead of pushing the padded voice WAV onto the music timeline, push a **silence item** of the right length as the `subwave_kind="talk"` entry, and fire the voice into `voice_queue` at that boundary. You get the real gap (there's nothing under it to duck), the voice keeps its full chain, and `say.txt` / `voice-playing.json` / the `onSpoken` air-time stamps all keep working instead of needing a parallel bookkeeping path — which also dissolves most of finding 5 below.

The honest tradeoff: the silence length now has to track the rendered clip length, and a mistimed silence is worse than a mistimed pad — too short and the voice gets clipped by the incoming track, too long and you've bought dead air. `wav-pad.ts` is already computing exactly that number, so the arithmetic exists; the risk moves from "does the clip sound right" to "do the two land together". It's your implementation and you may see a reason this doesn't hold up — but the current routing has a real audible cost and I'd want that weighed before it ships.

## What I'd carry over largely intact

- `broadcast/pause-talk.ts` — the pure helpers are the right decomposition and the unit test is the right shape.
- `audio/wav-riff.ts` — a faithful extraction of the existing parser, not a rewrite. Good change on its own merits.
- `audio/wav-pad.ts` — still needed for the head/tail math whichever mechanism wins.
- `docs/pause-talk.md` — accurate and well-pitched.
- The `talk_on_air` ref and the jingle gate in `radio.liq`. That gate is the correct answer to the stinger hazard: jingles play outside the controller's voice serialiser and `music_meta` is captured above the rotate, so nothing downstream can hold a stinger back once it's in the cross buffer. Making the jingles source *unavailable* for the duration is the only place that decision can be made, and mirroring `bed_on_air` means it fails the same way beds already do.

## Other findings for the re-land

- **`announce()` returning `void`** — callers book a held clip as aired. `_agent.ts:537` writes the durable curiosity ledger (#577) with `aired: true` for a clip that may still be superseded or dropped stale, permanently suppressing a line nobody heard. `playSfx(…, { underVoice: true })` fires alone over the music minutes early, divorced from its segment. `POST /dj/skill` answers 200 for something the operator never hears. And on today's `develop`, `programme.ts:311` calls `runCapability` for feature beats that are marked aired *before* generation — a held beat that goes stale is silently lost for that episode, and a restart won't retry it.
- **`radio.liq:1446`** hardcodes `/var/sub-wave/talk-playing.json` where every other path in the file uses `#{state_dir}` (resolved from `SUBWAVE_STATE_DIR` at line 20). On a multi-station install the controller reads `${STATE_DIR}/talk-playing.json` under `state/stations/<id>/` and the mixer writes to the install root, so they never meet. It also needs classifying in `stations/pure.ts` (`conversionAction`/`duplicateAction`) as a new state file.
- **`maybePushPauseTalk` needs a `stemSeam` guard.** When the predecessor drained with a stem blend armed, the clip already carries the *incoming* track's identity (now-playing flips at blend start) and that track is cued in past its pre-mixed head. Wedging a break between them gives you: predecessor cuts at `blendStartSec` → blend clip plays under the next song's name → 40s break → the song starts headless. `applyMixTransition` already strips entry effects on `item.stemSeam` for the same reason.
- **The coasting flush is a second, unsynchronised writer of `next.txt`.** It runs from `onTrackStarted` outside the drain's `senderBusy` section, samples the busy flags once and then awaits ~1.5s in `writeHandoff`. A `POST /request` landing in that window can end up behind the break — which is precisely the case `maybePushPauseTalk` refuses by design. The silence-item approach above would let this live inside the drain instead.

## One question

Do you want to take the re-land on? The pure core plus the mixer reasoning is the bulk of the thinking here and I'd rather it stayed yours. If you'd rather hand it off, say so and I'll pick it up from this branch — no hard feelings either way, and #551 stays open as the tracking issue regardless.

uhhuhyeah:
Replaced by #1645

PR #1149, opened 2026-07-23
feat(web): accessibility & UI-fundamentals pass (shadscan 41→72), plus landing broadsheet polish

perminder-klair:
## What

Two web-quality workstreams:

1. **Landing broadsheet AI-tell cleanup** (original scope) — reduces generic "AI-design tells" on the marketing/landing broadsheet (asymmetric DJ-habits lead-plus-pair layout, tighter transitions, small copy/markup trims across `what/*` + `globals.css` + `StationFooter`).
2. **Accessibility & UI-fundamentals pass** — remediation of a static **shadscan** audit across the web app.

## Shadscan result

**41/100 (F) → 72/100 (C)** — verified by re-running the pinned `npx --yes @shadscan/cli@0.1.1 --json`. 9 findings flipped to **pass**, 4 dropped fail → advisory (score-neutral), no regressions.

### Passed
- `error-state-retry-present`, `route-loading-boundary-present` — recognized retry control on the admin error boundary; one root `app/loading.tsx` (+ `app/landing/loading.tsx`) gives *all* runtime-dynamic routes loading coverage.
- `theme-provider-configured`, `theme-provider-mounted-in-shell`, `theme-hotkey-present` — explicit light/dark override on the app-shell `ThemeProvider` (`data-theme` + `.dark` sync, pre-paint script), a guarded bare-`d` hotkey that ignores typing targets, and a visible toggle in the appearance menu.
- `toast-provider-present`, `toast-provider-mounted` — the existing Sonner toaster is promoted to the root shell (deduped from the player/admin shells) so every route can surface `notify()`.
- `global-hotkeys-are-safe` — bare-key handling is now unreachable before the text-entry guard.
- `personal-data-autocomplete-present` — `autoComplete` on all password/secret fields.

### Downgraded fail → advisory (no longer costing score)
`icon-buttons-have-labels`, `custom-controls-have-labels`, `invalid-fields-associated-with-errors`, `button-icons-have-data-icon` — accessible names / associations added to the reported controls and swept across the admin tree.

### Also added
- **Admin command menu** — Cmd/Ctrl+K palette in `AdminShell` (reuses the existing cmdk `CommandDialog`) to jump between panels.
- Accessible names on inputs / selects / switches / icon buttons, `aria-describedby` wiring for invalid fields, `role=alert`/`role=status` live regions, and keyboard-dismissible (Escape) modals across the admin panels.

## Remaining shadscan findings (kept visible, with rationale)

- **`command-menu-present` / `-hotkey-present`** — *product decision:* command menus are shipped per-surface (player palette in the classic skin; the new admin Cmd/Ctrl+K menu). No single "app-level" mount by design — player and admin are distinct command surfaces and marketing pages don't warrant one. The static detector only credits one app-shell mount.
- **`mobile-nav-present`** — *waived:* the flagged `<nav>` is the news-article Newer/Older pager, already accessible-named and fully visible at 320px; nothing to collapse.
- **`async-action-pending-state`**, **`status-messages-announced`** — the pending/announced behavior lives in shared components (`SignInForm`'s disabled trigger; `ErrorState → V3Alert` which carries `role="alert"`, plus Sonner). Real behavior is correct; the static detector can't trace across the component boundary.
- **`focus-visible-not-suppressed`** — the one genuine defect is fixed; remaining flags are elements using the `.v3-focus:focus-visible` utility (a real focus ring the scanner can't resolve statically).
- **`forms-have-labels`**, **`validation-wired-to-form`**, **`interactive-elements-are-semantic`** — substantially improved (every genuinely nameless control labeled; native validation on the real forms; modal backdrops made keyboard-dismissible). The stragglers are visible-but-not-programmatically-associated `<Label>`/`<Input>` pairs, a vendored free-text chat form, and a `stopPropagation` island in a clickable card — left as a follow-up systematic pass rather than churn.

## Gates
`npm run lint` (`eslint . && tsc --noEmit`) · `npm run typecheck` · `npm run build` — all green. Accessibility changes are additive: no behavior or layout changes.


perminder-klair:
Pushed `87366cf4` with the review fixes. Summary of what changed and why, in severity order.

### Blocking

**1. The light/dark override couldn't repaint surfaces.** `applyTheme` writes the palette as *inline* styles on `<html>`, so an inline `--bg` beats `:root[data-theme="dark"]`. Flipping the attribute alone left surfaces at the palette's own mode while `dark:` utilities and the `[data-theme="dark"]` grain/frame rules flipped underneath them — a half-dark page. Not visible on an empty registry, but `applyEffective` falls back to `registry[0]`, so in practice a palette is nearly always active.

Reproduced against the pre-fix code before changing anything:

```
OLD  --bg inline : #faf7f0   <- light palette, still inline
OLD  data-theme  : dark
OLD  .dark class : true
```

Palette and mode are now resolved together by a pure `resolveAppearance`. A pinned mode **keeps** the palette when it was authored for that mode and otherwise **pauses** it in favour of the built-in base (globals.css ships a complete light/dark pair). Pausing rather than recolouring, and rather than silently substituting a different palette — that would leave the picker highlighting a row that isn't on screen. `applyMode` / `clearMode` / `clearThemeTokens` are no longer exported, so nothing can flip the attribute without clearing the tokens first. The pre-paint script mirrors the same rule, so there's no first-frame mismatch.

**2. `d` raced the tune-in gate.** Every skin's gate accepts any printable key to tune in, and that tap is the browser's audio-unblock gesture. Both listeners sit on `window` and registration order is mount-order dependent — skins are lazy chunks, so they can register *after* the root provider. The handler no longer calls `preventDefault` and settles on the next task instead of acting inline; if any other listener claimed the press it stands down. Tuning in always wins. It also now ignores keystrokes aimed at Radix typeahead popups (`role="listbox"`/`"menu"`/…), which are divs rather than `<select>`.

**3. Modal backdrops.** `role="button" tabIndex={0}` put a full-viewport control in the tab order and wrapped the dialog in a role whose children are presentational; the Escape handler on it never fired because focus was never moved into the dialog. Backdrops are plain divs again, panels are `role="dialog" aria-modal="true"` with a label, focus moves in on open and back to the opener on close, and Escape stays with the document-level handler that was already there.

### Medium

**4.** Dropped `minLength={2}` from library search — it rejected legitimate one-character queries (an album called "1", an artist filed under "M") with a native validation bubble. `required` stays.

**5.** API keys, bearer tokens and session tokens carried `autoComplete="current-password"`, which tells password managers those boxes are this site's login — offering to fill it, and to *update* the saved password on submit. All opaque secrets are `off` now; the station password (being set, not entered) is `new-password`. Real login fields — admin sign-in, the station gate, the Navidrome step — are untouched.

**6.** `toggleMode` could only move between light and dark, so one stray `d` detached a listener from the operator's palette for good. `cycleMode` releases the pin whenever auto would render the mode being asked for, so the common flow is still a two-press flip but always parks back on auto. The switcher row is three-state (AUTO / LIGHT / DARK) and says when a pin has paused the palette.

**7.** `BlockMenu` drops `role="menu"`/`role="menuitem"` — that pair is a contract for arrow-key roving focus, Home/End and focus-into-first-item, and announcing a menu that only answers to Tab is worse than the plain buttons it would otherwise be. Kept the outside-dismiss (now `pointerdown`, so touch works, plus `focusin`) and added focus-return to the trigger on Escape.

### Nits

`Toggle`'s `ariaLabel` is now **required** so the sweep can't go partial again — tsc found the three call sites that were missed. `SignInForm` owns `aria-busy` on its own `<form>` (dropping the parallel pending state in Observatory) and the no-op `minLength={1}`. `ThemeBootstrap.tsx` → `ThemeProvider.tsx` so the file, the export and the import identifier agree. One dark-class writer instead of three. Root `loading.tsx` uses `--muted` rather than a shadcn token.

One review note withdrawn: the `bs-eyebrow` split is coherent — the removals were per-card kickers, the ones kept are section kickers ("PART TWO · THE DJ"). Left alone.

### Verification

`npm run lint` (eslint + tsc) green. The theme path is the risky part and isn't lint-visible, so I exercised both halves against a fake DOM — `resolveAppearance`/`applyAppearance` across the palette × mode matrix, and the pre-paint script string across the cache × override matrix, including the specific regression above. All green, and the two agree with each other.

Not done, flagging rather than assuming: web has no test runner, so those checks were throwaway. `resolveAppearance` is pure and exactly the kind of thing `controller/scripts/listener-auth.test.ts` pins on the controller side — worth a follow-up if you want a web equivalent wired into CI.

PR #1193, opened 2026-07-26
feat(analyze): backfill the stem cache, and let the budget be edited

perminder-klair:
Two reports from the Discord thread on stem transitions, both of which check out.

> **JDizzy:** Is there a way to Backfill the Stem scan? Like the regular acoustic analysis? At this moment all my songs are scanned the acoustic way. How can I use the Stem scan without effecting my current scans?

> **Jaz666:** the advice it gave about setting `audio.stemCacheGb` seems to be false, set mine to 100GB, and the folder is sat dead on 15GB (the default). Ultimately if you re-scan your whole library, the way the code works today, it will only ever cache the last 600 songs.

## 1. The stem cache never backfilled

`runAnalysisPass` built its scope from `needsAnalysisIds()`, widened for tracks missing a CLAP vector and tracks missing vocal ranges. There was no widening for "analysed already, but no stems on disk", so turning `audio.stemCache` on for a fully-scanned library did **nothing** — the next pass printed `nothing to analyse — all tracks current` and exited. Stems only landed for music added *after* the toggle.

The only escape was `--re-analyze`, which is exactly what JDizzy was trying to avoid: `clearAnalysis` wipes bpm, key, intro, outro, structure, beats, bars, key ranges, audio moods and vocal ranges, and runs `DELETE FROM track_audio_vectors`. It also re-clears on every invocation, so it cannot resume across nights.

The gap was structural. The audio and vocal backfills are cheap SQL. Stems live on **disk** at `state/stems/<id>/`, and the worker's `stems_cached` answer (already returned by `analyze_worker.py`) was being discarded — so there was no column to query.

**`tracks.stems_at`** (schema v17) is that column, and `db.needsStemsIds()` is the fourth widening.

### The stamp records the attempt, not disk presence

This is the load-bearing decision, and it's deliberate — same shape as `original_year_checked_at`. Two independent reasons it must not mean "stems are on disk right now":

1. A track whose stems can't be written would be re-targeted on every pass forever — the "275/7093" churn class this codebase has been bitten by before.
2. **The LRU sweep deletes stem dirs by design.** A presence-based scope would drag every evicted track back in on the next pass and never converge on any library larger than the budget.

A cache miss at seam time is already handled: the transition falls back to a plain crossfade.

### Capped at what the budget can hold

Separating thousands of tracks the sweep evicts minutes later is hours of Demucs time for nothing — the mechanism behind "it will only ever cache the last 600 songs". The widening is capped at `stemCache.headroomTracks()` and announces what it left behind rather than truncating silently:

```
[analyze] stem backfill: +614 tracks with no cached stems (6,479 left for later passes — budget holds ~614 more)
[analyze] stem backfill skipped — cache is at its 15 GB budget (raise it in Settings → Transitions to cache more tracks)
```

`clearAnalysis` gains `clearStems` so a re-analyse with the cache **off** doesn't discard stamps it isn't going to rewrite.

## 2. The budget setting was unreachable

`audio.stemCacheGb` was wired to the sweep and `update()` accepted 1–500 all along. But there was no input for it in the admin UI, and "Save transitions" sent `audio: { stemCache }` only. Since settings are held in memory and `update()` writes the whole object back, a hand edit to `settings.json` on a running controller gets clobbered by the next UI save. Hence 100 GB set, 15 GB observed.

Adds the field with a live "holds ~N tracks" readout, and `update()` now throws on out-of-range instead of silently ignoring — the old behaviour showed a saved budget the sweep never used.

## 3. Doctor coverage warning

A blend fires only when **both** tracks of a pair have cached stems, so a station with the feature on and a near-empty cache hears plain crossfades with nothing to explain why. The Doctor now reports actual coverage and distinguishes the two causes (backfill hasn't caught up vs. budget too small to hold the library).

## For anyone who wants stems today, before this merges

With the controller stopped, on `state/library.db` (back it up first):

```sql
UPDATE tracks SET analysis_version = NULL;
```

`needsAnalysisIds` keys purely off that column and each track re-stamps it on completion, so a plain `--limit N` pass re-visits the library and writes stems, resumably, without touching CLAP vectors or vocal ranges.

## Testing

- `scripts/stem-backfill.test.ts` — 11 assertions over a real SQLite DB and a temp cache dir: scope before/after a pass, the miss-is-stamped contract, COALESCE preservation, `--limit` resumption, both `clearAnalysis` modes, headroom math, the zero floor, and the `tail-meta.json` requirement.
- Full suite: **55/55 test files pass**.
- `npm run lint` clean in both `controller/` and `web/`.

Migration is additive (`ALTER TABLE tracks ADD COLUMN stems_at TEXT`). Existing installs see every track as un-stemmed and backfill from there, which is the intended behaviour. No analyzer image change — the worker already reported `stems_cached`.


perminder-klair:
Review follow-up (b52f9e6) — four fixes:

1. **Doctor can no longer blank the Tuning section.** The coverage check was the doctor's first library-db call, and the DB opens lazily on the first pick — a doctor run before that (fresh restart, stalled picker) threw `library-db not opened` and `safe()` swallowed every other Tuning finding. Now guarded with `db.isOpen()`.
2. **Backfill log is exact under `--limit`.** The announced `+N` was computed before the final `slice(0, cap)` dropped ids the bpm/CLAP/vocal scopes' slots had already claimed — a silent truncation that read as "the backfill finished". Room is now sized off `cap - ids.length`, matching the vocal widening's accounting.
3. **Doctor coverage counts stem dirs on disk**, not `stems_at` stamps. Stamps record attempts and keep growing while the sweep evicts, so the stamp count drifts optimistic and would eventually mute the warning on exactly the stations that need it. The attempt count rides along in the detail (the gap between the two numbers is diagnostic). Stamp semantics untouched — the scope still converges on attempts.
4. Nits: "third widening" → fourth, doctor derives holds-per-GB from `APPROX_TRACK_BYTES`, web UI's mirrored `/25` annotated.

New test pins the divergence contract: `cachedTrackCount()` (disk) vs `stemsCachedCount()` (stamps), and eviction dropping disk coverage. 55/55 test files, both lints clean.

PR #1210, opened 2026-07-27
fix(tts): romanize Japanese kana so the DJ stops saying "japanese letter"

### LINKED ISSUE #1179: Japanese letter japanese letter chinese letter japanese letter japanese letter
While that title sounds like someone had a stroke..I listen to fair bit of japanese music and everytime either the artist or the song title is in japanese like " ウルフルズ - バカサバイバー " for example.. the AI DJ's just says "japanese letter japanese letter"... you get the point.

I would prefer if it actually actually attempted saying it with a broken accent what the song is supposed to be than hearing it just saying multiple times "japanese letter" or "chinese letter"

perminder-klair:
Fixes #1179.

## The bug

A track like `ウルフルズ - バカサバイバー` made the DJ say "japanese letter" twelve times instead of the artist and title.

## Root cause

Not the LLM and not a model limitation, which is where the issue thread landed. Every spoken line is phonemized by espeak-ng under the station's language (`en-gb` by default), and espeak has no reading for Japanese codepoints, so it falls back to the literal character-class entry in its dictionary. Reproduced at the phonemizer inside the running controller:

```
$ espeak-ng -v en-gb -q -x "Up next, ウルフルズ with バカサバイバー"
,Vp n'Ekst
dZ'ap@ni:zl,Et3 dZ'ap@ni:zl,Et3 dZ'ap@ni:zl,Et3 ...   (x12)
```

That `dZ'ap@ni:zl,Et3` is literally "japanese letter" — the issue title, straight out of espeak.

```
$ espeak-ng -v en-gb -q -x "Up next, Urufuruzu with Bakasabaibaa"
,Vp n'Ekst
j,U@ru:fjU@r'u:zu: wID b'aka#s,abeIb,A:
```

## The fix

Hand the engine Latin letters. New `controller/src/audio/romanize.ts` converts kana to modified-Hepburn — a broken-accent approximation, explicitly what the issue asks for over the placeholder.

It goes in `normalizeForSpeech()`, the one chokepoint every booth-bound string already passes through, so **Kokoro, Piper, Chatterbox, PocketTTS and the cloud engines are all fixed at once** rather than patching one worker.

Two properties worth reviewing:

- **Speech-only.** `normalizeForDisplay()` is untouched, so the player feed, booth log and session keep the real `ウルフルズ`. Same rule the `Ye` → `Yay` correction layer follows: a listener seeing `Urufuruzu` written down would be a bug.
- **After operator corrections, deliberately.** A `settings.tts.corrections` rule is written in the original script (`ウルフルズ` → `Ulfuls`, the band's own Latin name), so romanizing first would stop every such rule matching. Corrections therefore always beat the transliteration, which is the escape hatch for names this approximates badly.

## Scope: kana only

Kanji and Han characters have no reading without a morphological dictionary — `宇多田` is "Utada" only because a dictionary says so, and the smallest Node analyzer that knows carries a ~41 MB IPADIC (`kuromoji`, CJS-only, last published at v0.1.2). Those runs **pass through untouched and still hit the espeak fallback.**

Both halves of the reported example are katakana, so the issue as filed is fixed. @torytyler's follow-up note about kanji specifically is **not** covered here. Adding an analyzer later means one branch in `romanizeCjk()`; the pipeline position and the speech-only rule do not change.

## Cost

No new dependencies. No settings, no admin UI change. `romanizeCjk()` bails after one regex test on text without kana, so a Latin-only station is byte-identical.

## Testing

`controller/scripts/romanize.test.ts` — 16 cases: the reported example end to end, digraphs (`きゃ` → `kya`, not `kiya`), the prolonged mark, sokuon incl. Hepburn's `tch`, moraic n before labials, plus negative cases pinning that Latin, punctuation, digits, kanji, Cyrillic, Hangul and Arabic all survive byte-for-byte, idempotency, and the display-vs-speech split.

- `npm test` — all 57 files pass
- `npm run lint` — 0 errors; neither new nor edited file appears in the output

https://claude.ai/code/session_01M4514Jut2csAspq5HbSgWz

### DIFF
--- a/controller/scripts/romanize.test.ts
+++ b/controller/scripts/romanize.test.ts
@@ -1,0 +1,152 @@
+// Unit tests for audio/romanize.ts — the kana → Latin layer that stops espeak
+// reading Japanese metadata as "japanese letter japanese letter" (issue #1179).
+//
+// The cases that matter most are the NEGATIVE ones: this runs over every spoken
+// line on every station, so anything that isn't kana must survive byte-for-byte.
+//
+// Run: npm test -- romanize
+
+import assert from 'node:assert/strict';
+import { romanizeCjk } from '../src/audio/romanize.js';
+import { normalizeForDisplay, normalizeForSpeech } from '../src/audio/speech-text.js';
+
+let failures = 0;
+function test(name: string, fn: () => void | Promise<void>) {
+  return Promise.resolve()
+    .then(fn)
+    .then(() => console.log(`  ✓ ${name}`))
+    .catch((err) => { failures++; console.error(`  ✗ ${name}\n      ${err?.message || err}`); });
+}
+
+async function main() {
+  console.log('the reported case:');
+
+  await test('the issue\'s own example romanizes end to end', () => {
+    // Both halves of "ウルフルズ - バカサバイバー" are pure katakana, which is
+    // why this case is fixed without any kanji dictionary.
+    assert.equal(romanizeCjk('ウルフルズ'), 'Urufuruzu');
+    assert.equal(romanizeCjk('バカサバイバー'), 'Bakasabaibaa');
+    assert.equal(romanizeCjk('ウルフルズ - バカサバイバー'), 'Urufuruzu - Bakasabaibaa');
+  });
+
+  await test('a real DJ line keeps its English intact', () => {
+    assert.equal(
+      romanizeCjk('Up next, ウルフルズ with バカサバイバー here on SUB/WAVE'),
+      'Up next, Urufuruzu with Bakasabaibaa here on SUB/WAVE',
+    );
+  });
+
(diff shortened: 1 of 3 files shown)


perminder-klair:
Closing unmerged. The diagnosis here holds (root cause and phoneme evidence written up in #1179), but the fix is the wrong shape: it ships a hand-maintained kana table in-repo and only ever produces literal transliterations, so ウルフルズ reads as "Urufuruzu" rather than the band's actual name, and kanji stays unfixed without a ~41 MB dictionary.

Going with the prompt-level approach instead: have the DJ write the Latin form itself. No table to maintain, no dependency, real romanizations, and it covers kanji and Chinese too.

Branch left in place for the espeak evidence.

PR #1244, opened 2026-07-29
fix(analyzer): release CLAP + Demucs after idle on CPU hosts (#1204)

### LINKED ISSUE #1204: Analyze_worker RSS/swap grows unbounded over long uptime (glibc allocator not releasing   freed audio-analysis buffers)
# Analyze Worker Memory Growth

## Summary

`controller/scripts/analyze_worker.py` is a long-lived worker process. Per its own docstring, it is spawned once by `controller/src/music/analyzer.ts` and kept alive for the lifetime of the service, so the multi-second `librosa` import is paid once rather than per track.

The process accumulates memory monotonically with the number of requests served and never returns it. This happens with `ANALYZE_AUDIO_EMBEDDING` and `ANALYZE_VOCAL_ACTIVITY` both disabled, meaning no Torch, Transformers, CLAP, or Demucs objects are ever constructed. The growth therefore comes from the default, lightest configuration of the worker, not from the optional heavy features.

On a low-traffic station, where analysis is triggered per newly ingested track, the process reached a peak RSS of approximately **2.4 GB** after about **four days** of uptime. Once the workload subsided, approximately **1.5 GB** of that memory remained parked in swap while current RSS fell to about **210 MB**.

---

## Environment

**Host**

- 6 GB RAM
- 3 GB Swap
- `vm.swappiness=10`
- Docker

Analyzer environment from `/proc/<pid>/environ`:

```text
ANALYZE_SECONDS=40
ANALYZE_AUDIO_EMBEDDING=
ANALYZE_VOCAL_ACTIVITY=
CLAP_MODEL=
CLAP_MODEL_PATH=
DEMUCS_MODEL=
```

Both optional analysis features are disabled.

---

## Evidence

`/proc/<pid>/status` after approximately four days:

```text
VmHWM:    2412076 kB
VmRSS:     216584 kB
VmSwap:   1516228 kB
```

The large gap between `VmHWM` and `VmRSS` indicates that the process previously allocated approximately **2.4 GB** of anonymous memory, later freed it at the Python level, but the allocator retained the address space instead of returning it to the operating system.

Eventually, the kernel reclaimed the cold pages into swap, where they remain because they are never touched again.

### Comparison

| Worker | Loads Model Weights | Peak RSS | Swap |
|--------|---------------------|----------|------|
| `pocket_tts_worker.py` | Yes | ~934 MB | ~747 MB |
| `kokoro_worker.py` | Yes | ~682 MB | ~655 MB |
| `analyze_worker.py` | No | ~2.36 GB | ~1.48 GB |

The worker that loads **no model weights** has the largest memory footprint by roughly **2.5×**. This strongly suggests the growth comes from accumulated request processing rather than resident model memory.

Host memory remained largely free throughout:

```text
Mem:  6.0Gi total, 622Mi used, 5.2Gi free
Swap: 3.0Gi total, 3.0Gi used
vmstat si/so: single-digit kB/s
```

This indicates retained memory rather than active paging or memory pressure.

---

## Root Cause

No Python-level memory leak was found.

The active code path retains no growing Python objects across requests.

- Results are serialized to stdout and discarded.
- Temporary audio files are deleted in a `finally` block.
- CLAP and Demucs singleton objects are never initialized because the corresponding environment variables are disabled.

Every `/analyze` request still executes the core audio analysis pipeline:

- `librosa.load`
- `librosa.beat.beat_track`
- `librosa.feature.chroma_cqt`
- `librosa.feature.mfcc`
- `librosa.segment.agglomerative`
- `librosa.onset.onset_strength`
- `librosa.feature.rms`

These operations allocate and free large temporary NumPy and SciPy buffers.

The observed behaviour is consistent with **glibc malloc arena fragmentation**.

Over time:

1. Large temporary allocations are created.
2. Python releases them.
3. glibc retains the underlying heap arenas instead of returning them to the operating system.
4. The heap high-water mark continually increases.
5. Since the worker process is intentionally long-lived, the memory is never fully reclaimed until the process exits.

---

### Option 1 (Preferred)

Use an allocator with better memory reclamation.

Preload one of:

- `jemalloc`
- `tcmalloc`

using `LD_PRELOAD` in the analyzer container.

These allocators are widely used for long-running NumPy and SciPy workloads because they return unused pages to the operating system much more aggressively than glibc.

### Option 2

Recycle the analyzer worker periodically.

The parent process in

```text
controller/src/music/analyzer.ts
```

already owns spawning and keep-alive logic.

Restart the worker after:

- a configurable request count,
- a maximum uptime, or
- a configurable idle timeout.

This bounds memory usage regardless of allocator behaviour while preserving the existing architecture.

---

## Recommendation

I recommend implementing **Option 1** as the primary fix because it directly addresses allocator fragmentation without changing worker lifecycle behaviour.

Option 2 can be added as an operational safeguard to guarantee bounded memory usage during extremely long runtimes.

Happy to open a PR implementing the allocator-based solution if that is the preferred direction.

geekylakshya:
## What

Arms the existing idle model unloader on CPU devices, adds `malloc_trim(0)` to `_release_models()`, and splits the idle clock so it tracks heavy model usage rather than every request.

## Why

Fixes #1204.

`_maybe_start_idle_release()` only armed when `resolve_device() == "cuda"`, so on CPU-only hosts nothing ever released the CLAP/Demucs singletons after a backfill or stem cache pass. One field host eventually reached:

- `VmHWM: 2412076 kB`
- `VmRSS: 216584 kB`
- `VmSwap: 1516228 kB`

## How Tested

Testing was performed on an isolated worker running inside a purpose-built LXC matching the field host:

- 6291456 kB RAM
- 3145728 kB swap
- `vm.swappiness=10`
- image digest `sha256:adca29e2...`
- device: `cpu`
- `ANALYZE_IDLE_UNLOAD_S=120`
- `ANALYZE_AUDIO_EMBEDDING` and `ANALYZE_VOCAL_ACTIVITY` unset so models were only loaded through the per-request `force=True` path

The workload consisted of:

- a five-track backfill with `embed`, `vocal`, and `complete`
- lightweight BPM/key requests every 15 seconds throughout the idle window

Peak `VmHWM` reached **2,164,900–2,172,044 kB** across the three runs, compared with **2,412,076 kB** on the field host, indicating the synthetic workload reproduces the real memory profile rather than a substantially smaller one.

Three variants were run on the same host using the same workload.

| | A: shipped | B: gate removal + trim | C: this PR |
|---|---:|---:|---:|
| Before load | 40,540 kB | 41,680 kB | 41,660 kB |
| After backfill | 1,820,900 kB | 1,885,276 kB | 1,752,976 kB |
| After idle window | 1,734,172 kB | 1,652,520 kB | 1,070,352 kB |
| Reclaimed | 86,728 kB | 232,756 kB | 682,624 kB |
| **Release fired** | **NO** | **NO** | **YES** |

The result that matters is the final row rather than the reclaimed RSS. Variants A and B never logged a release, while Variant C logged:

```text
[analyze-worker] analysis device: cpu
[analyze-worker] idle model release armed — models unload after 120s without heavy use
[analyze-worker] idle 120s — released CLAP + Demucs from system memory (reloads on next use)
```

The reclaimed RSS values come from single runs and show visible variance. Variants A and B peaked 68 MB and 132 MB above Variant C on otherwise identical workloads, so those numbers should be treated as indicative rather than precise. The release/no-release behaviour reproduced consistently and is the primary result.

### Why the idle clock split is necessary

Variant B contains only the two obvious changes:

```diff
+ ctypes.CDLL(None).malloc_trim(0)          # in _release_models

- if ... or resolve_device() != "cuda":
+ if _idle_thread_started or IDLE_UNLOAD_S <= 0:
```

Despite those changes, Variant B still never releases the models.

The reason is that the existing `_touch()` updates the idle timer on **every** request. Lightweight BPM/key analysis never loads CLAP or Demucs, but it continually refreshes the timer, preventing the countdown from ever reaching the unload threshold. On any installation with regular analysis traffic, those two changes alone are effectively a no-op.

Variant C resolves this by tracking heavy model usage separately. The idle release occurred between **t = 502.9 s** and **t = 517.9 s** while lightweight requests continued arriving every 15 seconds.

Additional verification:

- `idle_release_test.py`: 16 test cases, all passing in the image's production virtual environment
- `vocal_gate_test.py`: unchanged and still passing
- `npm test`: the same 12 failing TypeScript tests reproduced on a clean `develop` checkout with this branch stashed; this PR does not modify any TypeScript

## Notes for Reviewers

This testing intentionally does **not** establish the following:

- **`VmSwap` remained 0 in all three runs.** The test host had sufficient free RAM, so the kernel never reclaimed pages. This fix prevents model weights from remaining resident long enough to become swap candidates, but it does not reproduce the original swap event.
- **Residual RSS after release remained 1,028,692 kB above baseline**, higher than the few hundred MB discussed in the issue. A CLAP-only run settled at 813,968 kB, suggesting most of the remaining memory is likely librosa, numba, or Torch scratch space rather than model weights. That is probably worth investigating separately.
- **The contribution of `malloc_trim()` is not isolated.** Variant B never reached the unload path, so `malloc_trim()` was never executed there. Its contribution is therefore included within Variant C's total reclaimed memory.
- **The 1800 second CPU default is a judgement call rather than a measured optimum.** All measurements used a 120 second timeout. The implementation is identical apart from the threshold, and I am happy to adjust the default if reviewers prefer a different value.

One remaining correctness gap is worth calling out explicitly: although unit tests verify the singleton and failure-flag behaviour after unloading, I have not yet exercised a real heavy analysis request after an idle release to confirm that models reload correctly from the on-disk cache. I can add that validation before merge if reviewers would find it valuable.

## Out of Scope

While tracing model load paths, I found another force-load path that is unrelated to this issue.

`{"texts": [...]}` requests always call:

```python
get_embedder(force=True)
```

This path is reached by:

- `/library/search-sound`
- the DJ picker's `searchBySound` tool

As a result, CLAP can become resident even without an admin backfill or stem cache pass. That behaviour is unchanged by this PR.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -157,9 +157,13 @@ SITE_URL=
 # the container. See docs/unraid.md.)
 # ANALYZE_DEVICE=    # auto (default) / cpu / cuda — torch device for CLAP/Demucs;
 #                    # only meaningful on the cuda analyzer flavour
-# ANALYZE_IDLE_UNLOAD_S=  # cuda flavour: seconds of no analysis before models are
-#                         # dropped from VRAM (default 300; 0 = keep resident).
-#                         # Frees the GPU for co-resident TTS/LLM between passes.
+# ANALYZE_IDLE_UNLOAD_S=  # seconds of no CLAP/Demucs use before the models are
+#                         # dropped (0 = keep resident). Defaults per device:
+#                         # 300 on cuda (frees the GPU for co-resident TTS/LLM
+#                         # between passes), 1800 on cpu (a backfill or sound
+#                         # search loads them once and they'd otherwise sit in
+#                         # RAM/swap forever; the longer window keeps the cold
+#                         # reload off interactive sound searches).
 #
 # Runtime flags — env wins ON over the admin toggles (settings.audio.*), never
 # off. A flag with no matching backend in the image is a clean no-op.
--- a/cli/src/assets.generated.ts
+++ b/cli/src/assets.generated.ts
@@ -719,8 +719,11 @@ services:
       # meaningful on the cuda analyzer flavour (docker-compose.analyzer-gpu.yml);
       # cpu-wheel images always resolve to cpu.
       - ANALYZE_DEVICE=\${ANALYZE_DEVICE:-}
-      # Seconds of no requests before the cuda flavour drops its models out of
-      # VRAM (default 300; 0 = never). CPU images ignore it.
+      # Seconds of no CLAP/Demucs use before the worker drops the models
+      # (0 = never). Applies on BOTH devices: default 300 on cuda (frees VRAM),
+      # 1800 on cpu — one backfill or sound search loads them into the
+      # long-lived worker and nothing else ever releases them, so on a small
+      # host they end up parked in swap (#1204).
       - ANALYZE_IDLE_UNLOAD_S=\${ANALYZE_IDLE_UNLOAD_S:-}
       - CLAP_MODEL=\${CLAP_MODEL:-}
       - CLAP_MODEL_PATH=\${CLAP_MODEL_PATH:-}
@@ -1002,8 +1005,11 @@ services:
       # meaningful on the cuda analyzer flavour (docker-compose.analyzer-gpu.yml);
       # cpu-wheel images always resolve to cpu.
       - ANALYZE_DEVICE=\${ANALYZE_DEVICE:-}
-      # Seconds of no requests before the cuda flavour drops its models out of
-      # VRAM (default 300; 0 = never). CPU images ignore it.
+      # Seconds of no CLAP/Demucs use before the worker drops the models
+      # (0 = never). Applies on BOTH devices: default 300 on cuda (frees VRAM),
+      # 1800 on cpu — one backfill or sound search loads them into the
+      # long-lived worker and nothing else ever releases them, so on a small
+      # host they end up parked in swap (#1204).
       - ANALYZE_IDLE_UNLOAD_S=\${ANALYZE_IDLE_UNLOAD_S:-}
       - CLAP_MODEL=\${CLAP_MODEL:-}
       - CLAP_MODEL_PATH=\${CLAP_MODEL_PATH:-}
--- a/cli/src/assets.generated.ts
+++ b/cli/src/assets.generated.ts
@@ -367,8 +367,11 @@ services:
       # meaningful on the cuda analyzer flavour (docker-compose.analyzer-gpu.yml);
       # cpu-wheel images always resolve to cpu.
       - ANALYZE_DEVICE=\${ANALYZE_DEVICE:-}
-      # Seconds of no requests before the cuda flavour drops its models out of
-      # VRAM (default 300; 0 = never). CPU images ignore it.
+      # Seconds of no CLAP/Demucs use before the worker drops the models
+      # (0 = never). Applies on BOTH devices: default 300 on cuda (frees VRAM),
+      # 1800 on cpu — one backfill or sound search loads them into the
+      # long-lived worker and nothing else ever releases them, so on a small
+      # host they end up parked in swap (#1204).
       - ANALYZE_IDLE_UNLOAD_S=\${ANALYZE_IDLE_UNLOAD_S:-}
       - CLAP_MODEL=\${CLAP_MODEL:-}
       - CLAP_MODEL_PATH=\${CLAP_MODEL_PATH:-}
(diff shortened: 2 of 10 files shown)


perminder-klair:
Ran the reload-after-release validation flagged in the PR notes ("I have not yet exercised a real heavy analysis request after an idle release"). **It passes.**

Setup: `subwave-analyzer-heavy:latest` image, this branch's `analyze_worker.py` mounted over the baked one, `ANALYZE_DEVICE=cpu`, `ANALYZE_IDLE_UNLOAD_S=45`, env embed/vocal flags unset (models load only via the per-request `force=True` path). Driven over stdio; persistent HF/torch-hub cache mounted so run 2 measures a true reload from disk.

| step | result | time |
|---|---|---:|
| `{texts}` first-ever (download + load) | ok, 2×512-d | 26.8s |
| `{texts}` warm | ok | 0.1s |
| analyze `vocal+embed` first-ever (Demucs download + separation) | ok, `audio_embedding` + `vocal_ranges` present | 41.8s |
| **idle release** | `idle 45s — released CLAP + Demucs from system memory` | fired |
| `{texts}` after release (**cold reload from disk cache**) | ok, 2×512-d | **2.2s** |
| analyze `vocal+embed` after release (**Demucs reload + separation**) | ok, same fields | **14.5s** |

Both models reload cleanly from the on-disk cache after a release, failure flags stay clear, and the worker never restarted. The cold CLAP reload (~2s on a desktop CPU) sits well inside the 20s interactive sound-search deadline; slower hosts have more headroom than feared.

perminder-klair:
thanks @geekylakshya !

PR #1255, opened 2026-07-31
feat(library): carry all derived data across Navidrome's canonical-ID migration

### LINKED ISSUE #1253: URGENT: Navidrome about to break 3rd Party Apps
https://github.com/navidrome/navidrome/pull/5824

From reading the subreddits of Navidrome, AudiomuseAI and Symfonium, it appears Navidrome are planning on releasing this update at the weekend, which changes the Navidome ID of every track.
Any Subwave user who allows their Navidrome server to auto-pull the new version is going to find no music in Subwave until Subwave is adjusted to the new changes in Navidrome.

Every Navidrome user is advised to take a backup of navidrome.db before updating, and it would be preferable to keep Navidrome on your current installed version until 3rd Party Apps like Subwave have had a chance to catch up.

@perminder-klair This could be worthy of a Announcement to all users on Discord.

Edit: Source [link](https://www.reddit.com/r/navidrome/comments/1vbq2yo/headsup_the_next_release_changes_all_internal_ids/)

perminder-klair:
Closes #1253.

Navidrome 0.64.0, released September 12, 2026, re-encodes internal IDs. Previously, the first SUB/WAVE reconciliation after upgrading could prune the old library rows and discard derived tags, enrichment, analysis and vectors. This change confirms the canonical successor against the live library and adopts the existing data before pruning genuinely deleted tracks.

The migration also preserves play attribution, stem caches, blocklist entries and playlist rules, listener likes, recipe seeds and playlist IDs, and show playlist pins. The ID transform matches Navidrome 0.64.0's released migration and its golden vectors.

Recovery is durable across an interrupted controller or tagger run:

- Adoption and its confirmed ID map commit in the same SQLite transaction (`id_rotation_journal`, schema 26).
- The controller applies pending mappings at the in-run notification, every child exit, and boot, including before the library/vector handle opens.
- State stores persist on replay even when an earlier failed write already changed their caches. The journal is acknowledged only after successful writes, and only for the processed snapshot.
- Playlist IDs are checked against the live playlist index. An unavailable index retains recovery state and holds post-tag playlist sync. Legacy JSON handoffs remain readable.

The carried track-column set is derived from the schema and audited by a drift test. Fresh successor data is preserved under the existing field-specific merge rules. Already-pruned data cannot be recovered without a backup or regeneration. The operator news post now describes the released Navidrome version and this limitation.

Validation:

- All four original ID-migration test files pass against the refreshed `develop` integration.
- Seven new recovery/journal tests pass: process exit after adoption commits, rollback on journal-write failure, preservation of later adoptions, and failed-write retries for blocklists, likes, recipes and schedule. The five crash/store recovery regressions fail on the unfixed implementation and pass with the repair.
- Controller, web and MCP lint/typechecks pass; generated schema and theme mirrors have no drift.
- Full controller suite: **1,690 tests passed, 0 failed, 0 skipped**, including the Python analyzer tests.
- Independent standards and specification re-reviews found no remaining blockers.

### DIFF
--- a/controller/scripts/id-adoption-columns.test.ts
+++ b/controller/scripts/id-adoption-columns.test.ts
@@ -1,0 +1,121 @@
+// The schema-mirror guard for Navidrome ID adoption
+// (music/library-db/id-adoption.ts).
+//
+// Adoption moves a rotated track's derived data onto its new id. Which columns
+// move used to be a hand-written list, and that list went stale the way every
+// hand-written mirror of a growing schema does: seven columns landed after it
+// was written (lead/tail silence, tail_start_ms, the three analyze-failure
+// counters, text_vector_dirty) and every one of them would have been dropped
+// on every adopted row — with `analysis_version` carried on top, so nothing
+// would ever have re-derived them. The library would have come out of the
+// migration quietly missing days of analysis, and no test would have noticed.
+//
+// So the carried set is now DERIVED (all physical columns minus the walk-owned
+// denylist) and this file is the reminder that survives the next migration:
+// every column must be CLASSIFIED — walk-owned, grouped, special, or carried
+// by the COALESCE default. The runtime default is safe either way, which is
+// the point: this test fails the suite, not the operator's library.
+//
+// Runs a REAL better-sqlite3 DB against a temp STATE_DIR so the classification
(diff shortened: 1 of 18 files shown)


joshp23:
Thank you for merging this. I may be missing something, is there a migration process to follow after updating navidrome?

PR #1259, opened 2026-08-01
fix(broadcast): stop a single failed Icecast poll from releasing the idle pause (#1256)

### LINKED ISSUE #1256: Idle pause releases ~28s after engaging: stream-idle fails open on a single failed Icecast poll
**Version:** controller 1.2.0
**Settings:** `stream.idleWhenEmpty: true`, `stream.idleAfterMinutes: 10`, `llm.pauseWhenEmpty: true`

## What happens

With `idleWhenEmpty` on and an empty room overnight, the programme does not stay paused. It pauses, releases about 28 seconds later, runs for the full 10 minute idle window, pauses again, and repeats. Thirteen cycles in one night:

```
01:38:09  [scheduler] programme idle-paused - no listeners for 10 min
01:38:37  [scheduler] programme resumed - idle pause released     (+28s)
01:48:39  [scheduler] programme idle-paused - no listeners for 10 min
01:49:07  [scheduler] programme resumed - idle pause released     (+28s)
...
```

Listener count was a solid 0 for the whole period, sampled once a minute, with no gaps. Net effect: the power save is inactive roughly 95% of the time it should be active.

## Why

The resume is the `count === null` branch in `nextIdleState()`, not the "someone connected" branch. The two are distinguishable by the log string: `programme resumed - idle pause released` is emitted when the count is null or 0, while `programme resumed - listener connected` is the count > 0 case.

`src/broadcast/stream-idle-pure.ts`:

```ts
if (prev.idle) {
  // Fail-open: an unknown count can't confirm the room is still empty
  if (count === null || count > 0) {
    return { state: { idle: false, zeroSince: null }, action: 'resume' };
  }
  return { state: prev, action: 'reassert' };
}
```

A *single* failed Icecast poll is enough to release the pause. There is no consecutive-failure tolerance, even though `listeners.ts` right next door already establishes that doctrine for the cached status:

```ts
// 4 polls x 15s ~ 1 min of sustained failure.
const STALE_STATUS_LIMIT = 4;
```

Two things make a single blip very likely to be hit:

1. `stream-idle.ts` deliberately forces a fresh poll every 5s while paused, rather than reading the 15s monitor cache. That is about 120 polls per 10 minute pause, so even a sub-1% poll failure rate is near certain to fire inside one pause window.
2. The fetch deadline is tight (`timeoutMs: 1500` in `fetchCount`).

This also explains why a short manual probe at the 15s monitor cadence looks perfectly healthy while the pause still never holds. The failing loop is the 5s one.

## Why it matters beyond the power save

`djCallsAllowed()` short-circuits on `streamIdle`, and its own next line has the same single-blip fail-open:

```ts
export function djCallsAllowed() {
  if (streamIdle) return false;
  if (!settings.get()?.llm?.pauseWhenEmpty) return true;
  if (lastCount === null) return true;   // same one-poll fail-open
  return lastCount > 0;
}
```

So one failed poll drops the idle gate and the LLM token gate in the same tick. On this install the DJ has spent millions of tokens in a day to a room the listener log shows as empty, which is the same complaint as #1023 (closed by the idle-pause feature in #1066).

I want to be careful not to overclaim: I have proven this mechanism releases the pause, but I have not yet proven it accounts for the full 24/7 spend on the worst days, since `lastCount` returns to 0 once the poll recovers. There may be a second cause. Flagging it in case it is obvious to you.

## Suggested fix

Give the gates the same hysteresis the status cache already has: keep serving the last reading that was actually read from Icecast, and only report "unknown" (and so fail open) once failures reach `STALE_STATUS_LIMIT`. A genuine outage still resumes, which preserves the documented intent, but one missed poll no longer does.

I am running this locally as a small patch: a single `gateListenerCount()` in `listeners.ts` that both gates read instead of the raw `lastCount`. Verified by driving the real module with a deliberately unreachable status URL:

```
healthy poll           raw=1     gate=1
blip #1 (below limit)  raw=null  gate=1     <- upstream resumes here
blip #2 (below limit)  raw=null  gate=1
blip #3 (below limit)  raw=null  gate=1
sustained outage (#4)  raw=null  gate=null  <- fails open, as designed
recovery               raw=1     gate=1
```

Happy to open a PR if that shape looks right to you.

## Smaller related note

At controller start the DJ can fire a pick before the first listener poll has returned, so `lastCount` is still null from "never polled" rather than from a failure, and the gate fails open. Minor (one pick per restart), but "not polled yet" and "poll failed" arguably deserve different treatment.

## Environment

- Docker Compose, Icecast in the merged `broadcast` container
- Icecast reachable and healthy: 60 consecutive successful status polls at the time of writing
- Behind a reverse proxy, `x-forwarded-for` not configured, so all connections carry the proxy address

perminder-klair:
Fixes #1256.

## The bug

`listeners.djCallsAllowed()` and the idle-pause monitor both fail **open** on an unknown listener count, and both read the **raw** poll result — where one 1.5s `fetchWithTimeout` abort on `/status-json.xsl` is indistinguishable from Icecast being gone.

`stream-idle.ts` deliberately forces its own poll every 5s while paused (~120 polls per 10-minute pause, *not* the 15s monitor cadence a short manual probe would sample), so a sub-1% failure rate fired inside nearly every pause window. The reporter measured: paused → `programme resumed — idle pause released` ~28s later, 13 cycles in one night, power save inactive ~95% of the time it should have been active. Every release also reopened the LLM token gate in the same tick, since `djCallsAllowed()` short-circuits on `streamIdle` and then hits the same one-poll fail-open on its next line.

**And the blips are mostly self-inflicted.** The reporter's follow-up (https://github.com/perminder-klair/subwave/issues/1256) pinned the mechanism: while paused, the idle monitor's 5s forced poll and the always-on 15s monitor poll line up every 15 seconds (15 is a multiple of 5), and the *overlapping* requests are what blow the 1.5s deadline. Measured on their station: **0** poll failures in 90 minutes with the room occupied, **196** in 49 minutes with it empty — clustered at :15–:17 and :45–:47 in pairs ~246ms apart, while direct curls to the same Icecast never exceeded 1.5s. The failures occur only in the state where they do damage.

Reproduced at the pure level — replaying a 120-tick pause with a single `null` at tick 37 against the pre-fix input:

```
resume fired at tick: 37 → 185 seconds into the pause
```

One extra facet not in the report: the same blip while **live** hit the "occupied (or unknown) — reset the empty clock" branch, so a blip inside the 10-minute window restarted it. The pause was being *delayed* by blips as well as released by them.

## The fix, layer 1: hysteresis

Give the count the hysteresis the cached `StreamStatus` has carried since #461. New pure `gatedCount(raw, lastGood, consecutiveFailures, limit)` in `listeners.ts`, right next to `statusAfterFailure` and pinned in the same test file:

- poll succeeded → that reading, always (a real `0` is an observation, not a hold)
- transient failure (`< STALE_STATUS_LIMIT`) → the last count actually read from Icecast
- sustained failure (`≥ limit`) → `null`, and the documented fail-open kicks in
- never polled → `null`

Both fail-open gates now read `gatedListenerCount()` instead of `lastCount`. A failed fetch is not an observation — it says nothing about the room, so the freshest real observation is strictly better information than "unknown".

**Fail-open is untouched.** A genuinely unreachable Icecast still reads null, still resumes the programme, still keeps the DJ on air. Only the definition of "unreachable" tightened, from one blip to ~20s while idle (the 5s forced polls reach the limit faster — the right way round: an unobservable stream should resume sooner) and ~1 min while live.

## The fix, layer 2: single-flight the poll

The reporter's first follow-up recommendation — "don't let the two pollers collide" — implemented at the chokepoint rather than by de-phasing timers. `fetchCount` is now a `singleFlight()`-guarded `pollCount`: a caller that fires while a poll is already in flight **joins it** instead of opening a competing request. When the 5s and 15s timers line up, Icecast sees one request, so the collision that manufactures the `AbortError`s can't happen at all. Hysteresis (layer 1) remains the safety net for any genuinely random blip.

The same guard closes a state race the collision made likely: the failure path commits ~1.5s after the success path (it waits out the fetch timeout), so a slow doomed poll could land *after* a younger successful one and overwrite its fresh state — a phantom `null` on the raw readers and a "consecutive" failure count that isn't. Joining the in-flight poll keeps every caller's freshness contract (`refresh()` included — the answer is at most one timeout old, strictly fresher than the 15s cache it exists to bypass). `probeListenerCount()` bypasses the guard deliberately: it runs in the analyze child process with no concurrent pollers, and joining would drop its skip-history flag.

## Deliberate non-changes

Documented in the code so they don't get "unified" later:

- **`presentListeners()` stays on the raw count.** It fails *closed*, where a blip costs at most one skipped scrobble — the safe direction, no hysteresis needed.
- **`getListenerCount()` stays raw**, and the reporting surfaces (admin table, DJ context) keep using it.
- **`POST /request`** rides `djCallsAllowed()` and so is affected. Under a real outage its forced poll's failures reach the limit and requests are accepted exactly as before; only a single blip — where the held reading is seconds old and accurate — now decides differently.

Also picks up the report's smaller note: `startListenerMonitor()` now returns its first poll and `server.ts` awaits it before `queue.startWatcher()`, so "never polled" can't fail open into one free agent pick on every controller restart. Bounded at 2s (a wedged admin endpoint costs 2s/mount on top of the 1.5s status fetch), and the HTTP server is already listening by that point, so it can't stall a health check. Only helps when Icecast is already up — on a cold `compose up` the broadcast container isn't serving yet and the gate fails open as designed.

## Tests

`controller/npm test` — all 75 files pass (post-develop-merge count). `npm run lint` clean (0 errors).

- `scripts/listeners-status.test.ts` — 7 new `gatedCount` cases (blip holds, blip-while-occupied, just-below-limit, at/past limit → unknown, never-polled, recovery) + 3 `singleFlight` cases (concurrent callers join one run; a settled run re-arms; a rejected run re-arms, so one timeout can't wedge polling).
- `scripts/stream-idle.test.ts` — 4 new cases that replay a full pause window through `gatedCount` → `nextIdleState` the way the monitor drives it: one blip at tick 37 holds, six scattered blips hold, a 10-tick outage resumes on the 4th failure, and a real listener still wakes the stream on the next tick. Plus 5 `replayLive()` cases for the **live** half — the facet with no production measurement behind it, where a blip used to reset the empty clock and *delay* the pause: a clean empty room pauses exactly on the window, a mid-window blip no longer defers it, scattered blips still land on time, a sustained outage still defers it (fail-open intact), and a real listener still restarts the clock.
- `scripts/analyze-quiet.test.ts` — 4 new cases pairing `gatedCount` with `quietGateDecision` the way `stream-idle.test.ts` pairs it with `nextIdleState`: a blip while occupied no longer starts the pass, a blip in an empty room doesn't restart the quiet window, a sustained outage still fails open, and the child's first probe still reads unknown.

`nextIdleState` itself is unchanged — its fail-open branching was always right; what was wrong was the input.

**Live end-to-end run** (isolated controller from this branch + a fake Icecast with on-command hung polls + a fake Liquidsoap telnet serving the `idle_on`/`idle_off`/`idle_status` gate; `idleAfterMinutes: 1`):

| phase | result |
|---|---|
| pause engages after 1 empty minute | ✅ |
| 1 hung poll (the #1256 trigger) | ✅ held |
| 2 scattered blips 12s apart | ✅ held |
| 6 consecutive hung polls | ✅ released 22s in (fail-open, ~4 × 5s cadence) |
| polls recover, room still empty | ✅ re-paused |
| listener connects on the mount | ✅ resumed in 4s |
| overlap audit across the run | ✅ 30 polls, 9 hung past the deadline, **0 overlapping requests** |

Both resume paths fired for their intended reasons (`resumed — idle pause released` vs `resumed — listener connected` in the controller log).

Credit to @mauriciopaim, who diagnosed this down to the `AbortError`, instrumented the consecutive-failure depth, ran the patch shape in production for a day, flagged that probing at 15s is the wrong loop — and then followed up with the collision mechanism that layer 2 implements and the field run that produced layer 3.

## The fix, layer 3: the third fail-open gate, and the deadline

Added after @mauriciopaim's 11.53 h field run (comment below), which confirmed the gate fix and the `server.ts` await but showed `singleFlight()` halving the poll failures rather than ending them — leaving a residue that is exactly 30 s periodic, exclusively an empty-room phenomenon, and bursty in a way nothing on their box explained.

**`QUIET_POLL_MS = 30_000` in `music/analyze.ts` is the only 30 s cadence in the controller.** It is the analysis quiet gate's wait loop, and it fires `probeListenerCount()` — a full `/status-json.xsl` fetch on the same deadline — from the **tagger child process**, which `singleFlight()` structurally cannot coalesce with anything. And the quiet gate only lets analysis run when the room is empty, which is why the failures are an empty-room phenomenon and why they track the analysis backlog rather than any host cron. The heavy DSP in that window is a plausible *cause* of the timeouts, not a co-traveller.

Two changes fall out.

**The quiet gate is the THIRD fail-open-on-unknown reader**, and it was still on the raw count. Its fail-open runs the *opposite* way from the other two — unknown means PROCEED — so a single timed-out poll started a Demucs pass over a listener's head instead of releasing a pause. Same class as the bug this PR opened with, in the one consumer it hadn't touched. `probeListenerCount()` now returns `gatedListenerCount()` off the child's own module state: the child polls once per track (or every 30 s while waiting out the window), so consecutive failures accumulate across calls exactly as they do in the monitor loop, and the first call still reads unknown because `lastGoodCount` is null. Boot behaviour is byte-identical.

**`STATUS_TIMEOUT_MS` goes 1500 → 3000.** 1500 was chosen when three cadences could each have a request in flight, where a longer deadline meant a deeper pile-up; layer 2 removed that objection. It was also measurably tight — hundreds of `AbortError`s per empty-room hour against an Icecast that direct curls never took near a second to answer. This attacks the failure *rate* rather than absorbing more of it, which is the reason not to widen `STALE_STATUS_LIMIT` instead: that constant is shared with #461's offline detection, which should not get slower. Still comfortably inside the idle monitor's 5 s tick, so the ~20 s fail-open budget while paused is unchanged. `FIRST_POLL_WAIT_MS` is now *derived* from the deadline rather than written flat — otherwise raising one silently gives up before every slow poll and hands back the "never polled" fail-open the boot await exists to close.

## On the margin, and a measurement request

@mauriciopaim flagged one excursion to `#3 in a row`, one poll short of the limit, and rightly declined to claim "never gets past #2" from the data. Worth knowing before anyone acts on that number: **the depth histogram may be conflating two processes.** If the instrumentation sits in `pollCount`'s catch, the tagger child logs there too, and the child's `consecutiveStatusFailures` is a separate counter that no idle/DJ gate reads. The `#3` excursion may never have been on the gate path at all.

If you keep it running, tagging that log line with the process (pid, or the tagger's `MANAGED_ENV`) would separate controller-gate failures from child probe failures — and the deadline change above should move the rate on both.

## Note on the token spend — answered

The original version of this section speculated about a second cause. @mauriciopaim's bucketing closed it: on their worst day, 656 LLM calls of which 645 ran to an empty room, 2.55 M tokens, 64 % `djAgentPick` at ~4.9 k input tokens each. The picker is input-heavy enough that ~4 fail-open ticks a minute for one night reaches millions on this gate's arithmetic alone. Confirmed from the other side after this branch went on: 9 LLM calls on the first full day, every one with a listener present. There is no second gate to find.


mauriciopaim:
Ran this branch (head `cd62605f`) on the station from the report for 11.53 h, mounted over v1.2.0 so only these four files changed. Same box, same Icecast, and deliberately the same duration as the 11.6 h baseline I posted for my own hysteresis patch, so the two are directly comparable.

Short version: the gate fix works, and the `server.ts` change works. `singleFlight()` does not do what we both expected, and I think that is worth knowing before this merges.

## The gate

| | my hysteresis (11.6 h) | this branch (11.53 h) |
|---|---|---|
| `status poll failed` | 1012 | **485** |
| consecutive-failure depth | 570 @ #1, **330 @ #2** | 483 @ #1, **1 @ #2, 1 @ #3** |
| phantom `resumed — idle pause released` | 0 | **0** |
| open-gate pick at boot | 2 per start | **0** |

Longest hold I have ever recorded on this station:

```
00:02:17  programme idle-paused — no listeners for 10 min
07:27:06  programme resumed — listener connected
```

7 h 25 m, released by a real listener on the `count > 0` branch. Both resumes in the window were real ones. No LLM calls at all during the empty stretch.

The `server.ts` await is confirmed too: zero open-gate picks across the window, against exactly 2 per controller start before it.

## singleFlight: the failures did not stop

I got a natural occupied/empty split during the window, which turned into a clean control:

| window | hours | failures | per hour |
|---|---:|---:|---:|
| empty | 2.86 | 205 | **71.7** |
| occupied | 0.90 | 1 | 1.1 |
| empty | 7.41 | 279 | **37.6** |
| occupied | 0.36 | 0 | **0.0** |

Still exclusively an empty-room phenomenon, which is the same signature as before the fix. Coalescing did halve the rate and it clearly killed the pairing: pre-fix the failures came in pairs about 246 ms apart at seconds :15-:17 and :45-:47, and there is not a single pair left in 485 failures.

What is left is a different shape. When it is active it is perfectly periodic at 30 s (264 consecutive gaps of exactly 30 s, at seconds :28-:30 and :58-:00), which is every *other* 15 s monitor poll rather than every collision. And it is bursty: some hours have ~120 failures, others have zero, including one 2 h 40 m stretch of an empty room with nothing at all.

I could not find a local cause for the bursts. I checked every scheduled job on the box that overlapped them and ruled them all out (the heaviest ran for 9 seconds). So I would read this as your second recommendation rather than your first: the 1500 ms deadline is tight enough that ordinary host pressure still reaches it, independently of the collision. That is not an argument against `singleFlight()`, which is clearly doing something, just against the expectation that it makes the failures impossible.

## One caveat on the margin

There was a single excursion to `#3 in a row`, one poll short of `STALE_STATUS_LIMIT`. My own patch never went past #2 in a comparable window. The hysteresis absorbed it correctly and nothing fell open, so this is not a problem with the change. But "never gets past #2" is not something I can claim from this data, and if the deadline is left at 1500 ms it seems worth saying out loud.

## Note on the token spend

Bucketing `events-*.jsonl` the way you suggested, on the worst day I have (656 LLM calls, of which 645 ran with zero listeners):

| kind | calls | tokens | share | per call |
|---|---:|---:|---:|---:|
| djAgentPick | 338 | 1,645,822 | 64.4% | 4,869 |
| djAgentRepick | 153 | 502,189 | 19.7% | 3,282 |
| djAgentSegment | 88 | 331,028 | 13.0% | 3,761 |
| generateStationId | 52 | 43,308 | 1.7% | 832 |
| everything else | 25 | 31,766 | 1.2% | |

2,554,113 tokens, 2,483,757 of them input. So it clusters on picks at track boundaries with the room empty, which is the pattern you named, but it is not a second gate. It is this one. The picker is input-heavy enough (~4.9k tokens of prompt per pick) that ~4 fail-open ticks a minute for a whole night reaches millions on its own arithmetic, without needing anything else to be broken.

The confirmation is the same measurement after this branch went on: on the first full day, 9 LLM calls, every one of them with a listener present, nothing to an empty room. I do not think you need to keep looking for a second cause.

## On the offer

Thanks for the credit, and no, please keep it under yours. Your version is better than mine on all three axes: `gatedCount()` sitting next to `statusAfterFailure` and reusing `STALE_STATUS_LIMIT` is the right home for it, the `server.ts` await closes a boot window I had written off as harmless, and I had missed the LIVE case entirely, where a blip resets the empty clock and delays the pause instead of releasing it. Mine only covered the half I could see from the symptom.

Happy to keep it running and report anything that changes.

perminder-klair:
## Verification run: isolated controller + fake Icecast, with a pre-fix control

Re-ran this branch end to end against a controllable fake Icecast (`/status-json.xsl` + `/admin/listclients`, on-demand hung polls, and a per-request audit recording start/end/abort so overlap and deadline behaviour are measured rather than assumed) and a fake Liquidsoap telnet serving the `idle_on`/`idle_off`/`idle_status` gate. Isolated controller on a spare port with a temp `STATE_DIR`, `idleAfterMinutes: 1`. Nothing pointed at a real station.

**This branch: 10/10 scenarios pass.** A suite that cannot fail proves nothing, so the identical suite ran against `origin/develop` as a negative control.

| scenario | `develop` (pre-fix) | this branch |
|---|---|---|
| 1 failed poll while paused | **released** — #1256 reproduced | held |
| 3 consecutive failures (limit−1) | **released** | held |
| 4× polls that are slow (2s) but **succeed** | **released** | held, 4/4 served |
| sustained 8-poll outage | fails open | fails open, 23s |
| polls recover, room still empty | re-pauses | re-pauses, 67s |
| real listener connects | wakes, 5s | wakes, 5s |
| measured abort deadline | 1501 ms | 2999–3004 ms |

The third row is the direct argument for `STATUS_TIMEOUT_MS` 1500 → 3000: on `develop`, an Icecast that merely *answers in two seconds* is enough to break the pause. Those polls are not failures in any meaningful sense, and the old deadline turned every one of them into a release.

### The quiet gate (layer 3)

`probeListenerCount()` driven directly against the fake, same sequence on both branches:

```
                                     develop      this branch
success with 3 listeners        ->   3            3
one timed-out poll              ->   null         3       <- the case under test
failures #2 and #3              ->   null, null   3, 3
failure #4 (limit)              ->   null         null    <- fail-open intact
recovery, room now empty        ->   0            0
```

On `develop` the blip reads `null`, and for this gate `null` means **proceed** — the DSP pass starts over a live listener. Same class of bug as the one this PR opened with, in the one consumer it had not touched.

### The derived boot wait

A/B on this branch with only `FIRST_POLL_WAIT_MS` changed, against a first poll that is slow (2500 ms) but succeeds:

```
derived (STATUS_TIMEOUT_MS + 500):  resolved in 2518ms, count = 4     -> got the reading
flat 2000:                          resolved in 2000ms, count = null  -> gave up early
```

That is exactly the regression the derivation prevents: raise the deadline past a hardcoded wait and boot silently abandons every slow poll, handing the watcher back the "never polled" unknown the await exists to close.

### Three things this run does NOT show

1. **It did not independently prove `singleFlight()`.** The overlap audit came back clean on both branches, but `develop` only made 12 polls to this branch's 34 — its pause released early, so it never sustained the forced 5 s cadence that produces collisions in the first place. Not evidence against the guard; @mauriciopaim's field data (pairing gone from 485 failures) is still the real measurement there.
2. It is a fake Icecast on loopback. It validates the controller's logic, not real Icecast under host pressure.
3. The quiet gate was exercised at the function level, not by running a full analysis pass.

@mauriciopaim — on the `#3 in a row` margin you flagged: worth checking whether the depth histogram is mixing two processes before reading much into it. If the instrumentation sits in `pollCount`'s catch, the tagger child logs there too, and the child's `consecutiveStatusFailures` is a separate counter no idle/DJ gate reads. Tagging that line with the pid (or the tagger's `MANAGED_ENV`) would split controller-gate failures from child probe failures. The deadline change should move the rate on both, and the 30 s periodicity you measured lines up exactly with `QUIET_POLL_MS = 30_000`, which is the child's loop.

mauriciopaim:
Checked the two-process question before reading anything else into the depth histogram, because if you were right it would retire the `#3` caveat entirely. On this station you are not, and I can show it three independent ways. Then I went looking for what the 30 s actually is, and I think I have it.

## The depth histogram is not mixing processes here

`probeListenerCount()` is never called on this box, so the tagger child never polls Icecast at all.

1. **Structural.** `audio.analyzeQuietOnly` is `false` here and `ANALYZE_QUIET_ONLY` is unset, so `readQuietConfig()` returns `enabled: false`. The probe is guarded by that flag (`quiet.enabled ? await probeListenerCount() : null`), and `quietGateDecision` short-circuits on `!enabled` with `proceed: true` — so there is no probe and no 30 s sleep either.
2. **Behavioural.** Zero `Analysis paused` / `Stream is quiet` / `Waiting for quiet` lines in 37 h of controller log. Child-origin events do reach my logs — `tag-library` / `tag-library-batch` show up there — so the absence means something.
3. **Temporal.** The tagger child only existed for about 20 s (02:00:37–02:00:57 UTC on 08-03) and 4 s on 08-01, and **not at all on 08-02**. 08-02 alone carried roughly 1,000 failures spread over 12 separate burst hours.

So all 1,942 failures in the window are the controller's own monitor loop, and the `#3 in a row` excursions were on the gate path. The margin is real. Worth keeping the process tag anyway — it is cheap and it makes this answerable from the log instead of by inference. Mine now prints `pid=` and `proc=controller|tagger-child` off `MANAGED_ENV`.

## What the 30 s actually is

It is real, and it is phase-locked to process boot. Across the whole window, **1,871 of 1,941 gaps between consecutive failures are exactly 30 s**. Per boot epoch:

| epoch (UTC) | failures | `(t − boot) mod 30` | in 4 buckets |
|---|---:|---|---:|
| 08-01 20:16:55 | 494 | 2 s–5 s | 98.8% |
| 08-02 07:54:24 | 931 | 2 s–5 s | 98.5% |
| 08-02 23:45:32 | 517 | 2 s–5 s | 99.2% |

`mod 15` returns an **identical** distribution, which means every failure lands on an even multiple of 15 s from boot: every *other* monitor tick, never the ones in between.

That ruled out `QUIET_POLL_MS` for me but left the question of what else is boot-locked at 30 s. There is exactly one other candidate in the tree, and it is enabled by default:

**`audio/ttsHeavyClient.ts` self-starts `cachedHealthProbe(...).start()` at import**, `intervalMs` 30 s, and `docker-compose.yml` wires `TTS_HEAVY_URL=${TTS_HEAVY_URL:-http://tts-heavy:8080}` unconditionally — the comment right above the service says so explicitly ("wired ... unconditionally; when the profile is off the URL is unreachable"). The `tts-heavy` profile is off by default, so on a default install that probe resolves a hostname that does not exist, every 30 s, forever.

Measured from inside my controller, four consecutive times:

```
FAILED after 1589 ms: ENOTFOUND
FAILED after 3944 ms: ENOTFOUND
FAILED after 3613 ms: ENOTFOUND
FAILED after 3637 ms: ENOTFOUND
```

Docker's embedded DNS is slow to answer NXDOMAIN on this host. `getaddrinfo` runs on the libuv threadpool (default 4 threads), which the Icecast fetch shares. The failures are also logged **2–5 s after the tick**, which tracks that stall duration rather than the 1500 ms deadline.

**What I could not do is reproduce it minimally.** Firing one probe alongside one status fetch on an otherwise idle container changed nothing (control 3–35 ms, test 4–7 ms). So either it needs the real process's threadpool load (Kokoro worker, sqlite, file I/O) to bite, or the mechanism is not the one I am describing. I would rather hand you the correlation and the caveat than a conclusion I have not earned.

It does explain the things that were odd about this, though: why it is host-specific and you could not reproduce it against a loopback fake, why direct `curl`s to Icecast are always fast, why the burst hours land on ~120 failures (3600/30 = 120, exactly one per probe), and why it is bursty at all (DNS negative caching would come and go on its own schedule).

## On the deadline

Running `eec86038` since 09:52 UTC with `STATUS_TIMEOUT_MS = 3000` and the process tag, and I will report the rate against the 485-per-11.53 h baseline.

One prediction worth writing down first, so it counts as a test rather than a story: if the probe is the cause, 3000 ms should **cut the rate but not clear it**, because three of the four stalls I measured are over 3 s. If failures do disappear entirely, the DNS theory is wrong and the deadline was simply tight.

Happy to run the A/B directly — unset `TTS_HEAVY_URL` and re-measure — if that is useful to you. On this station it is a no-op for audio, since the engine is Kokoro and the sidecar has never been started.

One thing that might be worth a look independently of my box: a health probe for an optional sidecar that is wired on by default and unreachable by default costs a DNS lookup every 30 s on every install that does not run the profile. Gating `.start()` on the sidecar actually being configured would make that cost zero, whatever it turns out to be doing to my poll deadline.

mauriciopaim:
Narrowing my own claim before you spend time on it, because I went and measured the DNS side properly and half of what I wrote is my box, not yours.

The stall is **not specific to `tts-heavy`**. Any unknown name costs the same:

```
broadcast                    OK       0 ms      (known container)
navidrome                    OK       1 ms      (known container)
tts-heavy                    FAIL  3924 ms  ENOTFOUND
nonexistent-random-xyz123    FAIL  3999 ms  ENOTFOUND
also-not-real-abc789         FAIL  3829 ms  ENOTFOUND
```

And on the host itself, outside Docker entirely, the same lookups take **~8 s**. So this is my resolver chain (systemd-resolved with only a fallback upstream configured), not Docker's embedded DNS and not anything subwave is doing. On a host that returns NXDOMAIN in single-digit milliseconds — which is most of them — the probe costs nothing and nobody would ever see this. I am going to fix my end separately.

That splits the finding cleanly, and I would rather you had the split than the headline:

- **Universal:** a 30 s health probe self-starts at import against a hostname that the default compose file guarantees does not resolve. True on every install that does not run the `tts-heavy` profile.
- **Mine:** that lookup costing seconds instead of milliseconds, which is what turns a free probe into something that can plausibly land on a poll deadline.

So the second half of my previous comment describes an amplifier that most operators do not have. Treat the `.start()` gating as a small tidiness fix worth doing on its own terms, not as a fix for #1256, and please do not go chasing Docker DNS behaviour on my account.

What does not change is the first half: the depth histogram here is not mixing processes, the tagger child never polls on this station, and the `#3 in a row` was on the gate path. That part is measured, not inferred.

The 3000 ms prediction also still stands as a clean test, and it is arguably now a better one — if my DNS is the thing feeding the deadline, this station is close to a worst case for it, so whatever margin 3000 ms buys here should be generous elsewhere. I will report the rate either way.

PR #1260, opened 2026-08-01
fix(broadcast): stop a station ident stacking onto a track's own link (#1258)

### LINKED ISSUE #1258: Station ident and spoken link air back-to-back: wall-clock idents collide with track-boundary links (#310 follow-up)
**Version:** controller 1.2.0

## What happens

A station ident and a spoken link air back to back, a few seconds apart, with no music between them. Two examples from one hour on the same station:

```
09:30:02  generateStationId          <- the :30 cron tick
09:34:06  [station-id] aired
09:34:22  [link] aired                  16 seconds later

09:45:01  generateStationId          <- the :45 cron tick
09:46:18  [link] aired
09:46:43  [station-id] aired           25 seconds later
```

## Why

The two talkers are scheduled on different clocks and never consult each other.

- Idents are **wall-clock** scheduled: `cron.schedule('15,30,45 * * * *', stationId)` in `broadcast/scheduler.ts`.
- Links are **track-boundary** scheduled: they air as the next pick starts.

The ident is generated on its tick but does not air until the next boundary, which is precisely where a link also lands, so the two stack.

This looks like the same failure mode as #310. The comment on that cron line says the ident slots deliberately avoid `:00` because firing alongside the hourly check "stacked two voice segments on each other". That de-conflicted the ident against the *other wall-clock talker*, but a link is not on the wall clock at all: its timing depends on track length, so no fixed minute can avoid it.

## Suggestion

A single "last spoke at" timestamp consulted by every voice segment before it airs, with a minimum gap, would cover ident vs link, ident vs skill segment, and any future talker, without needing each pair to be de-conflicted by hand. The per-kind cooldowns that already exist don't help here because these are two different kinds.

Happy to be told this is intended and I should just lower the persona frequency, but at `moderate` the result reads as the DJ talking twice in a row rather than as two separate beats.

perminder-klair:
Fixes #1258.

## The collision

Two talkers reach the same track boundary from different clocks:

- **Idents are wall-clock scheduled** — `cron.schedule('15,30,45 * * * *', stationId)`, which calls `runStationId({ atNextTrack: true })` → `announceAtNextTrack()`, parking the rendered WAV in `_pendingVoice` until the next track start.
- **Links are track-tied** — `airIntro()` fires when the pick's own song starts.

`onTrackStarted` then ran `airPendingVoice()` immediately before `airIntro()`, with a comment claiming `ident → link` "reads as a natural hand-off". They're different kinds, so no per-kind cooldown sees the pair, and `airVoice`'s `_voiceChain` only stops them *overlapping* — back to back with no music between is precisely what it produces. The reporter's 16s and 25s gaps are one segment's own playback length.

Their second example has the order reversed, which is a **bed**: `onBedStarted` airs the link over the instrumental bed seconds before the song starts, then the ident lands at the real track boundary.

## The fix

`airPendingVoice` now asks whether the incoming track already brings its own spoken line, and **holds** the ident for the next boundary if so.

Holding rather than dropping is deliberate: the ident isn't lost, it just waits one track, so the cadence stays close to what the frequency rung intends and nothing is regenerated — the WAV sits in the same one-slot `_pendingVoice`, and a fresh cron fire supersedes it as before. The existing `PENDING_VOICE_MAX_AGE_MS` check is moved ahead of the new one, so a long run of link-carrying boundaries drops the clip as stale instead of re-deferring it forever.

The rule is the pure `boundaryCarriesTrackVoice()` in `queue/pure.ts`:

- **Ignores `introAired`** — an item still sitting in `upcoming` whose intro already fired is the bed case above, and the listener has just heard it.
- **Defers to `shouldDropStaleLink`** — a link `airIntro` is about to drop as a stale back-announce leaves the boundary genuinely silent, so the ident airs there after all rather than trading one voice for none.
- Covers listener-request intros (`dj-speak`) as well as auto-DJ links; both are track-tied and unmovable.

`onTrackStarted`'s inline upcoming matcher is extracted to `matchUpcomingIndex()` so both call sites resolve the same incoming item rather than drifting apart. The decision runs synchronously, before `airPendingVoice`'s first `await` and before the consume+splice, so it sees the item while it's still in `upcoming`.

## What is *not* changed

- **Ident vs skill segment** — the segment director already floors against `getLastVoiceAt(['station-id','hourly-check','handoff','banter'])`. Its exclusion of track-tied kinds is deliberate (a global gap over links would mute it outright on a chatty station), so this PR leaves it alone.
- **No new setting.** The rule is structural, not a cadence dial.
- Boundaries with no link behave byte-for-byte as before.

## Note for the reporter

They said they're on `moderate` but their log shows idents at `:30` *and* `:45`. `effectiveFrequency` bumps a DJ-mode persona one rung up, so a moderate DJ-mode persona idents at :15/:30/:45 — three an hour, not two. Lowering persona frequency would only have made the collision rarer, never fixed it.

## Testing

- `controller/scripts/ident-link-collision.test.ts` — 9 assertions pinning both observed orderings (link-about-to-air, and the bed case where it already aired), the stale-link exception, and that a silent boundary still lets the ident through.
- `npm run lint` in `controller/` — 0 errors.
- `npm test` in `controller/` — all 65 test files pass.

https://claude.ai/code/session_01CQfhcgxXdaXsFrFSGpLtnZ

### DIFF
--- a/controller/scripts/ident-link-collision.test.ts
+++ b/controller/scripts/ident-link-collision.test.ts
@@ -1,0 +1,119 @@
+// Regression test for the ident-vs-link boundary guard (boundaryCarriesTrackVoice).
+// Run: `tsx scripts/ident-link-collision.test.ts`.
+//
+// Issue #1258. Two talkers reach the same track boundary from different clocks:
+//
+//   - the station ident is WALL-CLOCK scheduled (cron at :15/:30/:45), then
+//     deferred by announceAtNextTrack to the next track start;
+//   - a link/intro is TRACK-TIED — it airs when its own song starts.
+//
+// So the ident lands exactly where a link already lives. They're different
+// kinds, so no per-kind cooldown sees the pair, and airVoice's serialiser only
+// stops them OVERLAPPING — back to back with no music between is what it
+// produces. Reported as 16s and 25s gaps, i.e. one segment's own length.
+//
+// The guard answers one question — "does this boundary already speak?" — and
+// airPendingVoice holds the ident for the next boundary when it does. These
+// asserts pin both observed orderings and, crucially, that the ident is NOT
+// suppressed at a boundary that turns out to be silent (idents must still air).
+//
+// node:assert-via-tsx style, matching scripts/stale-link.test.ts.
+
+import assert from 'node:assert/strict';
+import { boundaryCarriesTrackVoice } from '../src/broadcast/queue.js';
+
+const PREV = { id: 'song-P', title: 'Previous Pulsar', artist: 'Artist P' };
+const OTHER = { id: 'song-O', title: 'Other Orbit', artist: 'Artist O' };
+
+let failures = 0;
+function test(name: string, fn: () => void) {
+  try {
+    fn();
+    console.log(`  ✓ ${name}`);
+  } catch (err: any) {
+    failures++;
+    console.error(`  ✗ ${name}\n      ${err?.message || err}`);
+  }
+}
+
+function main() {
+  console.log('ident vs track-tied link at one boundary (boundaryCarriesTrackVoice):');
+
+  test('untracked auto-playlist track (no queue item) → boundary is silent, ident airs', () => {
+    assert.equal(boundaryCarriesTrackVoice(null, PREV), false);
+    assert.equal(boundaryCarriesTrackVoice(undefined, PREV), false);
+  });
+
+  test('tracked item with no intro at all → boundary is silent, ident airs', () => {
+    assert.equal(boundaryCarriesTrackVoice({}, PREV), false);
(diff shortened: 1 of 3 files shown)


perminder-klair:
Follow-up commit da29c4e addresses review observations: `boundaryCarriesTrackVoice` now mirrors `airIntro`'s runtime drop paths too — station voice off (`settings.tts.enabled`) and a reaped WAV with no script left to re-render from both leave the boundary silent, so the ident is no longer held for it. Injected via opts (`voiceAllowed`, `wavExists`) so the rule stays pure; both call-site probes are synchronous, keeping the decision ahead of `airPendingVoice`'s first await. Also widened the item type with `introKind`/`introAired` so the deliberately-ignored fields are part of the visible contract. Test file now pins 12 assertions (was 9); lint 0 errors, all 65 test files pass.

Issue #1187, opened 2026-07-26
[Bug] Follow-up to issue #1124 (repeat artists)


Jaz666:
The v0.46 guard still permits repeats when the agent run’s `seen` map contains only the current artist. This occurs when `tracksLikeThis` returns eight same-artist results and no other tool contributes candidates. The wider library is not thin, but the guard treats the run-local result set as though no alternative exists. Could the repeat path invoke the normal fallback pool, excluding the current artist, before relaxing the guard?

perminder-klair:
You've read it right. The guard was treating the run's own tool results as the whole library: `tracksLikeThis` answers with eight tracks by the on-air artist, nothing else contributes, `seen` ends up single-artist, and the guard takes that as "no alternative exists" and lets the repeat through. Your Ride and Faith No More traces are exactly that shape.

PR #1190 does what you suggested. When the run can't produce an other-artist pick, the guard now asks the normal fallback pool for one, with that artist hard-blocked through the pool's whole starvation cascade so it can't circle back to the artist it was called to avoid. The guard only relaxes if even the pool comes up empty, and that case is logged with the candidate count so a genuine "the library had nothing" stays visible in the logs.

One thing left alone: `tracksLikeThis` still returns same-artist neighbours on purpose. Intentional album runs are a separate feature, as noted in #1124.

Issue #1246, opened 2026-07-29
tracksLikeThis effectively ranks by artist/album text when track enrichment is empty


spaceshipmike:
### Summary

On a library where track enrichment is sparse, `formatTrackText` emits **only the
head line** — `artist — title · album (year) [genre]`. Nothing about the music
enters the vector. The mood/lyric similarity tools built on that index
(`tracksLikeThis`, `searchByLyrics`) therefore rank candidates by how similar
their *label text* is, while presenting themselves to the picker agent as
mood/semantic similarity.

On my library that is **22,778 of 23,298 tracks (97.7%)**.

Two consequences, both observed on air:

1. **Dense catalogues self-cluster.** Every track by one artist begins with the
   same artist string, so a seed by that artist returns that artist. With 315
   tracks by one act, they are the tightest cluster in the space and fill top-K.
2. **Similar artist *names* rank as similar music.** `Nick Drake` and `Drake`
   share a token. As text they are close; as music they are unrelated.

I want to be clear this is not a re-report of #1124 / #1187. Those were about the
agent picker lacking an artist guard, and #1190 (v1.0.0) is running here. This is
about *why the guard keeps having nothing to choose from* — the candidate set
arriving from the similarity tool is single-artist before any guard runs.

### Expected vs. actual

**Expected:** `tracksLikeThis` returns tracks that are similar in mood, theme or
character to the seed.

**Actual:** with no enrichment present it returns tracks whose artist/album text
is lexically close to the seed's.

### Reproduction

Requires a library where most tracks have no `lyric_excerpt` and no
`lastfm_tags` (the default state after a normal scan + LLM tag run — see
"Enrichment" below for why that is easy to hit).

1. Have the agent picker on (`llm.pickerAgent: true`) and a text embedding index
   built (`embedding.enabled: true`; mine is `ollama:qwen3-embedding:0.6b`, dim 1024).
2. Let a track by an artist whose name is a substring of another artist's name
   play — e.g. `Nick Drake`.
3. The next agent pick calls `tracksLikeThis` and returns tracks by the other
   artist.

Observed sequence (times UTC, `ai-pick` log, tool from the LLM call trace):

```
22:32:59  Nick Drake — Parasite                     [similarSongs]
22:36:35  Drake — Wu-Tang Forever                   [tracksLikeThis]   <- name jump
22:40:01  Drake — Tie That Binds                    [tracksLikeThis]
[...]

PR #1261, opened 2026-08-01
fix(picker): give the artist guard's re-pick recency memory and lead-artist keys (#1251)

### LINKED ISSUE #1251: Back-to-back artist guard causes every-other-slot repeats: its re-pick has no artist-recency memory, and artistKey misses collaborations
**Version:** v1.2.0 · agent picker path (dj-agent.ts artist guard → repickFromSeen)

## Summary

The #1124/#1187 back-to-back artist guard correctly vetoes a same-artist pick,
but its constrained re-pick excludes **only the on-air artist** — it has no
memory of what played in the slots before. When one artist ranks high in the
run's candidate set (common on any catalogue with a deep soul/funk bench), the
re-pick returns to that artist every time the guard fires, producing a stable
A→B→A oscillation: no *adjacent* repeats, but the same artist every other slot.

Separately, the guard's `artistKey` comparison is a raw lowercase compare, so
`"Marvin Gaye & Tammi Terrell"` ≠ `"Marvin Gaye"` and collaborations slip
through the guard entirely.

## Reproduction (live log, v1.2.0, 23k-track library)

Five consecutive slots during a mid-energy daytime show. All three Marvin Gaye
plays were chosen **by the guard's re-pick**, not by the agent:

```
[playing] Living in Fame — The Clash
[picker] back-to-back artist "The Clash" avoided — re-picked "Heavy Love Affair" by Marvin Gaye from 6 other-artist candidate(s)
[playing] Heavy Love Affair — Marvin Gaye
[playing] Ob‐La‐Di, Ob‐La‐Da — The Beatles
[picker] back-to-back artist "The Beatles" avoided — re-picked "You're What's Happening in the World Today" by Marvin Gaye from 4 other-artist candidate(s)
[playing] You're What's Happening in the World Today — Marvin Gaye
[playing] Fun — Sly & the Family Stone
[picker] back-to-back artist "Sly & the Family Stone" avoided — re-picked "If I Could Build My Whole World Around You" by Marvin Gaye & Tammi Terrell from 6 other-artist candidate(s)
[queued] If I Could Build My Whole World Around You — Marvin Gaye & Tammi Terrell
```

Net: Marvin Gaye in 3 of 5 slots — every one placed by the anti-repeat
mechanism. The third is also the collaboration case: had the previous track
been solo Marvin Gaye, the guard would not have recognized
"Marvin Gaye & Tammi Terrell" as the same artist.

## Why it happens

1. The guard filters the run's `seen` by `artistKey(s) !== curArtist` — one
   artist removed, nothing else. Whatever ranks next-highest for the show's
   filters wins, and it wins **again on the next guard trigger**, because
   nothing records that it just played. The agent path deliberately carries no
   recentArtists/maxPerArtist filter (#618, understandably — a hard strip
   gutted niche pools), but the *re-pick* inherits that absence at exactly the
   moment variety is the goal.
2. `artistKey` (music/recency.ts) is a raw lowercase compare. Any
   `X & Y` / `X feat. Y` string is a distinct key from `X`, so both the guard
   check and its candidate filter treat collaborations as unrelated artists.

## Suggested fixes

1. **Give the re-pick a short recency window.** When filtering `alt`,
   exclude artists seen in the last N plays (N=5 would have prevented all
   three repeats above), falling back to the current single-exclusion filter
   when that empties the pool — same never-starve philosophy as #1187's pool
   rescue.
2. **Normalize artistKey**: strip a trailing ` & …` / ` feat. …` /
   ` featuring …` segment (or compare on the primary artist prefix) so
   collaborations share a key with the lead artist. One-line change in
   recency.ts; fixes both the guard check and the candidate filter.

The two compose: without (2), a window of solo-Marvin plays still admits
"Marvin Gaye & Tammi Terrell"; without (1), normalization alone still allows
the A→B→A oscillation with a fresh artist each cycle.

## Environment

- SUB/WAVE v1.2.0, all-container deploy
- Library: 23k tracks, well-enriched (Last.fm tags ~96%, lyrics ~78%, full
  CLAP audio embeddings) — the candidate pools themselves are healthy; this is
  purely the choice layer.
- Related: #1124, #1187, #1190 (guard lineage) — this report is about the
  guard's *re-pick*, which those did not change.

perminder-klair:
Fixes #1251.

## What was wrong

Two independent defects in the same guard, which compose into the reported behaviour:

1. **The re-pick had no memory.** When the back-to-back artist guard vetoed a same-artist pick, its constrained re-pick filtered the run's candidates by "not the artist I'm avoiding" and nothing else. On a catalogue with a deep bench for the show's filters, whichever artist ranked next-highest won that re-pick — and won it again on the next guard trigger, because nothing recorded that they had just played. No *adjacent* repeats, but the same artist every other slot (reported live: Marvin Gaye in 3 of 5 slots, every one placed by the anti-repeat mechanism).
2. **`artistKey` was a raw lowercase compare**, so `"Marvin Gaye & Tammi Terrell"` was simply a different artist from `"Marvin Gaye"` — collaborations walked past both the guard check and its candidate filter.

## The fix

**`artistRootKey` (`music/recency.ts`)** — a MATCHING key that collapses a credit onto its lead act. Feature credits (`feat.` / `ft` / `featuring`, bare or parenthesised) always strip; a join (`&` / `and` / `+`) strips *unless* the tail starts with `the`. That exception is load-bearing: `X & the Y` is one act, and without it half the soul/Motown bench keys onto a first name. `"Sly & the Family Stone"` stays whole, `"Marvin Gaye & Tammi Terrell"` → `marvin gaye`.

`artistKey` itself is deliberately **unchanged**. It is the identity key: it feeds `trackKey`, whose `title|artist` shape must stay byte-identical to the keys `queue.recentlyPlayed` builds from raw tag text. Collapsing it would silently stop every count-based no-repeat key from matching its own recent-play row.

**`broadcast/dj-agent/artist-guard.ts`** (new, pure, unit-pinned) now owns the one question the re-pick has to answer — which candidates it may choose from — instead of that policy sitting inline at the call site, matching how `dj-budget.ts` / `voice-policy.ts` / `drain-policy.ts` are split. It excludes the artists of the `ARTIST_VARIETY_WINDOW` (5) surrounding slots as well as the predecessor, and **falls back to the bare predecessor exclusion when that empties the pool** — same never-starve philosophy as #1187's pool rescue. A repeat five slots later beats a repeat one slot later, and escalating to the pool rescue would be wrong when the run genuinely did surface another artist.

**`queue.neighbourArtistRoots(n)`** is the data source: queued-and-unaired tracks (the queue's tail — a pick appends to the end), the track on air, and the last `n` *distinct* plays. Count-based and clock-independent like `recentlyPlayedByCount`, and it collapses the sidecar's double-write the same way, so `n` means n songs rather than n rows. The queued side matters because with pair-aware drains — or any request stacked ahead — the pick is not adjacent to the on-air track, and a queued track has no play row yet.

**The pool rescue's hard artist block** (`blockedArtists`) now matches a candidate's raw *and* lead key, so a collaboration can't answer a block on the artist fronting it. The rescue is steering around an act, not a string.

The booth log reports how many candidates the recency window skipped (`recencySkipped` on `pick.artistGuard`), so the narrowing is visible rather than mysterious.

## Deliberately out of scope

The pool picker's ordinary `recentArtists` filter still matches raw-against-raw. It is a *relaxable* preference on a different path, and widening it is exactly the kind of artist strip that gutted niche pools in #618 — worth doing separately, with its own evidence.

Lead-only normalisation also means `"Tammi Terrell"` and `"Marvin Gaye & Tammi Terrell"` still key apart. Matching every credited act needs the same name-vs-credit judgement on the tail that the `the …` exception exists to dodge, and the guard's job is the artist a listener just heard fronting a track.

## Verification

- New `controller/scripts/artist-guard.test.ts` — the `artistRootKey` table (collapses + the band names that must stay whole, incl. all three artists from the issue's live log), the re-pick's candidate set (on-air exclusion both ways across a collaboration, recency narrowing, wholly-recent fallback, untagged candidates kept, empty-window = pre-#1251 behaviour, single-artist run still escalates), the hard block's collaboration matching, and `neighbourArtistRoots` (distinct counting, queue tail, on-air, `n <= 0`).
- `npm test` — all 65 controller test files pass. `eslint . && tsc --noEmit` clean (0 errors).
- No live-station verification: this path only fires on a real agent pick with a same-artist collision.

https://claude.ai/code/session_0125tvkQWj8EnuocHfwhwXvU


perminder-klair:
Cross-linking: #1246 reports this same defect as its secondary finding (`artistKey` treating `Drake feat. Young Jeezy` as a different artist from `Drake` — 40+ such strings in that library, four consecutive same-artist tracks with the guard logging once). I'd started fixing it in #1265 and backed it out in favour of this PR — the identity-key / matching-key split here is the right call and mine would have moved `trackKey` with it. Worth closing that half of #1246 against this.

One edge case on the `&` rule, only as an observation: the `the …` exception catches `Sly & the Family Stone`, but two-name acts still collapse to the first name — `Simon & Garfunkel` → `simon`, `Hall & Oates` → `hall`, `Ike & Tina Turner` → `ike`. As a *matching* key that's mostly benign and sometimes right (a Paul Simon solo track probably shouldn't follow S&G back-to-back), and the cost of a false collapse is one blocked candidate rather than a bad dedup — which is exactly why keeping `artistKey`/`trackKey` out of it matters. Just flagging it in case the "Deliberately out of scope" section wants to name it.

Issue #570, opened 2026-06-24
tool_choice: "required" crashes newer Intel vLLM images


diablo581:
This may not really be a sub/wave bug, but I am filing it this way just to document it and then if you feel like it is appropriate to adjust the calls you can.  I am also filing a bug with Intel too

I have been working through a bug in the newer Intel vLLM images.  Sub/wave would consistently crash them immediately on any tool call, but not the older images.

I have whittled it down to Sub/Wave sending:

tool_choice: "required"

anything even an effectively empty tool call with that will crash the server, but changing it to:

tool_choice: "auto"

will run cleanly and I haven't been able to make it crash even on large complex Sub/Wave calls.

So like I said, this isn't specifically a S/W problem, but I wanted to document it, so @perminder-klair if you think the system can be changed so it sends auto instead of required it would probably "fix" this on the newer images.

perminder-klair:
Thanks for digging into this and narrowing it down so cleanly, that's really helpful.

You're right that it's on the vLLM side. The `required` value makes newer vLLM force the tool call through its guided-decoding backend, and some images (looks like the Intel/XPU ones) crash on that path. `auto` skips it, which is why it runs fine.

We send `required` on purpose because a lot of local models will just ignore the tool and reply with prose otherwise, so flipping it for everyone isn't safe. Instead I added an opt-in: there's a new "Forced tool calls" setting under your OpenAI-compatible provider (admin > Settings > LLM). Leave it on Required normally, switch it to Auto if your server crashes like this. Default behaviour doesn't change.

It's up in #588. Good idea filing with Intel too.

PR #1262, opened 2026-08-01
fix(picker): stop losing the pick when a seed tool comes back empty (#1247)

### LINKED ISSUE #1247: Pick agent returns the seed track's own id, losing the pick to the pool picker
### Summary

The pick agent intermittently returns **the seed track's own id** as its choice.
`enqueuePick` correctly rejects it (the on-air track is not in `extras.seen`), both
salvage stages miss, and the run is discarded — so the slot falls through to the
stateless pool picker.

Observed three times in ~14 minutes on v1.2.0, then stopped after a prompt-level
workaround. Every rejected id resolved to exactly the track that was playing when
the pick ran.

### Expected vs. actual

**Expected:** given `tracksThatSoundLikeThis(songId = <now playing>)`, the agent
returns one of the ids that tool returned.

**Actual:** it returns `songId` — the seed it just passed in.

### Evidence

Controller log, with the rejected id resolved against `tracks` afterwards:

```
23:09:11  agent returned unknown id qGlvGNc5jXlYcqkOR0DCMs  -> Neil Young — Heart of Gold
23:14:37  agent returned unknown id zHPZijh0QjNFWqAz1QuSZF  -> The Miles Davis Quintet — If I Were A Bell
23:22:46  agent returned unknown id cvclBTsYsCEdaoFYCu6OTa  -> Cannonball Adderley — Somethin' Else
```

Cross-referencing the `playing` entries in the DJ log, each of those was the track
on air at the moment the pick ran. So the ids are neither fabricated nor stale —
they are the seed, echoed back.

Notable: the LLM call itself succeeded every time. From `/debug`:

```
23:09:11  djAgentPick ok=True  tools=['tracksThatSoundLikeThis']  -> invalid id -> pool (src=frequent)
23:12:10  djAgentPick ok=True  tools=['similarSongs']             -> valid id   -> src=agent
23:14:37  djAgentPick ok=True  tools=['tracksThatSoundLikeThis']  -> invalid id -> pool (src=similar)
23:22:46  djAgentPick ok=True  tools=['tracksThatSoundLikeThis']  -> invalid id -> pool (src=embedding-similar)
```

So this is not an agent-loop or parse failure — the schema is satisfied, the id is
well-formed and real, it is simply the wrong one.

### Why it matters beyond the lost pick

Each rejection hands the slot to the pool picker, and in my case it reached for
`frequent` — i.e. the most-played albums from the underlying Subsonic server. On a
station being deliberately steered toward deeper catalogue, a silent fallback to
"the household's existing favourites" is close to the opposite of the intent, and
there is nothing in the operator-visible log that says *why* the DJ suddenly got
predictable. The `ai-pick` line looks normal; only `meta.source` distinguishes it.

### Workaround (and what it suggests)

Adding one sentence to the show brief stopped it immediately:

> Never pick the track already playing: its id is only the seed you pass these
> tools, never a valid choice.

Before: 3 of 3 `tracksThatSoundLikeThis` picks rejected.
After: **7 consecutive clean picks**, all `src=agent`, zero `unknown id` errors —
23:30 through 00:07, same session, no restart in between.

That it responds to a prompt line suggests the model simply lacks the constraint.

**Worth ruling one thing out up front:** the tools already exclude the seed from
their own results. `knnAudioById` / `knnById` pass the seed as `excludeId` into
`knnByBuffer`, which over-fetches `k + 1` and drops it
(`if (excludeId && r.id === excludeId) continue`). So the agent is *not* echoing
something it saw in the tool output — it is taking the id from the event message
in the session window, which names the currently-playing track precisely so the
agent can pass it as a seed. It has the id in its prompt, hands it to the tool,
then returns it as the answer.

That points at two fixes rather than a data-shape change:

1. **State the constraint in the `id` field description** in `PICK_SCHEMA`
   (`broadcast/dj-agent/schemas.ts`), alongside the existing "never invent or
   compose ids" — something like "never the id of the track already playing, which
   is only a seed for the discovery tools". Cheapest, and it travels with the
   schema to every provider.
2. **Recover it in the salvage path.** Treat "agent returned the on-air track's
   id" as its own case and re-pick from `seen` immediately. `repickFromSeen`
   already runs here but does not recover it, so the run reaches `pick.rejected`
   and the slot goes to the pool — which is the part that actually costs the
   operator something.

### Environment

- SUB/WAVE **v1.2.0**
- Split container deployment, external Subsonic library (Navidrome), ~23.3k tracks
- `llm.pickerAgent: true`, provider `openai-compatible`
- Text embeddings on (`ollama:qwen3-embedding:0.6b`), CLAP audio embeddings at 51% coverage
- Occurred with `tracksThatSoundLikeThis`; `similarSongs` produced a valid id in the
  same window, so I cannot say whether it is tool-specific or just where it landed.
  Since the seed id reaches the agent through the event message rather than the tool
  output, I would expect any seed-based tool to be able to trigger it.

Possibly related to #1246 only in that both surface as "the picker keeps handing
the slot to the pool" — the causes are independent.

perminder-klair:
Closes #1247.

## What was happening

The pick agent intermittently answered with the **on-air track's own id**. `enqueuePick` correctly rejected it, both salvage stages missed, and the slot fell through to the stateless pool picker — which reached for `frequent` (the household's most-played albums) on a station being deliberately steered toward deeper catalogue, with nothing in the operator-visible log to explain why the DJ suddenly got predictable.

The reporter's mechanism holds exactly as described: the pick event message hands the agent the current track's `[id: …]` so it can seed the discovery tools, and the tools all exclude the seed from their own results (`knnByBuffer` over-fetches `k+1` and drops `excludeId`). So the id is copied out of the prompt, not echoed from tool output. And nothing forbade it — *"never invent or compose ids"* is literally satisfied by an id that is real and was handed over.

## But the trigger is one step further up

The harness allows **exactly one** discovery call (`COMMIT_AFTER_STEPS = 1`) before `prepareStep` pins `activeTools: ['done']`. And `tracksThatSoundLikeThis` returns nothing whenever the seed has no CLAP vector — routine while audio analysis backfills (51% coverage in the report), because the tool is registered on whether the index holds **any** vectors, never on whether it covers **this** seed. The title fallback inside `tracksLikeThisAudio` searches the seed *id* as a title, so it can't rescue an id-seeded call either.

So the model is cornered into committing with an empty candidate set, and the only real, well-formed track id anywhere in its context is the seed. Both salvage stages are then structurally unable to help — `nearestId` has no keys to match against, `repickFromSeen` returns `null` on its first line. That is why all three observed failures were `tracksThatSoundLikeThis` while the `similarSongs` pick in the same window was clean.

The `candidates` field now on `pick.rejected` will confirm this on any install that hits it again.

## Changes

**1. Cross-index rescue for both seed-similarity tools** (`picker-tools.ts`)
Both indexes answer the same question. When the one the model reached for holds no vector for the seed, answer from the other and say so in the note, instead of returning a result that can only end the run. Deliberately gated on the primary matching *nothing at all* — a primary that matched but whose hits were all filtered by recency keeps today's `emptyResult`, which steers the model correctly and is a different situation from a dead index.

**2. One shared wording for the seed rule** (`util/pick-seed.ts`)
Policy chokepoint, like `util/request-guard.ts`. Used by the pick **and** request schema `id` descriptions and by the `emptyResult.rule` the model is looking at at the exact moment it fails. The reporter proved a prompt line alone stops the echo (3/3 rejected before, 7 consecutive clean picks after), so it stays as the backstop behind change 1.

**3. A zero-candidate run no longer trips the circuit breaker** (`dj-agent.ts`)
It's a library-coverage problem, not a model that can't drive the tool loop. Counting it opens the breaker after three tracks, disables the session-aware picker for ten minutes, and points the operator at *"switch model"* — which repairs nothing. Same carve-out, and same reasoning, as the existing already-queued case directly above it.

**4. Honest logging** — `pick.rejected` carries a `cause` (`no-candidates` / `seed-echo` / `unknown-id`) and the booth log says which happened, instead of a bare "unknown id".

## Testing

- `scripts/pick-seed.test.ts` pins the clause content and all six classification cells (including the null-seed cases at boot/recover, which must not read as an echo).
- `npm run lint` clean (0 errors), `tsc --noEmit` clean, all 65 test files pass.
- Not exercised against a live station — the mechanism is pinned by unit tests and the tool-level change is behind an "index returned nothing" branch that today ends the run regardless.

## Note on the reporter's proposed fix 2

"Recover it in the salvage path" can't work as written — with zero candidates there is nothing to re-pick *from*. Change 1 attacks the reason the candidate set is empty, which is the part that actually costs the operator a pick.

Related: #1246 (their stated workaround is "steer to `tracksThatSoundLikeThis`", which is what this PR makes safe on a partially-analysed library).

https://claude.ai/code/session_015afDatKDbzDFSnVRRzhjR1


perminder-klair:
Reviewed. Two tightenings pushed in e6201e7c:

1. **Breaker exemption now requires a real discovery call.** `candidates === 0` alone conflated "discovery came back empty" with "the model never made a discovery call at all" (possible on the `toolChoice:'auto'` downgrade #570, or a provider ignoring `'required'`, with a salvage leg fabricating an id against an empty trail). That second case is exactly the can't-drive-tool-calls failure the breaker exists for — under the previous cut such a model would never trip it while every pick burned the full agent deadline. `classifyPickFailure` now takes `toolCalls` (the `done`-excluding count already at the call site) and classifies zero-discovery runs as a new `no-discovery` kind that still counts.

2. **Coherent message when the cross-index rescue's hits all get filtered.** Reporting the fallback index's raw match count with zero surviving tracks rendered "N exist but were all played recently" (about the index the model never asked) glued to "no embedding yet" (about the one it did). All-filtered rescues now fall through to `matched: 0`.

Also opened #1266 (main→develop back-merge) — merging that first shrinks this PR's diff to just the picker commits, since this branch is based on main's tip.

Left as-is, deliberately: the request path's `pick.rejected` still has no `cause` field — its seed echo is prevented by the shared schema clause and backstopped by the stateless cascade, so classification there can be a follow-up if it ever shows up in the logs.

PR #1265, opened 2026-08-01
fix(picker): give similarity real musical signal, not just label text (#1246)

perminder-klair:
Addresses the primary finding in #1246. **The secondary finding (`artistKey`) is deliberately not here** — see the bottom.

## The problem

`formatTrackText` builds a head line plus two **optional** blocks (Last.fm tags, lyric excerpt). With both empty the vector *is* `artist — title · album (year) [genre]`, so cosine similarity over it ranks by label wording — while presenting itself to the picker as mood/semantic similarity. Both reported consequences follow directly: a prolific artist shares an artist string across every track and so becomes the tightest cluster in the space, and `Nick Drake` sits next to `Drake`. On the reporter's library that shape covered **22,778 of 23,298** tracks.

## Why the proposed fix doesn't work, and what does

> the LLM tagger had completed and written `moods` and `energy` for **every** track… including them looks like it would change the character of the index substantially for zero new inference cost

It can't. The tagger runs **enrich (0) → embed (1) → seed (2) → propagate (3)**, and propagation votes over the KNN graph built *from* these vectors (`tag-library.ts` fuses `knnById` + `knnAudioById`). Moods are an **output** of the text index, so feeding them back is circular — and on a fresh library they don't exist at embed time at all.

The signal that *is* available, waveform-derived and already sitting unused, is the **analyzer's**:

```
Nick Drake — Parasite · Pink Moon (2013) [Folk]
Sound: wistful, mid-tempo, minor key
```

- **`audio_moods`** — zero-shot CLAP moods, scored from the waveform, so genuinely independent of the label.
- **tempo band from `bpm`** — a word, not a number. An embedding model reads "upbeat tempo" as a musical property and `128` as an arbitrary token. `bpm: 0` (ID3's "unset") never reads as "slow".
- **major/minor from `musical_key`** — the analyzer stores a Camelot code (`8A`), which embeds as noise; the mode is the legible half. The tonic number is deliberately dropped (two tracks sharing a tonic are no more alike than two sharing a BPM digit).

The line is **omitted entirely** when nothing has been analysed — a constant `Sound:` label on every un-analysed track would be noise shared by all of them, clustering exactly the tracks it says nothing about.

One honest trade-off: `audio_moods` derive from the CLAP vector, so the text and audio views the mood propagation fuses become somewhat less independent. It isn't a feedback loop (nothing here changes as a result), and the tags actually being propagated stay out of the text. Worth it — on the libraries that have the data, this is the difference between musical similarity and label matching.

## Versioning (needed, or the fix causes a subtler bug)

Changing the text shape without a marker means newly-embedded tracks sit slightly apart from old ones in the same KNN space. So: `EMBED_TEXT_VERSION` + `embedding_meta.text_format` (schema v18).

Kept **strictly separate** from `embeddingStale`. A model/dim change makes stored vectors unusable and blocks the next tag run; an older text format does not — those vectors still embed the same head line and still answer KNN. Folding them together would fire the red *"Embedding model changed — tagging is blocked"* banner over an advisory. `resolveIndexTextFormat` records the **oldest** shape still in the index, because a forward run only embeds tracks that have no vector yet — recording the current version would erase the one signal that a re-embed is worth running.

## The 0% Last.fm tags

Not a bug — `enrichment.lastfmTags` defaults to `null` = auto, and `lastfmEnrichEnabled` only fetches when an API key is configured. Setting one and running `--re-enrich` is the cheapest real improvement available on that library and needs no code at all. But it was invisible from the admin UI, so `labelOnlyVectors` now counts embedded tracks carrying **no** musical signal, and the Library panel says so when it dominates (≥50%, so a partly-enriched library isn't nagged).

## The `artistKey` finding is out of scope here — #1261 already has it

I first wrote it into this branch and then backed it out. #1251 / **#1261** fixes the same defect with a better design: a separate **`artistRootKey`** for *matching*, leaving `artistKey` as the *identity* key that `trackKey` depends on. My version changed `artistKey` itself (so `trackKey` moved with it) and only stripped `feat.`, where #1261 also collapses `&`/`and`/`+` joins with a `the …` exception. Two changes to one function in flight at once is worse than either alone, so this PR is now purely the embedding-signal half.

## Testing

- `scripts/similarity-signal.test.ts` — 13 cases: tempo bands, `bpm: 0`/negative/NaN never reading as a tempo, Camelot parsing (incl. non-Camelot input contributing nothing), **v1 text byte-identical when nothing is enriched or analysed** (the whole install base is embedded at that shape), line ordering, determinism, and all four `resolveIndexTextFormat` cells.
- `npm run lint` clean on `controller/` **and** `web/` (0 errors), `tsc --noEmit` clean on both, all 65 controller test files pass.
- Not run against a live library — no re-embed was performed, so the on-air effect of the new signal is unmeasured. The schema migration is additive and nullable, and existing vectors keep working unchanged until an operator chooses to re-embed.

## Deliberately not done

Stripping artist/album from the head line. The reporter is right that it leaves `— Parasite · (2013) [Folk]`, and #618 already established that direction as known-bad.

Related: #1247 / #1262 (their stated workaround — steer to `tracksThatSoundLikeThis` — is what that issue shows can fail hard on a partially-analysed library).

https://claude.ai/code/session_015afDatKDbzDFSnVRRzhjR1


perminder-klair:
Two follow-ups pushed since the review:

**d9a2d0cb — review fixes.** The re-embed advisory now stands on its own instead of riding inside the label-only banner. `labelOnlyVectors` counts rows as they are today, so running analysis (the banner's own advice) dropped the share under the threshold and hid the "now re-embed" step exactly when it became the one that mattered. Also in: `musical_key` joined the label-only SQL, `keyModeWord` only accepts Camelot 1–12, `setEmbeddingMeta` requires `textFormat`, and the future-format test now says what it asserts.

**92053daa — two more signals in the v2 text.** `instrumental` from measured `vocal_ranges` (`[]` = no vocals found), minority-marked so vocal tracks get no word. And `Era: 1990s` as its own line, the decade word from `resolveEraYear`. Era is label-derived, so it doesn't count as musical signal in `labelOnlyVectors`. Still format v2: v2 never shipped, so operators re-embed once.

A pace energy word (`calm`/`high energy`) was implemented and backed out. Per-track pace means on a fully-analysed library sit at 0.02–0.5 (mean 0.09), so any fixed threshold writes `calm` on 99.5% of tracks while CLAP calls the same tracks `energetic` in the same line. Documented on `TrackAcoustics` so it isn't re-tried.

Verified against a copy of the real library: v17→v18 migration, all four banner states in the admin panel, and a full reseed (1413 vectors, 62s on local Ollama) that stamps `text_format=2` and clears the advisory.

PR #1278, opened 2026-08-02
Design contrast refresh: deepen the newsprint

perminder-klair:
## What this is

A contrast + character pass across the web UI — landing, admin, onboarding, and all six player skins — with the newsprint identity kept intact (radius 0, ink borders, editorial type) and depth coming from tonal surface tiers rather than shadows. Iterated live against the dev stack with operator review on every landing change.

## Foundation

- **Real second accent ink.** `--accent-2` was a no-op alias of `--accent` in classic-light/dark; it's now a genuine steel blue (and flare's low-contrast second ink deepened to burnt umber). Adopted by charts (peak lines, gauge peak-hold, playlist target arc), skins, and the landing's new section devices.
- **Surface tiers over shadows.** Admin `--card-bg` rides the theme's `--surface`, so cards lift off the page on all 8 themes. A hard-offset "print plate" drop shadow was tried and **explicitly rejected as AI-slop** — it's fully removed (including the pre-existing PressRun one) and the ban is written into `.design-sync/conventions.md`.
- **Base fixes:** system-dark `--muted` unified with explicit-dark (`#c1c0bd`); light `--muted` raised to ~5.6:1; V3Alert's hardcoded `#c5302a` → `--destructive`; AI theme-fill schema expanded from 7 to 13 color tokens (foundation tokens optional so weak local models still emit valid themes); the manual's phantom theme list (Sunset/Lab/Press/Neon) replaced with the real 8.

## Landing

Three sections got **structurally distinct redesigns** so the headline+blurb card grid stops carrying the whole page:

- **Part Two · The DJ** — the three habits print as one small hour of the booth log: `03:04 PICK` / `03:07 LINK` / `04:00 HANDOFF`, mono timestamps + kind chips down a hairline spine, echoing the dash screenshot above it.
- **Part Three · The Stack** — a wiring-manual seam sheet: dotted patch leaders (`SEAM A ····· THE MIND`), copy at full measure, a wrapping FITS rail of provider chips.
- **Part Five · The Console** — a numbered directory of the console's real departments (Dash → System), ghosted mono numerals + second-ink slugs, ruled rows.

Plus: the request band prints on a quieted accent tint (full-bleed), the hero strip re-centred between its rules, station tabs scroll sideways only, whatis cards stay unboxed print columns, micro-type floor raised (+1px eyebrows/motto), tempo variation across sections. An inverted paper-on-ink band was tried and cut — sections stay paper. Also fixes a real CSS bug found on the way: same-element custom-property stash+remap is a dependency cycle that silently invalidates every token in the subtree.

## Admin + onboarding

- Cards on the surface tier; spotlight card on the settings active theme; top bar unified with the sidebar chrome; content padding breathes; micro-type floor raised (eyebrows/captions/kv/metric labels/field hints); admin muted 55→62% ink.
- Show colours blend 25% toward the theme paper so the 12-hue palette sits in-palette on blueprint/flare (own commit, one-revert).
- Onboarding rewritten fully on-system: shared Button/Input/V3Alert, eyebrow voice, Fraunces headings — no more rounded corners or raw amber/sky/green Tailwind colors that broke under themed palettes.

## Skins (one commit each)

- **classic**: ambient art wash 32/22 → 45/32 (58/42 dark), waveform 0.22 → 0.34/0.45, console deck tints deepened with inset wells, mood/up-next on the second ink.
- **subamp**: real ink-tinted desk (the old `#fff` mix inverted on dark themes), line-strength window frames, era-correct accent-soft titlebar gradient, cover-tint radial, two-color Winamp analyzer caps.
- **platter**: metadata column brought up to the deck — surface panels, bigger sleeve (104/128px), thicker progress, second-ink markers.
- **drift**: chrome inked (10px muted → 11px ink), dark-theme wash saturation lift, listen gate gets a paper backing.
- **tty**: flat by creed — outer chrome to full ink borders, `:log` joins `:req` as a command-color pair.
- **unit**: status readout on accent-2, grille art tint deepened.
- Subamp + platter admin picker miniatures updated to match.

## Verification

- `web` + `controller` lint green; controller theme-token tests pass; `gen:themes` mirror in sync (CI drift check).
- Every landing change verified live in the browser during operator review.
- Pre-existing failures on develop, unrelated: `aio-log-link.test.ts`, `analyzer-python.test.ts`.

## Remaining before merge

- [ ] Re-capture the 8 Press Run gallery plates against a running station (`cd web && npm i --no-save playwright sharp && npx tsx scripts/capture-gallery.mjs`) — skins changed, so the shipped screenshots are stale.
- [ ] Final theme pass across the 8 palettes light+dark (`s`/`t`/`d` hotkeys; stress cases flare, blueprint, signal).

### DIFF
--- a/.design-sync/conventions.md
+++ b/.design-sync/conventions.md
@@ -1,6 +1,6 @@
 # SUB/WAVE UI — build conventions
 
-**Look**: "newsprint broadsheet meets hi-fi gear" — cream paper (`--bg: #f3efe6`), near-black ink, one hot-vermilion accent, **sharp corners everywhere** (radius is pinned to 0 — never use `rounded-*` classes), 1px ink borders instead of shadows, uppercase letter-spaced mono labels.
+**Look**: "newsprint broadsheet meets hi-fi gear" — cream paper (`--bg: #f3efe6`), near-black ink, one hot-vermilion accent, **sharp corners everywhere** (radius is pinned to 0 — never use `rounded-*` classes), 1px ink borders plus the hard-offset print-plate shadow (`shadow-plate`, `2px 2px 0 var(--shadow-ink)`) as the one elevation device — never soft/diffuse shadows — uppercase letter-spaced mono labels.
 
 ## Setup
 
@@ -19,7 +19,7 @@ Semantic classes (all verified in `_ds_bundle.css`):
 | Borders | `border-ink` (strong 1px), `border-separator-soft`, `border-separator-strong` |
 | Accent | `bg-primary` (vermilion fill), `text-vermilion` |
 | Type | `font-mono` (JetBrains Mono — data/labels), `uppercase`, `font-bold`, `tracking-eyebrow` (0.2em) |
-| Effects | `shadow-drawer` (the one sanctioned shadow — drawers/dialogs only) |
+| Effects | `shadow-plate` / `shadow-plate-lg` (hard-offset print shadow — raised cards, framed figures), `shadow-drawer` (drawers/dialogs only) |
 
 CSS variables (defined in `:root`, dark-mode aware): `--bg`, `--ink`, `--ink-soft` (hover wash), `--muted`, `--accent` (vermilion), `--field`, `--separator-soft`, `--separator-strong`, `--destructive`, `--drawer-shadow`, `--font-display` (Fraunces serif — headlines), `--font-sans` (Plus Jakarta Sans — body), `--font-mono`.
 
--- a/.design-sync/conventions.md
+++ b/.design-sync/conventions.md
@@ -1,6 +1,6 @@
 # SUB/WAVE UI — build conventions
 
-**Look**: "newsprint broadsheet meets hi-fi gear" — cream paper (`--bg: #f3efe6`), near-black ink, one hot-vermilion accent, **sharp corners everywhere** (radius is pinned to 0 — never use `rounded-*` classes), 1px ink borders plus the hard-offset print-plate shadow (`shadow-plate`, `2px 2px 0 var(--shadow-ink)`) as the one elevation device — never soft/diffuse shadows — uppercase letter-spaced mono labels.
+**Look**: "newsprint broadsheet meets hi-fi gear" — cream paper (`--bg: #f3efe6`), near-black ink, one hot-vermilion accent, **sharp corners everywhere** (radius is pinned to 0 — never use `rounded-*` classes), 1px ink borders and tonal surface tiers instead of shadows (no drop shadows on boxes — the hard-offset "print plate" shadow is explicitly banned as AI-slop), uppercase letter-spaced mono labels.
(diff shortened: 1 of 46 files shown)


perminder-klair:
### Dark-mode follow-up (e8e4661)

Two skin values tuned for dark were not reaching listeners on a dark OS. Root cause turned out to be the `dark:` variant itself, so the fix is one definition rather than a patch per call site.

**The divergence.** shadcn's stock recipe matches `[data-theme='dark']` alone, which is safe there because next-themes always writes an attribute. This app deliberately doesn't — `clearMode()` removes it to hand the document back to `prefers-color-scheme`:

https://github.com/perminder-klair/subwave/blob/e8e4661097ce22f2d4c24a1c660aec177ddf826f/web/lib/theme.ts#L117-L124

So a listener on a dark OS who never touched the switcher carries no attribute at all. The `:root` media block still flipped their tokens, but every `dark:` utility went unapplied — a fully dark page wearing light-tuned raw values.

**The fix.** `dark:` now matches both routes, with the `:not([data-theme='light'])` guard mirroring the token block exactly so pinning light on a dark OS still reads light:

https://github.com/perminder-klair/subwave/blob/e8e4661097ce22f2d4c24a1c660aec177ddf826f/web/app/globals.css#L23-L34

Drift's wash is raw CSS and never went through the variant, so it gets the same pair by hand:

https://github.com/perminder-klair/subwave/blob/e8e4661097ce22f2d4c24a1c660aec177ddf826f/web/components/skins/drift/Drift.module.css#L26-L35

**What this touches.** Nine call sites, all wanting "the palette is dark" and none wanting "explicitly pinned". Beyond the classic waveform opacity and the drift wash: the shiki code block's dark palette, `conversation.tsx`'s `dark:bg-background` (set only under dark, so gating it dropped the property entirely), and the field / input-group opacity steps. `Waveform.tsx` needs no change now and is back to its original bytes.

**Not the `.dark` class.** `syncDarkClass` already maintains one that's correct for both routes, which looks like a tidier target — but no CSS anywhere styles off it, and its listener only feeds React state:

https://github.com/perminder-klair/subwave/blob/e8e4661097ce22f2d4c24a1c660aec177ddf826f/web/components/ThemeProvider.tsx#L115-L126

A live OS flip would repaint the tokens via pure CSS while that class went stale — the same styles-disagree-with-palette bug in a new coat. The media query is true by construction.

**Verification.** Compiled `globals.css` through `@tailwindcss/postcss` and asserted both the attribute rule and the media rule emit for four representative utilities (waveform opacity, shiki bg, conversation bg, field `primary/10`) — all `attr OK / media OK`. `npm run lint` in `web/` passes with 0 errors.

Worth one manual look with the OS in dark and the switcher untouched: verifying via the `s`/`t`/`d` hotkeys sets an explicit pin, which is the path that already worked.

PR #1280, opened 2026-08-02
fix(stem-cache): keep dirFor inside the cache root for dot-only ids

eeshsaxena:
### What
`stem-cache.dirFor()` sanitizes track ids with `path.basename` to keep the per-track stems dir inside the cache root. But `path.basename` returns `.` / `..` verbatim, and `path.join(stemsRoot(), "..")` resolves to the **parent** of the cache root. A hostile or degenerate Navidrome id of `.` or `..` would therefore point stem writes (`stem-blend`, `analyze` -> `stems_dir`) outside the cache root, breaking the module's documented "can never escape the cache root" invariant.

### Fix
Neutralise any empty or dot-only basename to a safe in-root token so `dirFor` always stays under `stemsRoot()`. Verified: `dirFor("..")`, `dirFor(".")`, `dirFor("")` now all stay inside the root.

### DIFF
--- a/controller/scripts/stem-backfill.test.ts
+++ b/controller/scripts/stem-backfill.test.ts
@@ -20,7 +20,7 @@
 import assert from 'node:assert/strict';
 import { mkdtempSync, mkdirSync, writeFileSync, rmSync } from 'node:fs';
 import { tmpdir } from 'node:os';
-import { join } from 'node:path';
+import { join, sep } from 'node:path';
 
 let failures = 0;
 function test(name: string, fn: () => void | Promise<void>) {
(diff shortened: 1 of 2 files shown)


perminder-klair:
Thanks for this. I pushed a commit onto your branch adding the regression pin to `controller/scripts/stem-backfill.test.ts`, which already runs stem-cache against a temp `STATE_DIR`, so it needed no new setup.

Two tests. The first checks containment for `..`, `.`, `""`, `...`, `a/..`, `x/../..` and `/etc/passwd`. The second is the inverse: a well-formed id still maps to its own dir. A sanitiser that rewrote normal ids would orphan every cached stem set and re-separate the library, and nothing else in the suite covered that.

Verified both ways. Against the pre-fix `dirFor` it fails with `dirFor("..") escaped: /tmp/subwave-stems-XXXXXX`, the state dir, which is the parent of the stems root. With your guard, 14/14. Full controller suite 67/67, lint clean.

Your fix is complete for POSIX, since `basename` only ever returns the trailing segment, so anything it hands back that isn't dot-only is a safe single name.

One thing I left alone: `''`, `.`, `..` and `...` all land in the same `_` dir, so two degenerate ids would share one stem set and a blend could air the wrong track's stems. The ids come from Navidrome so it's close to theoretical. `safe = '_' + safe` keeps them distinct for the same line count if you want it.

PR #1288, opened 2026-08-03
feat(tts): operator-configurable fallback voice

perminder-klair:
Adds the TTS equivalent of `settings.llm.fallback`: an operator-chosen rescue **engine + voice**, editable in Settings → TTS.

## Why

TTS already rescued itself, but the chain was hardcoded and **voiceless**. A persona whose engine was unavailable (sidecar down, no cloud key) or that failed mid-render was rerouted by *engine* only — `tts.defaultEngine` → `piper` → `kokoro` — and spoke with whatever global or baked-in default that engine happened to carry. There was no way to say "if my ElevenLabs persona dies, use *this* Kokoro voice", which is exactly the control the LLM side has had all along.

## What

`settings.tts.fallback` — `{enabled, engine, voice, cloudProvider}`, default **off**.

It carries the same shape as a persona's own `tts` block, deliberately: `validateTtsBlock()` / `normalizeTts()` are exported and reused rather than copied, so the per-engine voice rules live in exactly one place, and `speakWith()` consumes the slot as a synthetic persona — the same trick `synthesizeSample()` already uses for previews.

The chain now traffics in voice **slots** rather than engine ids:

- Order is configured-fallback → `defaultEngine` → `piper` → `kokoro`.
- Dedup is by engine, **first-wins**, so a fallback naming the same engine as `defaultEngine` keeps its voice instead of reverting to that engine's global default.
- Only the configured rung carries a `personaTts` override. The hardcoded rungs still pass `null`, preserving the existing rule that a cloud rescue uses the station default's credentials rather than the provider that just failed.
- `usable()` probes the configured rung with its **own** cloud provider and the hardcoded rungs with `null`, so probe and call always agree.

One slot serves **both** triggers — the pre-flight reroute in `resolveEngine` and the mid-render rescue in `speak` — so there's a single answer to "what speaks instead". Disabled or absent is byte-for-byte the old behaviour.

## UI

New "Fallback voice" card in Settings → TTS. The engine grid, the six per-engine voice pickers and the engine-change voice normalization move out of `PersonaVoiceCard` into a shared `tts/EngineVoiceFields`, so this is the second consumer rather than a third copy of that logic. `/debug` now reports the configured fallback and whether it can currently speak.

## Verification

- `controller/scripts/tts-fallback.test.ts` extended — ordering, first-wins dedup preserving the voice, disabled reproducing today's exact chain, unusable/primary-equal fallbacks dropped, hardcoded rungs staying voiceless, and the per-rung cloud-provider probe. Full suite green (76 files).
- `npm run lint` clean in `controller/` and `web/`.
- Isolated controller on a spare port: save/round-trip, `tts.fallback.*` validation errors, partial-patch merge, and absent-key upgrade path (no `fallback` in `settings.json` → disabled defaults).
- End-to-end reroute proven via `/debug`: a persona on an unavailable engine speaks with the fallback's voice when enabled, and reverts to the built-in default when disabled.
- 19 Playwright checks against `/admin/settings` and `/admin/personas` — card render, toggle, engine switching, save persistence, and the refactored persona card (level/speed controls, engine grid, Kokoro and Cloud field swaps) — all passing, no console errors.

Spec: `docs/superpowers/specs/2026-08-03-tts-fallback-voice-design.md`

## Deliberately out of scope

Per-kind fallbacks (a different rescue for jingles vs links), and per-persona fallback voices — the station-level slot is the analogue of `llm.fallback`, and both extensions are easy to add later on top of the slot shape.


Jaz666:
Thanks for adding this. 

The new configurable fallback voice is ideal for Fish Audio users who keep matching voices in ElevenLabs. Could the TTS dispatcher strip engine-specific emotive tags before handing fallback text to engines that do not support them? Ideally, approved Fish cues would remain for Fish synthesis, while bracketed performance tags would be removed from ElevenLabs, Piper, Kokoro and other fallback paths. This would prevent the fallback voice from reading markup aloud and keep the listener-facing transcript unchanged.

Also a fallback per persona would be ideal, but it's totally understandable if that's a bit too complex.

PR #1289, opened 2026-08-03
feat(web): community apps directory at /apps

perminder-klair:
People are building clients and integrations against SUB/WAVE stations — players, bots, TUIs, MCP servers — and there was nowhere to list them. This adds a public directory at **`/apps`**, a landing-page teaser, and a one-click submission path.

Pairs with **getsubwave/community#16**, which adds `apps` as the fifth catalog type. Merge order doesn't matter: `arr()` maps a missing `apps` key to `[]`, verified against a catalog with the key absent entirely.

**No controller change.** Its registry picks named keys off `catalog.json` and ignores the rest, so `apps` is inert to a running station.

## What's here

- **`lib/apps.ts`** — the loader. Six-type vocabulary, `parseApp` skipping malformed entries, pure `appStats` / `presentTypes` / `getShowcaseApps` helpers.
- **`app/apps/`** — the directory page, shell and loading state, mirroring `/stations` (force-dynamic, promise started not awaited, Suspense around the stat strip and grid).
- **`components/apps/`** — `AppCard` (server) and `AppTypeFilter` (client). The filter wraps the server-rendered grid and drives CSS rather than rendering cards itself, so the cards stay server components and only the chip row ships JS. Chips render only for types present, so a selection can never empty the grid — there is no unreachable "nothing matched" state.
- **`components/what/TheReceivers.tsx`** — the landing teaser, after Navidrome, before the coda. **Renders nothing when the shelf is empty**, so a self-hosted station that can't reach the catalog has no dead section with a CTA pointing at an empty directory.
- **`StationFooter`** — apps becomes back page **06** (grid 5 → 6, `§§ 01–05` → `01–06`).
- Sitemap entry, and a cross-link from `/manual/clients` (which is "listen with players that already exist" — adjacent but distinct).

## Two things that would have silently broken

**Both routes that render `<Landing>`.** `/landing` always, and `/` when `SUBWAVE_HOMEPAGE=landing`. Wiring props by hand at each site is how one of them quietly loses the section — `/landing` would look right while a landing-mode homepage dropped it. Both now spread one `landingApps()` helper.

**"See all N apps"** uses the full directory count, not the capped shelf length. "See all 6" beside exactly 6 tiles reads as though that's everything.

## Images

Submitter-hosted URLs, restricted to GitHub-owned hosts in three places: the community builder rejects them at submission, `lib/apps.ts` drops the field at parse (the catalog is a **live remote fetch**, so the builder that validated it can't be assumed to have produced what we received — and an off-list URL reaching `next/image` throws at render), and `next.config.js` `remotePatterns` is the real enforcement. Images are proxied by the site rather than hot-linked, so a visitor's browser never contacts the submitter's host.

## Verification

`npm run lint` clean (3 pre-existing warnings in an unrelated file). Production build succeeds with `--webpack`; `/apps` and `/landing` both compile. (`next build` defaults to Turbopack, which panics on this worktree's symlinked `node_modules` — a known worktree limitation, not this change.)

Rendered against a stub catalog with deliberate edge cases:

| Case | Result |
|---|---|
| 9 entries, 2 malformed (no `type`, no `url`) | 7 render, 2 dropped, stat reads "7 apps · 6 kinds" |
| Off-allowlist icon + `http://` screenshot | Both fields dropped, card renders normally, no image request |
| `next/image` allowlisted vs other host | 200 (12KB jpeg) vs 400 |
| Type filter, in a real browser | Mobile→1, Terminal→1, Web→2, Integration→1, All→7, `aria-pressed` tracking |
| 280-char description | Clamped behind the shared `CatalogBrief` disclosure |
| Landing teaser | 6 tiles capped from 7, featured first then newest, "See all 7 apps" |
| `apps: []` | Teaser vanishes; `/apps` shows its empty state |
| No `apps` key at all | `/apps`, `/landing`, `/stations` all 200, no errors |

## Before this is useful

The community PR must merge first for the directory to be non-empty, and the **`app-submission`** / **`apps`** labels need creating in that repo or the submission workflow silently no-ops. Both are called out there.

The two seeded entries are first-party (the reference web player and the MCP server) — worth adding the genuinely community-built ones you know about before this goes live, so the directory doesn't launch looking like a shelf of our own stuff.


perminder-klair:
### Community side is merged and live

The catalog dependency is done — [getsubwave/community#16](https://github.com/getsubwave/community/pull/16) is merged and the published index now carries `apps`, so this PR no longer depends on anything.

- **Labels created:** `app-submission`, `app-report`, `apps` (matching the repo's existing colour conventions). I'd under-counted these as two in the description — `report-app.yml` needs `app-report` as well, or takedown reports come in unlabelled.
- **[community#19](https://github.com/getsubwave/community/pull/19) merged** — the directory is seeded with four real entries rather than launching empty: SUB/WAVE Web Player (`web`), SUB/WAVE Desktop (`desktop`), SUB/WAVE MCP Server (`integration`), Pico SUB/WAVE (`integration`).
- **[community#20](https://github.com/getsubwave/community/pull/20) merged** — adds the `skin` type, matched by the commit pushed here.

### New in this push: the `skin` type

Player skins are now a listable type, sitting between `bot` and the `integration` catch-all.

Skins are the odd entry in this directory. The player resolves its face from a **compile-time registry** (`components/skins/index.ts`), so unlike a skill or a persona nothing installs one at runtime — a listed skin is a pointer to source an operator builds into their own deployment. That's true of every entry here, but it needs saying or a reader goes looking for an install button that doesn't exist. The page prints that caveat under the grid **only when a skin is actually listed**, so a directory with none doesn't carry a note about something it doesn't have.

If skins ever become runtime-loadable, this should move to its own surface with an install button, the way skills and personas have one.

### Re-verified

Lint clean. Against a stub carrying the four real apps plus a skin:

- Chips render only present types, in vocabulary order: `All, Web, Desktop, Skin, Integration` — no empty `Mobile`/`Terminal`/`Bot` chips.
- `Skin` → Neon Face · `Desktop` → SUB/WAVE Desktop · `Web` → Web Player · `Integration` → Pico + MCP (hardware grouping under the catch-all, as designed).
- Against the **live** catalog (no skins): four cards, skin caveat absent, standard disclaimer present, landing teaser reads "See all 4 apps".

Builder still rejects unknown types — `type "theme" not in mobile|web|desktop|terminal|bot|skin|integration`.

PR #1309, opened 2026-08-04
feat(web): per-platform setup pages for macOS, Windows and Linux

perminder-klair:
The setup guide covered the install generically plus one host-specific
page (Unraid), so everything that actually differs per platform — which
Docker runtime to install, VM memory, architecture of the optional heavy
images, firewall, filesystem placement — was either missing or buried.

Adds /setup/macos, /setup/windows and /setup/linux, each a complete path
from no Docker to on-air:

- macOS: Docker Desktop / OrbStack, Resources sizing, what's native vs
  emulated on Apple Silicon (lean analyzer and the four core images are
  multi-arch; analyzer-heavy and tts-heavy are amd64-only), Rosetta,
  host.docker.internal, sleep, AirPlay's hold on ports 5000/7000.
- Windows: WSL2 + Docker Desktop WSL integration for the full CLI, or
  plain compose from PowerShell (curl.exe, not the Invoke-WebRequest
  alias). Keeps the install off /mnt/c because library.db is SQLite in
  WAL mode, plus CRLF, .wslconfig memory, login-not-boot autostart.
- Linux: Docker Engine over the distro packages (Compose v2 and
  host-gateway), firewall rules, SELinux relabelling of state/, local
  disk for WAL, rootless Docker breaking only the Stats socket proxy,
  sizing including arm64 SBCs.

Wires the three into SETUP_PAGES (so sidebar + prev/next pick them up),
the sitemap, and adds a platform picker on the setup overview alongside
pointers from Prerequisites.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_019eXyhMkKRKsTT5PbyQ8VTu


perminder-klair:
### Code review

Found 1 issue:

1. `BIND_ADDRESS=127.0.0.1` is a no-op on the compose file this page installs, so the hardening advice silently does nothing. Step 04 opens with "The default stack binds exactly one host port — `7700`, Caddy" (i.e. `docker-compose.yml`), then tells the reader to set `BIND_ADDRESS=127.0.0.1` "so nothing is exposed to the LAN at all". `docker-compose.yml` never reads that variable — its only port line is the hardcoded `"${CADDY_PORT:-7700}:80"`. `BIND_ADDRESS` is honoured solely by `docker-compose.byo.yml`, and `.env.example` scopes it explicitly: "Host interface the byo ports bind to." An operator on the default prod path who follows this leaves Caddy bound to `0.0.0.0:7700` while believing the port is closed — a false sense of security on a page whose stated purpose is host hardening. Either point the reader at `prod-byo` (as step 03 already does for the reverse-proxy case) or drop the sentence.

https://github.com/perminder-klair/subwave/blob/63e4d2040ab497554f29c781aa906a43d05a7aa0/web/components/setup/Linux.tsx#L96-L105

Two lower-confidence observations, not blocking:

- The firewall `CodeBlock` bundles two mutually-exclusive distro commands with trailing `#` comments, so a Debian reader's Copy button also grabs the Fedora `firewall-cmd` line. That is the pattern [#204](https://github.com/perminder-klair/subwave/pull/204) split apart elsewhere ("the three cliamp examples ... split into separate boxes with explanations in prose").

https://github.com/perminder-klair/subwave/blob/63e4d2040ab497554f29c781aa906a43d05a7aa0/web/components/setup/Linux.tsx#L94-L97

- All three pages size RAM for `tts-heavy` and describe its arch, but none mentions it is profile-gated (`docker compose --profile tts-heavy up -d` / `COMPOSE_PROFILES`), which [`ManualInstall.tsx`](https://github.com/perminder-klair/subwave/blob/63e4d2040ab497554f29c781aa906a43d05a7aa0/web/components/setup/ManualInstall.tsx#L172-L177) does spell out. Mostly mitigated because `subwave setup` writes `COMPOSE_PROFILES` itself.

Otherwise checked clean: route/`SETUP_PAGES`/sitemap wiring and prev/next ordering all agree, `pageMeta()` usage matches siblings, internal links resolve, and the CLI commands, ports, `ANALYZER_HEAVY`, mem-limit defaults and multi-arch vs amd64-only image claims all match the compose files and the publish workflow.

Issue #1300, opened 2026-08-04
Roundup: bugs, feature requests and docs gaps from support traffic (2026-07-08 → 2026-08-04)


perminder-klair:
Consolidated list of bugs and feature requests gathered from support traffic between **2026-07-08 and 2026-08-04**. Raising as one tracking issue; happy to split any line item into its own issue.

> **Code validation pass — 2026-08-04, against `develop` @ `d456cc61`.** Every item below carries a **Validated:** line: whether it's already fixed, real and unfixed, or needs a repro, with the file/line the verdict rests on. Summary: **4 already shipped** (bugs 1, 4, 17; FRs 7, 12 and parts of 2/14), **1 needs a repro** (bug 13), the rest confirmed. Suggested split at the bottom.

> **Ship-first progress — 2026-08-04.** All five **Ship first** items are now open as separate PRs against `develop`:
>
> | Item | PR | What shipped |
> |---|---|---|
> | Bug 11 | [#1302](https://github.com/perminder-klair/subwave/pull/1302) | Skills declare their own settings fields in `tool.mjs` — kills the class, not just the News instance |
> | Bug 5 | [#1303](https://github.com/perminder-klair/subwave/pull/1303) | Opus upgrade gated on the advertised `opusEnabled` + the `^/stream` proxy route table |
> | Bug 16b | [#1304](https://github.com/perminder-klair/subwave/pull/1304) | `streamStatus()` behind a 10s TTL + single-flight cache |
> | FR 7 + FR 12 | [#1305](https://github.com/perminder-klair/subwave/pull/1305) | `docs/custom-tts.md` (new) + "Running the analyzer on another machine" |
> | Bug 8 | [#1306](https://github.com/perminder-klair/subwave/pull/1306) | Credential form documented, plus a 401 in the app that names the cause |
>
> **Two validation calls above turned out to be wrong once the code was touched** — both are marked inline below and matter to anyone acting on this issue:
> - **FR 12** — the analyzer split is HTTP *and* filesystem, not HTTP alone. The controller prefetches each track to `/var/sub-wave/analyze-tmp` and hands the analyzer a **path**, with no url-fallback if that path can't be opened. The network-share attempts in the report had the right mechanism and needed a matching mount point.
> - **FR 7** — the in-app manual (`/manual/voices`) has documented the `remote` engine, contract and all, since #672. The gap is signposting from the README / `docs/` / `.env.example`, not the absence of docs.

[...]

PR #1311, opened 2026-08-04
fix(aio): say that ANALYZER_HEAVY does nothing on the all-in-one image (#1300 bug 9)

perminder-klair:
Ships **bug 9** from #1300 — *"`ANALYZER_HEAVY=1` silently does nothing on the AIO image"*.

## The bug

`ANALYZER_HEAVY` selects an image **tag**, and it does it by docker-compose variable interpolation:

```yaml
image: ghcr.io/…/subwave-analyzer${ANALYZER_HEAVY:+-heavy}:${SUBWAVE_VERSION:-latest}
```

The all-in-one image has no analyzer service to select — CLAP and Demucs are baked into the venv at build time, fixed by the tag you pulled. So on AIO the variable isn't merely unsupported, it's **unreachable**: zero references in `Dockerfile.aio` or `docker/aio/`.

Operators set it, watch no stem-transitions card appear, and reasonably file the feature as broken. The caveat *was* written down (`docs/unraid.md:190`, `doctor-knowledge.ts:163`) — just never at either moment the operator was actually looking.

## The fix

**Two places now say it**, and the second turned out to be the more consequential one.

### 1. A boot warning in the AIO supervisor

`warn_if_analyzer_heavy_ignored`, alongside the existing `warn_if_state_unmounted`. Fires only when the variable is set, and names the thing that actually changes the outcome:

```
WARNING: ANALYZER_HEAVY is set, and it does NOTHING on this image.
  …change this container's IMAGE to:
    ghcr.io/perminder-klair/subwave-aio-heavy
  (or -aio-cuda on an NVIDIA host). NOT subwave-analyzer-heavy — that is
  the bare analyzer micro-service, not a station image.
```

That disclaimer is load-bearing: pointing an AIO operator at `subwave-analyzer-heavy` replaces their whole station with an analysis micro-service (#966).

On a **heavy** build it's a note rather than a warning — the variable is still inert, but the capability the operator wanted is present, so there is nothing to fix and an alarming message would send someone whose setup is fine hunting a problem.

The tier is **probed** (torch in `/opt/analyzer/venv`) rather than read from a baked marker file, because torch *is* what makes the heavy tier work — a probe can't disagree with the thing it describes, a marker can drift.

### 2. The admin panel, which was actively giving AIO operators the wrong instruction

`LibraryTaggingPanel` told **every** operator to *"Set `ANALYZER_HEAVY=1` in `.env` and recreate the analyzer"* — in five places, including to AIO/Unraid users who have no `.env` at all. That copy is a large part of *why* people set the variable.

It now branches on `analysisBackend`, which `GET /library/coverage` already publishes and the panel's `Coverage` interface already declared — the same `backendLabel() === 'local'` split `doctor/checks-station.ts` makes. On a local backend the advice becomes the image swap; unknown/absent keeps the compose wording, since that's the majority install where the AIO advice would be actively wrong rather than merely incomplete.

## Testing

- `controller/scripts/aio-analyzer-heavy.test.ts` (new) — drives the function against scratch lean/heavy venvs through the existing `SUBWAVE_SUPERVISOR_LIB=1` seam (same harness as `aio-log-link.test.ts`). Pins: warn-and-name-the-image on lean; note-not-warning on heavy; **silence when unset** on both builds; and an empty value treated as unset, matching what `${ANALYZER_HEAVY:+…}` itself does with it.
- `npm run lint` clean in `web/` and `controller/`; `bash -n` clean on the supervisor.
- `npm test` in `controller/`: 76/77 pass. The one failure (`analyzer-python.test.ts` → `vocal_gate_test.py`, missing `numpy`) reproduces unchanged on `develop`.

### DIFF
--- a/controller/scripts/aio-analyzer-heavy.test.ts
+++ b/controller/scripts/aio-analyzer-heavy.test.ts
@@ -1,0 +1,136 @@
+// Regression tests for the AIO supervisor's ANALYZER_HEAVY warning
+// (docker/aio/supervisor.sh: warn_if_analyzer_heavy_ignored).
+//
+// ANALYZER_HEAVY picks an image TAG, and it does it by docker-compose variable
+// interpolation (`subwave-analyzer${ANALYZER_HEAVY:+-heavy}`). The all-in-one
+// image has no analyzer service to select — CLAP and Demucs are baked into the
+// venv at build time — so the variable is unreachable there, not merely
+// unsupported. #1300 bug 9: operators set it, watch nothing change, and
+// conclude stem transitions are broken. The caveat existed in docs/unraid.md
+// and the doctor; nothing said it at the moment they were looking.
+//
+// The load-bearing properties:
+//   1. Set + lean build  -> a warning that names the -aio-heavy IMAGE, since
+//      that is the only thing that actually changes the outcome.
+//   2. Set + heavy build -> a note, not a warning. The variable is still inert,
+//      but the capability the operator wanted is present, so there is nothing
+//      to fix and a red herring would send them chasing it.
+//   3. Unset             -> silence on both builds. This runs on every AIO boot.
+//
+// Run: `tsx scripts/aio-analyzer-heavy.test.ts`.
+//
+// node:assert-via-tsx style, matching scripts/aio-log-link.test.ts.
+
+import assert from 'node:assert/strict';
+import { execFileSync } from 'node:child_process';
+import { mkdtempSync, mkdirSync, rmSync, writeFileSync, existsSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join, dirname } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const here = dirname(fileURLToPath(import.meta.url));
+const SUPERVISOR = join(here, '..', '..', 'docker', 'aio', 'supervisor.sh');
+
+assert.ok(existsSync(SUPERVISOR), `supervisor.sh not found at ${SUPERVISOR}`);
+
+// Drive warn_if_analyzer_heavy_ignored() against a scratch venv by sourcing the
+// supervisor in library mode. Returns its stderr (the log lines the operator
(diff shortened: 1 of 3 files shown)


perminder-klair:
### Code review

Found 1 issue:

1. The panel's `analyzerIsLocal` branch gives image-swap advice to the dev-venv half of `local`, which has no container to swap. The comment states `'local'` covers "the all-in-one image, **or a dev ANALYZE_PYTHON venv**" and claims it "Mirrors the doctor's split in `doctor/checks-station.ts`" — but the doctor's `local` remedy names both cases ("On the all-in-one image, switch the container to ...; **on a dev venv, install the heavy Python deps**"), while `heavyUpgradeBox` renders only the image swap. A contributor on `ANALYZE_PYTHON` is told to "Switch this container's image ... and recreate it".

https://github.com/perminder-klair/subwave/blob/dfcbb53172b0ccc4d4137a7199755537d70800c2/web/components/admin/LibraryTaggingPanel.tsx#L360-L382

Doctor's three-way split for comparison:

https://github.com/perminder-klair/subwave/blob/dfcbb53172b0ccc4d4137a7199755537d70800c2/controller/src/doctor/checks-station.ts#L240-L244

Narrow audience, and the AIO half — the actual subject of #1300 bug 9 — is correct, including the load-bearing "NOT `subwave-analyzer-heavy`" disclaimer from #966. The supervisor warning, the `SUBWAVE_SUPERVISOR_LIB=1` test seam, and the unchanged compose-branch wording all check out.

perminder-klair:
Fixed in 6690290d — `heavyUpgradeBox` now names both remedies the way `checks-station.ts` does ("on the all-in-one image, switch this container's image to …; on a dev `ANALYZE_PYTHON` venv, install the heavy Python deps"), and the bold `ANALYZER_HEAVY` clause is backend-neutral since "does nothing on this image" doesn't parse for a bare venv. `npm run lint` clean in `web/`.

PR #1313, opened 2026-08-04
feat(shows): steer a show toward instrumental or vocal tracks (#1300 FR 13)

perminder-klair:
Ships **FR 13** from #1300 — *"Instrumental-only steering for shows"*.

## The ask

A show can lean on mood, genre, era and energy, but had no way to prefer instrumental music. The only lever was hoping the wording of the brief carried through, which it doesn't for track selection.

As the validation pass noted, the signal already exists and is already trusted: `vocal_ranges = []` is the canonical instrumental marker, load-bearing in `bed-policy.ts` (instrumental ⇒ never bed) and `embeddings.formatTrackText` (an `instrumental` minority marker). This exposes it as the show filter it always could have been.

## The shape

`shows[].vocals`, single-valued where its neighbours are lists — instrumental and vocal are mutually exclusive, and wanting both is wanting neither. `''` is no constraint and is what every existing show carries, so an upgrade changes nothing until someone sets it.

Lives in `music/show-filter.ts` so both pick paths **and** the auto-playlist coast share one definition, per the issue's note about the pool picker and the agent's discovery tools not being allowed to drift.

## The load-bearing detail: the signal is tri-state

```
[]     measured instrumental
[…]    measured vocal
null   never measured
```

Vocal analysis is the **opt-in heavy tier** (Demucs), so `null` isn't an edge case — on most libraries it's *every* track. Collapsing it to "has vocals", the obvious boolean simplification, would make an instrumental show reject its entire library while appearing to work. So unknown is carried as its own state end to end:

- **soft lean** (`preferVocals`) keeps un-measured tracks eligible, and never-starves — mirrors `preferEnergy` exactly, where the un-measured track is the unknown-energy track;
- **hard filter** (`onlyVocals`) drops them; strict means strict, same as `onlyEnergy` drops unknown-energy tracks;
- **`applyStrictLocks`** turns a zero-coverage library into a *skipped step* rather than a silent show, via the existing `starve: false` per-dimension never-starve.

Vocals sits **last** in the strict cascade deliberately: it's the dimension most likely to have no coverage at all, and last is where a skipped step costs the least.

## Agent path gets the same coverage gate, one step further out

`dj-agent.ts` already drops the mood and energy locks when the library has no such tags, so a hard lock can't empty every discovery tool for a whole show and trip the circuit breaker with a misleading *"model can't drive tools"* reading. Vocals needs it more than either, so it gets it:

```ts
const vocalLock = strict && activeShow?.vocals && library.vocalAnalyzedCount() > 0 ? … : null;
```

Counted **lazily** — only a show that actually pins vocal steering pays for the query, so `library.stats()` stays untouched on the hot path.

## Wiring

| where | what |
|---|---|
| `music/show-filter.ts` | `trackInstrumental` / `preferVocals` / `onlyVocals` + the `applyStrictLocks` step |
| `settings/{vocab,validate,normalize,persona}.ts` | `SHOW_VOCALS`, `coerceShowVocals`, schema + strict validation |
| `music/picker.ts` | soft lean in `lean()`, hard lock in the strict pool |
| `dj-agent.ts` → `picker-tools.ts` | the coverage-gated `vocalLock` |
| `prompts/picker.ts` | rendered as prose (*"instrumental tracks (no singing)"*), never the stored token — a flag name embeds as noise to a model |
| `scheduler.ts` | the auto-playlist coast, so the fallback stays on-brand |
| `web/` | a chip row in the show editor + the facet on the show card |

The chip row picks single-select by **replacing** rather than capping at one selection — a `cap={1}` would grey out the chip you're switching to.

## Testing

- `controller/scripts/show-vocals.test.ts` (new) — 9 scenarios, weighted at the tri-state semantics because that's what a future simplification would break: both spellings of unknown (`null` and an absent key), soft-keeps-unknown, hard-drops-unknown, never-starve, hard-filter-may-empty, and the two `applyStrictLocks` directions (skipped step on a zero-coverage library vs. applied when there's coverage).
- `npm run lint` clean in `web/` and `controller/`.
- `npm test` in `controller/`: 76/77 pass. The one failure (`analyzer-python.test.ts` → `vocal_gate_test.py`, missing `numpy`) reproduces unchanged on `develop`.

### DIFF
--- a/controller/scripts/picker-lock-forwarding.test.ts
+++ b/controller/scripts/picker-lock-forwarding.test.ts
@@ -1,0 +1,97 @@
+// Pins the hand-off of strict show locks from pickViaAgent (broadcast/dj-agent.ts)
+// to the agent's discovery tools (broadcast/dj-agent/agents.ts → buildPickerTools).
+//
+// Why this needs a test rather than the typechecker: agent-factory's run() takes
+// an untyped args bag and spreads it into buildTools, and buildTools destructures
+// the keys it wants by NAME. A lock that pickViaAgent resolves and passes but
+// that agents.ts forgets to name is not a type error and not a crash — it falls
+// through to buildPickerTools' `null` default and that whole dimension silently
+// stops being enforced on the agent path. The pool picker still honours it, so
+// the two pick paths drift on the same show, which is the exact thing
+// music/show-filter.ts exists to prevent. Caught in review on #1300 FR 13, where
+// vocalLock was resolved, passed, and then dropped by both lists in agents.ts.
+//
+// The check is deliberately source-level and name-based, because the defect is
+// name-based: there is no runtime seam between "passed" and "destructured".
+//
+// Run: npm test -- picker-lock-forwarding
+
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import { fileURLToPath } from 'node:url';
+import { dirname, resolve } from 'node:path';
+
+let failures = 0;
+function test(name: string, fn: () => void) {
+  try {
+    fn();
+    console.log(`  ✓ ${name}`);
+  } catch (err) {
+    failures++;
+    console.error(`  ✗ ${name}\n    ${(err as Error).message}`);
+  }
+}
+
+const here = dirname(fileURLToPath(import.meta.url));
+const djAgent = readFileSync(resolve(here, '../src/broadcast/dj-agent.ts'), 'utf8');
+const agents = readFileSync(resolve(here, '../src/broadcast/dj-agent/agents.ts'), 'utf8');
+
+// Take the text of a brace-delimited block starting at `marker`, matching braces
+// so a nested object literal can't end it early.
+function blockAfter(src: string, marker: string): string {
+  const start = src.indexOf(marker);
+  assert.notEqual(start, -1, `could not find ${JSON.stringify(marker)} — did the call site move?`);
+  const open = src.indexOf('{', start);
+  assert.notEqual(open, -1, `no opening brace after ${JSON.stringify(marker)}`);
+  let depth = 0;
+  for (let i = open; i < src.length; i++) {
+    if (src[i] === '{') depth++;
+    else if (src[i] === '}' && --depth === 0) return src.slice(open + 1, i);
+  }
+  throw new Error(`unbalanced braces after ${JSON.stringify(marker)}`);
+}
+
+// Object keys ending in `Lock`, in shorthand (`fooLock,`) or explicit
+// (`fooLock: bar`) form.
+const lockNames = (block: string): string[] =>
+  [...block.matchAll(/(?:^|[{,\s])(\w+Lock)\s*[,:}]/g)].map(m => m[1]).sort();
+
+const passed = lockNames(blockAfter(djAgent, 'pickerAgent.run('));
+const destructured = lockNames(blockAfter(agents, 'buildTools: ('));
+const forwarded = lockNames(blockAfter(agents, 'buildPickerTools('));
+
+test('pickViaAgent actually resolves some locks (the parse found something)', () => {
+  // Guards the two tests below from passing vacuously if the call site is
+  // reshaped past what blockAfter/lockNames understand.
+  assert.ok(passed.length >= 4, `expected several *Lock args, parsed: ${passed.join(', ') || '(none)'}`);
+});
+
+test('every lock pickViaAgent passes is destructured by pickerAgent.buildTools', () => {
+  const missing = passed.filter(l => !destructured.includes(l));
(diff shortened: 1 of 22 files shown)


perminder-klair:
### Code review

Found 1 issue:

1. `vocalLock` never reaches the agent's discovery tools, so a strict vocals show is unenforced on the primary pick path. `pickViaAgent` computes the lock and passes it into `pickerAgent.run({ ..., vocalLock, ... })`, but `pickerAgent`'s `buildTools` closure destructures a fixed arg list and forwards a fixed list to `buildPickerTools` — `vocalLock` is in neither, so it falls back to the `vocalLock = null` default and every discovery tool ignores it. `agents.ts` wasn't touched by this PR. Net effect: the pool-picker fallback and the `auto.m3u` coast honour the lock, the live agent path does not — the exact drift `show-filter.ts`'s header ("used by BOTH pick paths") and the PR's own `dj-agent.ts` comment ("a hard lock would empty every tool for the whole show") assume is closed. `settings.llm.pickerAgent` defaults on, so this is the common path.

https://github.com/perminder-klair/subwave/blob/a5fcbe06884097e3a58845b39b1484d10eedf290/controller/src/broadcast/dj-agent/agents.ts#L32-L44

The value is passed correctly on the sending side, so the fix is confined to that destructure/forward pair:

https://github.com/perminder-klair/subwave/blob/a5fcbe06884097e3a58845b39b1484d10eedf290/controller/src/broadcast/dj-agent.ts#L264-L268

Two lower-confidence observations, not blocking:

- The community-show install path drops `vocals`. `CommunityShow` and its sanitiser in [`community/registry.ts`](https://github.com/perminder-klair/subwave/blob/a5fcbe06884097e3a58845b39b1484d10eedf290/controller/src/community/registry.ts#L79-L85) and the install object builder in [`routes/shows.ts`](https://github.com/perminder-klair/subwave/blob/a5fcbe06884097e3a58845b39b1484d10eedf290/controller/src/routes/shows.ts#L63-L69) thread every other show music filter; `vocals` is the only one missing, so a community show pinning instrumental steering installs without it.
- `preferVocals` is only reachable from inside the strict-gated `lean()` closure, and `scheduler.ts` doesn't import it at all — so a show with `vocals` set but `filtersStrict` off gets prompt text only, no pool change. That matches `preferGenre`/`preferEra`/`preferMood` (also strict-only), but not `preferEnergy`, which the function's own docstring says it "mirrors exactly".

perminder-klair:
Pushed fixes for the review findings in 0a31763e.

**Blocking one — the vocals lock never reached the agent's tools.** `pickerAgent.buildTools` now destructures and forwards `vocalLock`. Added `controller/scripts/picker-lock-forwarding.test.ts`, which pins the *class* rather than this instance: it parses the `pickerAgent.run()` call and both arg lists in `agents.ts` and asserts every lock passed is destructured and forwarded, so the next lock added can't be dropped the same silent way. Confirmed it fails against the un-fixed lists before fixing.

**Community install** now carries `vocals` through `CommunityShow`, the catalog normaliser (via `coerceShowVocals`, so a catalog typo lands on `''` exactly as a hand-edited show does) and the install object builder. I deliberately left `docs/community.md` alone — the SHOW.md → catalog builder lives in the `subwave-community` repo and doesn't emit `vocals` yet, so documenting the key would advertise a field that's still dropped upstream. That repo needs a matching one-line change before the field actually travels.

**The soft/strict asymmetry I raised: no behaviour change, on reflection it's correct.** `preferVocals` fills `preferEnergyStrict`'s slot with `preferEnergy`'s tolerance, and that's deliberate — dropping unknowns at source level would gut every discovery source before `applyStrictLocks`' coverage-gated `onlyVocals` ever got to make that call. Added a comment saying so, since two reviewers independently read it as a bug.

Also pinned `vocals` in `resolve-show.test.ts` (the file that exists for this silently-disabled-field failure mode) and refreshed the stale dimension list in `scheduler.ts`.

`controller`: 78/78 test files pass, `tsc` clean. `web`: lint + `tsc` clean.

perminder-klair:
Companion PR opened for the publishing side: getsubwave/community#23.

That adds `vocals` to the catalog builder's field list with validation (an unrecognised value now fails `npm run check` rather than being dropped) and documents the frontmatter key in the community repo's CONTRIBUTING.

Verified end-to-end across both branches — fed the shapes that builder now emits through this branch's `coerceShowVocals`: `instrumental` → `instrumental`, `''` → `''`, **key absent** (a pre-change catalog) → `''`, junk → `''`. Old catalogs against a new station and new catalogs against an old station both degrade to "no constraint" rather than breaking.

`docs/community.md` here still doesn't mention the key. Happy to add it to this PR if you'd rather they land together — I left it out only because it would be wrong until #23 merges.

PR #1316, opened 2026-08-04
feat: expose current track lyrics to players

kylejschultz:
## Summary

- expose current-track lyrics and client-scoped per-track timing offsets to players
- rebase the feature onto current `develop`, preserving released migrations and adding `lyric_offsets` as schema v26
- protect public structured-lyrics lookups with a bounded per-track cache (30s hit, 10s miss/failure) that coalesces concurrent requests
- keep the Classic lyrics drawer polling at a low rate while the listener is buffered behind the live edge instead of resolving a fixed timeout as no lyrics
- remove the unused clear-offset API

## Validation

- `cd controller && npm test -- lyric`
- controller migration/isolation test with a real SQLite v25-to-v26 upgrade
- `cd controller && npm run lint` reaches an existing upstream TypeScript error in `src/broadcast/scheduler.ts` (`ScheduledTask.destroy`); lint otherwise has warnings only
- `cd web && npm run lint` passes with five existing warnings
- broader controller suite ran but is blocked by existing environment failures: `aio-log-link.test.ts` and Python analyzer tests require unavailable host dependencies (`numpy`)

## Rollout

No deployment is included. The migration is additive and rollback leaves the unused `lyric_offsets` table harmless.

### DIFF
--- a/controller/scripts/lyric-offsets.test.ts
+++ b/controller/scripts/lyric-offsets.test.ts
@@ -1,0 +1,54 @@
+// Integration coverage for the listener-scoped lyric-offset store and its
+// migration. Uses a real temporary SQLite database because an upsert that only
+// works in a mocked handle is not useful migration coverage.
+
+import assert from 'node:assert/strict';
+import { mkdtempSync, rmSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+
+async function main() {
+  const stateDir = mkdtempSync(join(tmpdir(), 'subwave-lyrics-'));
+  process.env.STATE_DIR = stateDir;
+  const db = await import('../src/music/library-db.js');
+
+  try {
+    await db.open({ embeddingDim: 768, adoptStoredDim: true });
+    const raw = db.requireDb();
+    assert.equal(raw.pragma('user_version', { simple: true }), 26);
+
+    // Recreate the exact predecessor shape: a deployed v25 database has every
+    // prior migration but no lyric_offsets table. Reopen must apply v26.
+    raw.prepare('DROP TABLE lyric_offsets').run();
+    raw.pragma('user_version = 25');
+    db.close();
+    await db.open({ embeddingDim: 768, adoptStoredDim: true });
+
+    const migrated = db.requireDb();
+    assert.equal(migrated.pragma('user_version', { simple: true }), 26);
+    assert.equal(
+      migrated.prepare("SELECT name FROM sqlite_master WHERE type = 'table' AND name = 'lyric_offsets'").get()?.name,
+      'lyric_offsets',
+    );
+
+    db.setLyricOffset('client-alpha', 'song-1', 1200);
+    db.setLyricOffset('client-bravo', 'song-1', -400);
+    db.setLyricOffset('client-alpha', 'song-2', 900);
+    assert.equal(db.getLyricOffset('client-alpha', 'song-1'), 1200);
+    assert.equal(db.getLyricOffset('client-bravo', 'song-1'), -400);
+    assert.equal(db.getLyricOffset('client-alpha', 'song-2'), 900);
+
+    db.setLyricOffset('client-alpha', 'song-1', 1500);
+    assert.equal(db.getLyricOffset('client-alpha', 'song-1'), 1500, 'upsert replaces only that client/track pair');
+    assert.equal(db.getLyricOffset('client-bravo', 'song-1'), -400, 'one client cannot overwrite another client');
+    console.log('✓ lyric offset migration and isolation checks passed');
+  } finally {
+    db.close();
+    rmSync(stateDir, { recursive: true, force: true });
+  }
+}
+
+main().catch((err) => {
+  console.error(err);
+  process.exit(1);
+});
--- a/controller/scripts/lyrics-cache.test.ts
+++ b/controller/scripts/lyrics-cache.test.ts
@@ -1,0 +1,53 @@
+import assert from 'node:assert/strict';
+import { BoundedLyricsCache } from '../src/music/lyrics-cache.js';
+
+let failures = 0;
+async function test(name: string, fn: () => void | Promise<void>) {
+  try {
+    await fn();
+    console.log(`  ✓ ${name}`);
+  } catch (err: any) {
+    failures++;
+    console.error(`  ✗ ${name}\n      ${err?.message || err}`);
+  }
+}
+
+console.log('public lyrics cache:');
+
+await test('coalesces concurrent requests for one track', async () => {
+  const cache = new BoundedLyricsCache<string>();
+  let calls = 0;
+  const load = async () => { calls++; return 'lyrics'; };
+  assert.deepEqual(await Promise.all([cache.get('song', load), cache.get('song', load)]), ['lyrics', 'lyrics']);
+  assert.equal(calls, 1);
+});
+
+await test('caches present lyrics longer than misses', async () => {
+  let now = 0;
+  const cache = new BoundedLyricsCache<string>({ positiveTtlMs: 30, negativeTtlMs: 10, now: () => now });
+  let hits = 0;
+  let misses = 0;
+  assert.equal(await cache.get('hit', async () => { hits++; return 'lyrics'; }), 'lyrics');
+  assert.equal(await cache.get('miss', async () => { misses++; return null; }), null);
+  now = 11;
+  assert.equal(await cache.get('hit', async () => { hits++; return 'lyrics'; }), 'lyrics');
+  assert.equal(await cache.get('miss', async () => { misses++; return null; }), null);
+  assert.equal(hits, 1);
+  assert.equal(misses, 2);
+});
+
+await test('briefly caches upstream failures to avoid a retry storm', async () => {
+  let now = 0;
+  const cache = new BoundedLyricsCache<string>({ negativeTtlMs: 10, now: () => now });
+  let calls = 0;
+  const fail = async () => { calls++; throw new Error('upstream unavailable'); };
+  await assert.rejects(() => cache.get('song', fail), /unavailable/);
+  await assert.rejects(() => cache.get('song', fail), /unavailable/);
+  assert.equal(calls, 1);
+  now = 11;
+  await assert.rejects(() => cache.get('song', fail), /unavailable/);
+  assert.equal(calls, 2);
+});
+
+if (failures) process.exit(1);
+console.log('✓ public lyrics cache checks passed');
--- a/controller/scripts/lyrics-public.test.ts
+++ b/controller/scripts/lyrics-public.test.ts
@@ -18,23 +18,29 @@ test('empty when there is no current library song', () => {
   assert.deepEqual(toPublicLyricsPayload(null, null), {
     songId: null,
     synced: false,
+    offsetMs: 0,
     lines: [],
   });
 });
 
 test('trims blank lines and preserves synced timing', () => {
   assert.deepEqual(
-    toPublicLyricsPayload('abc123', {
-      synced: true,
-      lines: [
-        { startMs: 1000, text: ' first line ' },
-        { startMs: 2000, text: '' },
-        { startMs: 3000, text: 'second line' },
-      ],
-    }),
+    toPublicLyricsPayload(
+      'abc123',
+      {
+        synced: true,
+        lines: [
+          { startMs: 1000, text: ' first line ' },
+          { startMs: 2000, text: '' },
+          { startMs: 3000, text: 'second line' },
+        ],
+      },
+      2500,
+    ),
     {
       songId: 'abc123',
       synced: true,
+      offsetMs: 2500,
       lines: [
         { startMs: 1000, text: 'first line' },
         { startMs: 3000, text: 'second line' },
@@ -45,13 +51,18 @@ test('trims blank lines and preserves synced timing', () => {
 
 test('unsynced lyrics publish null start offsets', () => {
   assert.deepEqual(
-    toPublicLyricsPayload('abc123', {
-      synced: false,
-      lines: [{ startMs: Number.NaN, text: 'plain lyric line' }],
-    }),
+    toPublicLyricsPayload(
+      'abc123',
+      {
+        synced: false,
+        lines: [{ startMs: Number.NaN, text: 'plain lyric line' }],
+      },
+      -400,
+    ),
     {
       songId: 'abc123',
       synced: false,
+      offsetMs: -400,
       lines: [{ startMs: null, text: 'plain lyric line' }],
     },
   );
--- a/controller/scripts/lyrics-public.test.ts
+++ b/controller/scripts/lyrics-public.test.ts
@@ -68,6 +68,56 @@ test('unsynced lyrics publish null start offsets', () => {
   );
 });
 
+test('an all-marker instrumental body publishes as no lyrics', () => {
+  // Timed at 0 and left unfiltered, this single line would read as a synced
+  // lyric the player highlights for the whole track.
+  assert.deepEqual(
+    toPublicLyricsPayload('abc123', {
+      synced: true,
+      lines: [{ startMs: 0, text: '[au: instrumental]' }],
+    }),
+    { songId: 'abc123', synced: false, offsetMs: 0, lines: [] },
+  );
+});
+
+test('every instrumental marker spelling is screened off', () => {
+  for (const marker of ['Instrumental', 'INSTRUMENTAL', '(instrumental)', '[Instrumental]']) {
+    assert.deepEqual(
+      toPublicLyricsPayload('abc123', { synced: true, lines: [{ startMs: 0, text: marker }] }).lines,
+      [],
+      `expected ${marker} to be screened off`,
+    );
+  }
+});
+
+test('a lyric that merely sings the word instrumental survives', () => {
+  assert.deepEqual(
+    toPublicLyricsPayload('abc123', {
+      synced: true,
+      lines: [{ startMs: 500, text: 'this is instrumental to me' }],
+    }).lines,
+    [{ startMs: 500, text: 'this is instrumental to me' }],
+  );
+});
+
+test('markers are dropped without discarding the real lines around them', () => {
+  assert.deepEqual(
+    toPublicLyricsPayload('abc123', {
+      synced: true,
+      lines: [
+        { startMs: 0, text: 'Instrumental' },
+        { startMs: 4000, text: 'the words start here' },
+      ],
+    }),
+    {
+      songId: 'abc123',
+      synced: true,
+      offsetMs: 0,
+      lines: [{ startMs: 4000, text: 'the words start here' }],
+    },
+  );
+});
+
 if (failures) {
   console.error(`\n✗ ${failures} check(s) failed`);
   process.exit(1);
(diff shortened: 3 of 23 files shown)


perminder-klair:
### Code review

Found 4 issues:

1. `toPublicLyricsPayload` doesn't filter instrumental placeholder bodies, so a track whose entire "lyric" is `Instrumental` or `[au: instrumental]` renders as a permanently-highlighted synced line (bug due to `.filter((line) => line.text.length > 0)` being the only content filter). `lyric-vocal.ts` already carries `INSTRUMENTAL_RE` for the same `subsonic.getStructuredLyrics()` shape, with a comment calling the bare placeholder "common". Since `lines.length === 1`, `LyricsDrawer` skips its "No lyrics indexed" empty state and shows the Synced badge and offset slider instead, and `startMs: 0` keeps that one line active for the whole track.

https://github.com/perminder-klair/subwave/blob/43c5633a8140176f1ff76904449d5aa9fda4b4e7/controller/src/music/lyrics-public.ts#L27-L34

2. The offset save effect PUTs to `/lyrics/current/offset` on every synced track even with no user interaction (bug due to `loadedOffsetSongRef.current = songId` at L143 running in the same callback as `setOffsetMs`, so the L158 guard is already satisfied when the effect re-runs). `lyrics?.synced` also transitions on every track, so the effect re-fires whether or not the offset changed. Each transition costs a network write plus a SQLite upsert per listener, and spends the per-IP rate budget intended for genuine corrections.

https://github.com/perminder-klair/subwave/blob/43c5633a8140176f1ff76904449d5aa9fda4b4e7/web/components/skins/lyrics.ts#L156-L161

3. A response for a different track is reported to the listener as "No lyrics indexed" (bug due to `setFailed(payload == null)` at L148 leaving `failed` false when the payload is valid but its `songId` doesn't match, so `lyrics === null && !failed && !loading` falls through to the empty-state branch). `GET /lyrics/current` answers for whatever is on air at the live edge, while `useStationFeed` deliberately holds the client ~22s behind it, so opening the drawer inside that window tells the listener an indexed track has no lyrics.

https://github.com/perminder-klair/subwave/blob/43c5633a8140176f1ff76904449d5aa9fda4b4e7/web/components/skins/lyrics.ts#L136-L149

4. The drawer paints two adjacent 1px borders (bug due to the new `.v3-drawer-content::before` drawing `border: 1px solid var(--ink)` at `inset: 0` while `sheet.tsx` widens the element itself from `border-x` to `border` and adds `relative`). With the element now positioned, `inset: 0` resolves against its padding box, placing the pseudo-element's line flush inside the element's own rather than over it, so every edge renders as a continuous ~2px rule.

https://github.com/perminder-klair/subwave/blob/43c5633a8140176f1ff76904449d5aa9fda4b4e7/web/app/globals.css#L490-L499

https://github.com/perminder-klair/subwave/blob/43c5633a8140176f1ff76904449d5aa9fda4b4e7/web/components/ui/sheet.tsx#L119-L121

perminder-klair:
Pushed fixes for all four issues as a PR into this branch: kylejschultz/subwave#1

It targets `feat/lyrics-visualization`, so merging it there folds the fixes into this PR. Controller suite (77 files) and lint/typecheck on both packages pass; the instrumental-marker case picked up four new tests in `lyrics-public.test.ts`.

Happy to drop the border change (#4) if the doubled frame was deliberate.

perminder-klair:
The `schema.ts` conflict is resolved in kylejschultz/subwave#1 — merging that into this branch clears it and brings the four review fixes with it.

**What collided:** develop shipped its own `user_version = 19` after this PR opened (the per-track analysis failure stamp, #1300 bug 3c), so both branches claim v19.

**Why it needed renumbering rather than a hand-merge:** `migrate()` gates every block on `userVersion < N`. An install that had already run develop's v19 would skip a v19 `lyric_offsets` block outright — the table would never be created, and every offset read/write would fail with `no such table` on exactly the installs that are current with develop. So develop's v19 keeps the slot it has already released, and `lyric_offsets` moves to its own v20 block, otherwise unchanged.

Verified against a real SQLite DB, both directions:

```
fresh DB -> user_version=20 lyric_offsets=true
before   -> user_version=19 lyric_offsets=false   (simulating an install on develop's v19)
after    -> user_version=20 lyric_offsets=true
```

Controller suite (88 files) and lint/typecheck on both packages pass on the merged branch.

I've left this branch untouched rather than pushing the fast-forward myself — it's yours to take.

kylejschultz:
Apologies for the noise - pushed commits to the wrong repo so reverted them - should put my PR back to where things looked good.

kylejschultz:
Happy to work on this if anything has changed to bring it up to speed.

perminder-klair:
@kanrei-subwave review this pr please

kylejschultz:
Happened to be at my desk when this came through and put this together really quick - updated the PR based on the review feedback from @kanrei-subwave:

• rebased onto current `develop` and moved the lyric-offset migration to schema v26
• added a bounded per-track cache for public lyrics lookups, including short negative/failure caching and concurrent-request coalescing
• changed stale live-edge responses to keep polling at a low rate while the lyrics drawer is open, rather than giving up after 40 seconds
• added real SQLite coverage for the migration and per-client offset isolation, plus cache behavior tests
• removed the unused clear-offset API

Focused lyric tests and web lint/typecheck pass. Controller lint/full-suite are currently blocked by unrelated existing issues in the current upstream/environment, documented in the PR description.

perminder-klair:
Thanks for staying with this through three rounds. There's a direction change on the offset before the line-level notes, and it undoes work you've already done, so here's the reasoning rather than just the decision.

## The offset should be operator-scoped, not per-listener

A lyric timing offset is a property of the track's lyrics file, not of the listener. If an LRC is 2.9s out, it is 2.9s out for everyone. The per-client model asks every listener to independently rediscover and re-fix the same defect, and almost none of them will. They'll see lyrics that lag and conclude the feature doesn't work. Fixing it once, station-wide, is the more correct fix, and it's the rule the rest of the repo already runs on: operator curation outranks listener signal.

It also deletes a lot of surface instead of making you defend it. Behind the existing admin gate, the whole public write goes away: the unauthenticated PUT, the `client_id` scheme, the inline rate limiter, and the retention question I'd otherwise have to put to you. On that last one: `client_id` is caller-supplied, the limit allows 600 writes/hour/IP, there's no per-client row cap and no delete path (you removed the clear API last round), so `lyric_offsets` only ever grows. The limiter also leaks: the sweep's `!value.hits.length` is never true for an IP that doesn't come back, so the Map gains one permanent entry per distinct IP. `checkAuthRateLimit` in the same file has the correct form.

Most of the work survives. `toPublicLyricsPayload`, the cache, the drawer, the hook, the migration (drop `client_id`, key on `track_id` alone, and the `idx_lyric_offsets_track` index becomes the primary key so it stops being dead weight). If you still want individual listeners to nudge on top of the station value, a localStorage-only nudge is fine and costs nothing.

## 1. The lyric timebase is wrong

This one matters regardless of which direction we go.

`getStructuredLyrics` returns `startMs` measured from byte zero of the file. The client computes `elapsedMs = nowMs - trackStartedAt` (`web/components/skins/lyrics.ts:221`), and `trackStartedAt` is when Liquidsoap started *playing*, which with `silenceTrim` on is `liq_cue_in` seconds into the file. Different timebases, nothing reconciling them.

A track with a 6s leading blank at the default `minGapMs: 1500` gets `cueInSec ≈ 5.75`. A line stamped `startMs: 8000` should highlight 2.25s into the listener's playback. It highlights at 8.0s. Every line, 5.75s late, on every trimmed track.

`music/silence-trim.ts:158` exports `shiftOnsetMs()` for exactly this, and it's the documented rule (root CLAUDE.md, the silence-trim bullet: every byte-zero-relative measurement resolves through it, never by local subtraction). Do the shift server-side in `toPublicLyricsPayload` so the payload is already on the trimmed timeline that every client measures in.

`/now-playing` already publishes `playableDurationSec` (`routes/public.ts:294`), so the progress bar is on the trimmed timeline and the lyrics are not. They visibly disagree inside the same drawer. `silenceTrim.enabled` defaults to `false` so this only bites operators who turned dead-air trim on, but I'd bet this is a large part of why a manual correction slider felt necessary at all. Fix the timebase and the offset stops being something listeners need to reach for.

## 2. A Navidrome outage tells listeners the track has no lyrics

`subsonic.getStructuredLyrics` catches everything and returns `null`, so the route can't tell "no lyrics indexed" from "Navidrome is down". Both return 200 with `lines: []`. The hook's `setFailed(payload == null)` is `false` because the payload is a well-formed object, and the drawer falls through to the empty state: *"Blue Monday by New Order has no lyrics in the station library yet."*

That's a positive, wrong claim about the operator's library, it's cached for the negative TTL, and there's no retry path (see #6). Needs a real failure signal out of `getStructuredLyrics`, or a distinct shape from the route.

Side effect of the same thing: because the upstream never throws, the `catch` branch in `lyrics-cache.ts` and the "briefly caches upstream failures" test can't be reached in production. The test is green for the wrong reason.

## 3. `settings.lyrics.enabled` plus a `/settings` patch-registry entry

Every other public listener feature has an operator switch: `settings.likes.enabled` returns 403, `/request` has `REQUESTS_DISABLED` and `settings.requests`, the private-station locks live in `settings.privacy`. This ships always-on.

Two real reasons an operator would want it off: it republishes full third-party lyric text on an unauthenticated endpoint, which some operators will care about; and listener polling turns into authenticated calls against their Navidrome. Right now the only way to say no is a reverse proxy rule.

## 4. Move the validation to `controller/src/schemas/lyrics.ts`

`POST /request` is the pattern: shape in `src/schemas/`, zod-only imports, enforced at the boundary, mirrored into the web bundle by `npm run gen:schemas`.

Two things fall out of doing it. The ±30000 clamp currently exists in three independent hand-written copies (`library-db/lyrics.ts:3`, the inline magic numbers in `routes/public.ts:92`, and `web/components/skins/lyrics.ts:8`) and collapses to one. And it's what unblocks the route tests kanrei asked for last round: `lyricClientId`, `lyricOffsetMs` and the 409 guard are unexported locals inside a 900-line route file, so there's nowhere to test them from. That's why the coverage is still missing, not for want of trying.

Same applies to `PublicLyricsPayload`, currently hand-copied into `web/lib/types.ts:166` where it can drift silently.

## 5. `globals.css` and the sheet collateral

`web/CLAUDE.md`'s skin rules say to co-locate styles and never touch `globals.css`. The 27 lines added there include `[data-lyric-offset]`, which exists for one Classic drawer panel.

`sheet.tsx:125` drops `border-x border-ink` for the new `.v3-drawer-content::before`, which paints a border on all four sides. That silently adds top and bottom borders to timeline, booth, request and schedule. Those four drawers didn't ask for a visual change in a lyrics PR.

Small thing while you're in there: the `relative` you added to that `cn()` list never reaches the DOM. `cn` is `twMerge(clsx(...))`, `pos` comes later in the argument list, and tailwind-merge collapses the position group to the last one. The comment above it explains a mechanism that doesn't happen.

## 6. Retry posture and auto-scroll

The retry only fires on `stale`. A hard failure sets `failed = true`, `stale = false`, and nothing ever re-asks, while `currentLyrics` returns `null` on any non-2xx or network exception. One second of bad wifi and the drawer sits on "Lyrics are not reachable right now" for the remaining three minutes of the track. It's backwards: the recoverable case doesn't retry, the routine live-edge mismatch does.

`LyricsDrawer.tsx:57` calls `scrollIntoView` on every active-line change with no suppression while the listener is scrolling, so anyone who scrolls up to read an earlier verse gets yanked back within a few seconds. `behavior: 'smooth'` also wants a `prefers-reduced-motion` guard; `sheet.tsx` already uses `useReducedMotion` and there's a large reduced-motion block in `globals.css`.

---

The parts I'd have expected an outside contributor to get wrong, you got right. You reused `getStructuredLyrics` instead of reimplementing the Subsonic call, you exported `isInstrumentalMarker` from `lyric-vocal.ts` rather than copying the regex into a second file, and the cache keys per-track while keeping per-client data out of it. The commenting is better than most of what lands here.

Rough sizing: #1 and #2 are the real work. #3 and #4 are mechanical. Option C should be a net deletion. Call it a couple of days. The branch is currently even with `develop` and the v26 migration sits clean, but `schema.ts` picks up a migration every few weeks and you've already renumbered once, so it's worth landing before the next one rather than after.

perminder-klair:
Thanks for staying with this through three rounds. There's a direction change on the offset before the line-level notes, and it undoes work you've already done, so here's the reasoning rather than just the decision.

## The offset should be operator-scoped, not per-listener

A lyric timing offset is a property of the track's lyrics file, not of the listener. If an LRC is 2.9s out, it is 2.9s out for everyone. The per-client model asks every listener to independently rediscover and re-fix the same defect, and almost none of them will. They'll see lyrics that lag and conclude the feature doesn't work. Fixing it once, station-wide, is the more correct fix, and it's the rule the rest of the repo already runs on: operator curation outranks listener signal.

It also deletes a lot of surface instead of making you defend it. Behind the existing admin gate, the whole public write goes away: the unauthenticated PUT, the `client_id` scheme, the inline rate limiter, and the retention question I'd otherwise have to put to you. On that last one: `client_id` is caller-supplied, the limit allows 600 writes/hour/IP, there's no per-client row cap and no delete path (you removed the clear API last round), so `lyric_offsets` only ever grows. The limiter also leaks: the sweep's `!value.hits.length` is never true for an IP that doesn't come back, so the Map gains one permanent entry per distinct IP. `checkAuthRateLimit` in the same file has the correct form.

Most of the work survives. `toPublicLyricsPayload`, the cache, the drawer, the hook, the migration (drop `client_id`, key on `track_id` alone, and the `idx_lyric_offsets_track` index becomes the primary key so it stops being dead weight). If you still want individual listeners to nudge on top of the station value, a localStorage-only nudge is fine and costs nothing.

## 1. The lyric timebase is wrong

This one matters regardless of which direction we go.

`getStructuredLyrics` returns `startMs` measured from byte zero of the file. The client computes `elapsedMs = nowMs - trackStartedAt` (`web/components/skins/lyrics.ts:221`), and `trackStartedAt` is when Liquidsoap started *playing*, which with `silenceTrim` on is `liq_cue_in` seconds into the file. Different timebases, nothing reconciling them.

A track with a 6s leading blank at the default `minGapMs: 1500` gets `cueInSec ≈ 5.75`. A line stamped `startMs: 8000` should highlight 2.25s into the listener's playback. It highlights at 8.0s. Every line, 5.75s late, on every trimmed track.

`music/silence-trim.ts:158` exports `shiftOnsetMs()` for exactly this, and it's the documented rule (root CLAUDE.md, the silence-trim bullet: every byte-zero-relative measurement resolves through it, never by local subtraction). Do the shift server-side in `toPublicLyricsPayload` so the payload is already on the trimmed timeline that every client measures in.

`/now-playing` already publishes `playableDurationSec` (`routes/public.ts:294`), so the progress bar is on the trimmed timeline and the lyrics are not. They visibly disagree inside the same drawer. `silenceTrim.enabled` defaults to `false` so this only bites operators who turned dead-air trim on, but I'd bet this is a large part of why a manual correction slider felt necessary at all. Fix the timebase and the offset stops being something listeners need to reach for.

## 2. A Navidrome outage tells listeners the track has no lyrics

`subsonic.getStructuredLyrics` catches everything and returns `null`, so the route can't tell "no lyrics indexed" from "Navidrome is down". Both return 200 with `lines: []`. The hook's `setFailed(payload == null)` is `false` because the payload is a well-formed object, and the drawer falls through to the empty state: *"Blue Monday by New Order has no lyrics in the station library yet."*

That's a positive, wrong claim about the operator's library, it's cached for the negative TTL, and there's no retry path (see #6). Needs a real failure signal out of `getStructuredLyrics`, or a distinct shape from the route.

Side effect of the same thing: because the upstream never throws, the `catch` branch in `lyrics-cache.ts` and the "briefly caches upstream failures" test can't be reached in production. The test is green for the wrong reason.

## 3. `settings.lyrics.enabled` plus a `/settings` patch-registry entry

Every other public listener feature has an operator switch: `settings.likes.enabled` returns 403, `/request` has `REQUESTS_DISABLED` and `settings.requests`, the private-station locks live in `settings.privacy`. This ships always-on.

Two real reasons an operator would want it off: it republishes full third-party lyric text on an unauthenticated endpoint, which some operators will care about; and listener polling turns into authenticated calls against their Navidrome. Right now the only way to say no is a reverse proxy rule.

## 4. Move the validation to `controller/src/schemas/lyrics.ts`

`POST /request` is the pattern: shape in `src/schemas/`, zod-only imports, enforced at the boundary, mirrored into the web bundle by `npm run gen:schemas`.

Two things fall out of doing it. The ±30000 clamp currently exists in three independent hand-written copies (`library-db/lyrics.ts:3`, the inline magic numbers in `routes/public.ts:92`, and `web/components/skins/lyrics.ts:8`) and collapses to one. And it's what unblocks the route tests kanrei asked for last round: `lyricClientId`, `lyricOffsetMs` and the 409 guard are unexported locals inside a 900-line route file, so there's nowhere to test them from. That's why the coverage is still missing, not for want of trying.

Same applies to `PublicLyricsPayload`, currently hand-copied into `web/lib/types.ts:166` where it can drift silently.

## 5. `globals.css` and the sheet collateral

`web/CLAUDE.md`'s skin rules say to co-locate styles and never touch `globals.css`. The 27 lines added there include `[data-lyric-offset]`, which exists for one Classic drawer panel.

`sheet.tsx:125` drops `border-x border-ink` for the new `.v3-drawer-content::before`, which paints a border on all four sides. That silently adds top and bottom borders to timeline, booth, request and schedule. Those four drawers didn't ask for a visual change in a lyrics PR.

Small thing while you're in there: the `relative` you added to that `cn()` list never reaches the DOM. `cn` is `twMerge(clsx(...))`, `pos` comes later in the argument list, and tailwind-merge collapses the position group to the last one. The comment above it explains a mechanism that doesn't happen.

## 6. Retry posture and auto-scroll

The retry only fires on `stale`. A hard failure sets `failed = true`, `stale = false`, and nothing ever re-asks, while `currentLyrics` returns `null` on any non-2xx or network exception. One second of bad wifi and the drawer sits on "Lyrics are not reachable right now" for the remaining three minutes of the track. It's backwards: the recoverable case doesn't retry, the routine live-edge mismatch does.

`LyricsDrawer.tsx:57` calls `scrollIntoView` on every active-line change with no suppression while the listener is scrolling, so anyone who scrolls up to read an earlier verse gets yanked back within a few seconds. `behavior: 'smooth'` also wants a `prefers-reduced-motion` guard; `sheet.tsx` already uses `useReducedMotion` and there's a large reduced-motion block in `globals.css`.

---

The parts I'd have expected an outside contributor to get wrong, you got right. You reused `getStructuredLyrics` instead of reimplementing the Subsonic call, you exported `isInstrumentalMarker` from `lyric-vocal.ts` rather than copying the regex into a second file, and the cache keys per-track while keeping per-client data out of it. The commenting is better than most of what lands here.

Rough sizing: #1 and #2 are the real work. #3 and #4 are mechanical. Option C should be a net deletion. Call it a couple of days. The branch is currently even with `develop` and the v26 migration sits clean, but `schema.ts` picks up a migration every few weeks and you've already renumbered once, so it's worth landing before the next one rather than after.

kylejschultz:
I appreciate the feedback! All of that makes sense, for sure. I’ll take a pass over it in the next few days and work through the changes needed to get this cleaned up before the next migration.

PR #1322, opened 2026-08-04
fix(queue): commit the queued pick before an operator skip (#1300 bug 6)

perminder-klair:
Part of #1300 (bug 6): skipping a track could air something other than the queued next track.

## Root cause

`POST /dj/skip` was a bare telnet `skip` with no coordination with the controller's queue — and pair-aware drain (default on for DJ-mode personas) deliberately **holds** the pick out of `dj_queue` for most of each track's runtime, waiting on a successor. So an empty `dj_queue` at skip time is the *normal* state, not a race: the `fallback` drops to randomized `auto.m3u` while the admin queue shows a different "next". There's a second layer the issue's validation didn't name: even a **sent** pick loses the race if its request is still resolving (a full subhttp curl fetch) when the seam arrives — the fallback only picks a source that is *ready*.

## Fix — commit, confirm, then skip

- **`radio.liq`**: new `dj_queue_status` telnet command — `ready` / `resolving` / `empty`, from `request.status` over the pending queue. Type-checked against liquidsoap 2.4.5 (`--check` on the savonet image).
- **`broadcast/skip-policy.ts`** (new, pure): the state machine — when to commit, when the wait is satisfied, and the upgrade-skew grace path (older `radio.liq` without the command answers `unknown`; the wait then proceeds on a fixed grace after the send rather than never skipping or skipping at once). Pinned by `scripts/skip-policy.test.ts`.
- **`queue.commitBeforeSkip()`**: force-drains held items, then polls the probe until the HEAD of upcoming is sent **and** a resolved request is waiting — bounded at 20s, past which the skip proceeds anyway (ending the current track is the operator's intent) with the miss reported in the booth log and the response (`committed: false`).
- **`POST /dj/skip`** wires it in; an empty queue keeps the old bare-skip path byte-for-byte (auto.m3u is the honest next when nothing is queued). MCP and the admin dash inherit the fix — both call this route.

'empty' after a send is deliberately kept-waiting, never treated as satisfied: it can be the 1s poll race or a boundary prefetch (invisible to `queue()`, the same blind spot `dj_queue_remove` documents), and trusting it would reintroduce the bug for exactly the window the probe closes.

## Verification

- `npm test` — all 90 controller test files pass (including the new `skip-policy.test.ts`).
- `npm run lint` — clean (0 errors).
- `liquidsoap --check` on the edited `radio.liq` via `savonet/liquidsoap:v2.4.5` — OK.
- **Draft pending an on-air smoke test**: skip mid-track under a DJ-mode persona with pair-drain on, and confirm the pick shown in the admin queue airs (plus one skip with an empty queue to confirm the fallback path).


perminder-klair:
On-air smoke test done on the dev stack (worktree build, liquidsoap 2.4.5, glm-5.2:cloud picker, pair-drain on, all personas DJ-mode). Three paths exercised:

**1. Held pick — the reported bug.** Mid-track, admin queue showed *Someday — Alexey Krasnoslobodtsev* with `sent: false` and `dj_queue_status` answered `empty` — exactly the state where the old skip aired a random `auto.m3u` fill. `POST /dj/skip` returned `{ok, pending: true, committed: true}` in **1.5s** and *Someday* was on air 4s later. The booth log shows the full sequence: pick held → drain stamps applied during commit → skip → pick airing.

**2. Pick-lands-just-before-skip race.** A fresh pick was queued ~11s before the skip; commit confirmed it in 2.0s and it aired — the race resolves toward the queue's promise, not the filler.

**3. Empty queue.** With auto-pick off and nothing queued: response in **0.0s**, `{pending: false, committed: false}`, auto.m3u filled honestly — the bare-skip path byte-for-byte.

One 500 during testing turned out to be a transient TCP connect failure from the controller to the telnet port inside `skipTrack()` — pre-existing code this PR doesn't touch (the old route would have failed identically), and a 30/30 connect reliability re-test came back clean at 0–5ms. Worth a one-retry hardening someday, separate change.

The live `dj_queue_status` probe was also exercised directly over telnet in all three states.

PR #1323, opened 2026-08-04
feat: zod validation + shadcn form foundation (webhooks slice)

perminder-klair:
Implements `docs/superpowers/specs/2026-08-04-zod-forms-foundation-design.md`.

Foundation for schema-validated forms, proven end-to-end on Webhooks. Before this, `web/` had **no validation library at all**, and the controller's zod dependency was used only for LLM structured output — no HTTP request body was schema-validated anywhere.

## Status

**Reviewed, fixed, and green.** All three CI jobs pass, `develop` merged in, no mirror drift.

| Piece | State |
|---|---|
| Shared schema + mirror + route-boundary validation | landed |
| Webhooks form converted end-to-end | landed |
| Accessibility wired into the form template | landed |
| CI unblocked (lockfile regenerated for the node version CI pins) | landed |
| Code review, and every finding fixed | landed — see [Review pass](#review-pass) |
| Next conversion (`stations`) | not in this PR — see [Migration path](#migration-path) |

## Why webhooks

It was the smallest complete feature, and its rules were restated across **four files in five declarations**: the shape in `settings/vocab.ts`, again as `WebhookConfig` in `broadcast/webhooks.ts`, and a third time in `WebhooksPanel.tsx`; the event list in two files (with a comment conceding the duplication); and the rules in both `validateWebhooksStrict` and a client-side `valid()` predicate that re-implemented **only two of the server's four URL rules** — so the form could call a payload valid that the server rejected.

All five collapse to one definition.

## What landed

- **`controller/src/schemas/`** — shared zod schemas, import-restricted to `zod` only by an ESLint rule (`group: ['*','!zod']`), because the file is copied verbatim into the browser bundle.
- **Stateful rules isolated** in a non-mirrored `*-server.ts` sibling: the `authHeader: 'set'` redaction sentinel (needs the existing list), id minting (a side effect), and cross-item id de-duplication (needs sibling awareness). Keeping these out is what makes the schema file safe to mirror.
- **`controller/scripts/gen-schemas.ts`** → `web/lib/schemas.generated.ts`, drift-checked in CI — the same mechanism as the existing theme-token mirror, and for the same reason (the web package cannot import `controller/src` at build time: separate package, separate Docker build context). Guards against duplicate top-level names across schema modules, and asserts the two packages' zod ranges stay in lockstep — including that each side actually *declares* one (see below).
- **`controller/src/middleware/validate.ts`** — validates request bodies at the route boundary. The error payload is **additive**: `error` stays the flat string every existing client already reads from a 400; `fieldErrors` is new and keyed by dotted path (`webhooks.1.url`), which is also react-hook-form's `setError` syntax, so a server-only rule lands on the right input.
- **`settings.update()` remains the authoritative chokepoint.** It is reached by backup import and onboarding, paths that never touch a route, so `validateWebhooksStrict` keeps its name and signature and now calls the shared schema.
- **`web/`** gains zod + react-hook-form. Forms bind via `lib/form.ts`'s `useZodForm` and the **already-vendored** shadcn `Field` primitives — deliberately **no `ui/form.tsx`**, which would stand a second form abstraction beside one the repo already ships (and avoids the shadcn CLI, which emits Tailwind-v3 syntax that breaks this project's v4 setup).

## Deliberate behaviour change

The schema is **stricter** than the validator it replaces in three spots, each pinned by a test in `controller/scripts/schemas.test.ts`:

| Field | Was | Now |
|---|---|---|
| `enabled` | any non-`false` value coerced to `true` | non-boolean rejected |
| `authHeader` | non-strings → `''`; >500 chars silently truncated | both rejected |
| `id` | malformed id silently re-minted | rejected |

Rationale: `settings.json` is only ever written by this validator, so stored data is always well-formed. Silently truncating a 600-char auth token produces a broken header that fails later at fire time with no explanation. A hand-edited `settings.json` carrying such values will now fail import loudly.

## Accessibility, done in the template rather than retrofitted

Both gaps below would have been copied into all seven remaining forms, so they are fixed here where the cost is one file each.

**`fieldAria(baseId, error, opts)`** derives a field's ARIA from one base id. The `Field` primitives are presentational — `data-invalid` only colours the field, and `FieldError`'s `role="alert"` announces a message once when it appears but never associates it with the control, so a user tabbing **back** to a bad input got an unremarkable box whose explanation was unreachable. It returns spreadable groups rather than loose strings so no form open-codes the `-error`/`-description` suffixes; two forms deriving them differently is the same drift the schema mirror exists to prevent. Ids key off the `useFieldArray` `_rhfKey`, not the row index, so removing a row cannot hand its ids to a different row, and `aria-describedby` names only ids actually rendered.

The events picker needed a group variant: with no single labelable control, `htmlFor` would have pointed at a `<div>`. The `Field` is the named group (it already carried `role="group"`) and its label renders as a `<span>` via `asChild` instead of a `<label>` labelling nothing.

**`Pill` renders a real `<button type="button">` when it has an `onClick`.** `Badge` is a `<span>`, so a clickable pill was mouse-only — no tab stop, no Enter/Space, nothing announcing it as actionable, which made the webhook event picker unusable without a mouse. `pressed` supplies `aria-pressed` for on/off chips. Only two call sites pass `onClick` (this form and `ManualTagEditor`); the ~93 read-only pills keep the untouched `Badge` path. `text-start` is present solely so the two paths are provably style-identical — a `<button>` centres its text by default.

## CI + tooling fixes

The lockfile committed with the three new web deps was generated by npm 11.6.2, which writes a tree its own `npm ci` rejects (`@emnapi/*` transitives of the `*-wasm32-wasi` optional deps hoisted inconsistently). The web job was dying at the install step, **before eslint ever ran**. Regenerated with node 22 / npm 10.9.8 — the version `lint.yml` pins — starting from develop's lock so the only resolution delta is the three new dependencies.

**The zod lockstep check passed vacuously.** It compared `dependencies.zod` on both sides, so a zod that ever moved to `devDependencies` in *both* packages would compare `undefined === undefined` and go green having verified nothing — on a check whose whole job is guaranteeing the mirrored schema text runs against the same zod build. Confirmed against a simulated pair of manifests: the old logic printed `zod undefined on both sides` and exited 0. It now requires a declared range on each side before comparing.

## Verification

- `controller`: 93/93 test files pass, `npm run lint` 0 errors
- `web`: `npm run lint` (eslint + `tsc --noEmit`) 0 errors
- `npm run gen:schemas && git diff --exit-code` — no mirror drift
- All three CI jobs green
- All six of the converted form's behaviours verified **live** against an isolated controller + Playwright, asserting through the controller's own `settings.json` and captured network bodies rather than toasts: the listener-gate toggle firing its own request and staying togglable while a row is invalid; an untouched save re-sending `authHeader: "set"` with the stored secret surviving; and webhook ids preserved across edit+save with no `_rhfKey` leaking into the payload.
- Accessibility verified the same way, 17/17 assertions: chips are `<button type="button">` with `aria-pressed` tracking state, focusable, and toggled by both Space and Enter; an invalid URL sets `aria-invalid` plus an `aria-describedby` resolving to the element carrying the message; a valid one drops the attribute entirely rather than emitting `aria-invalid="false"`; the auth field links its description; the events group resolves its label to a `<span>`. A **keyboard-only** chip toggle round-tripped through the controller's `settings.json` — an interaction that was impossible before. Button vs span with identical classes differ in **zero of 22** computed properties at the same 108×23 box.

## Migration path

Converting a form is mechanical: add `schemas/<feature>.ts`, move stateful rules to `<feature>-server.ts`, `npm run gen:schemas`, add `validateBody()` to the route, convert the panel to `useZodForm` + `Field` + `fieldAria`.

Suggested order, cheapest-first: **stations → shows → skills → playlists → imaging → onboarding → public request boxes.** The `/settings` mega-endpoint stays last and needs its own design — most admin forms post to it, and it validates the entire settings object at once.

## Review pass

Four fixes since this PR was opened: three surfaced by a review of the branch, plus one of the follow-ups the PR had listed for itself.

**1. The mirror's collision guard could miss a name, silently.** `collectDeclarations` was a line-wise regex, and a MISS is the dangerous direction — `buildMirror` only throws on names it was told about, so an unseen name produced a mirror carrying two identical declarations with no error at generate time. That is precisely the "tsc error inside a generated file nobody may edit" the guard exists to prevent. It missed `export default function f()`, the second declarator of `const A = 1, B = 2`, destructured `const { a } = ...` and `export { x as y }`, and it judged nesting by leading whitespace rather than by syntax. It is now a TypeScript parse, which closes the misses and the earlier false positives (`const enum X` captured `enum` as the name; a declaration inside a block comment or template literal was seen) in one move. Export names and `default` are collected alongside bindings, since the flat file shares one export namespace too.

**2. Validation messages did not say which row was wrong.** `firstMessage` prefixed the dotted path only for `invalid_type`, on the theory that every other code carries a message that stands alone. It doesn't: zod's built-ins name a constraint and never a location (`invalid_format` reports the pattern, `too_big` reports the limit), and both are produced by this schema's own `id` regex and `authHeader` max. Even a custom message never names the array index, so two rows failing the same rule read identically. The path is now prefixed unconditionally — a per-code allowlist silently stops applying the moment a schema adds a constraint. `validateWebhooksStrict` parses the bare array, so it passes `webhooks` as a *root* that splices in front of the index (`webhooks.0.url: ...`) rather than string-prefixing the result (`webhooks: 0.url: ...`). `id` and `authHeader` also gained operator-facing messages, for the reason `url` already had them: zod's default text is written for a developer and leaks a raw regex into a toast.

**3. The auth-header field could be captured by a password manager.** It carried no `autoComplete`, so a manager read it as the site's own login — offering to fill it, and to *update* the saved password on submit. Every other credential box under `admin/` already sets `off`. Pre-existing rather than introduced here, but this is the form the remaining seven get copied from.

**4. A capped chip went quiet rather than unavailable.** `ManualTagEditor` dropped a chip's `onClick` once three moods were selected, sending `Pill` down the `Badge` `<span>` path — no tab stop, and nothing saying why it had stopped working, so the cap itself was invisible to a keyboard user. `Pill` now takes `disabled`, which keeps the `<button>` and marks it `aria-disabled` rather than setting the `disabled` attribute: a disabled button leaves the tab order, so a keyboard user could not land on it to hear the state either, and for a selection cap being reachable *is* the message. Verified by rendering every state — the capped chip is a `<button>` carrying `aria-pressed` + `aria-disabled` and no `disabled` attribute, the webhooks chip renders byte-identically to before, and the ~93 read-only pills still take the untouched `Badge` path.

**Verified.** 93/93 controller test files pass, both lint gates report 0 errors, the mirror drift check is green, and all three CI jobs pass. New tests pin fixes 1 and 2 specifically: that two rows failing the same rule now produce different text, that the root splices in FRONT of the index rather than beside it, and that every declaration form the regex missed now collides. Fix 4 is the one weaker link — it was verified by rendering every `Pill` state and reading the markup, which proves the load-bearing property (still a `<button>`, no `disabled` attribute, so still focusable, plus `aria-disabled`) but not real focus and key events the way the Playwright run behind this PR's earlier accessibility work did.

## Known follow-ups (non-blocking)

- `applyServerFieldErrors` is wired but structurally unreachable from this form, since both sides run the same schema. It will be exercised by the first form carrying a server-only rule.


perminder-klair:
### Code review

Merged `develop` in and resolved the two `CLAUDE.md` conflicts (both were "each side appended a bullet" — develop's `settings.js` THREE-edits note and Blocklist rules bullet kept in full, this branch's `schemas/` and zod-forms bullets added alongside). Post-merge on `0750334`: 97/97 controller test files pass, `npm run lint` 0 errors, `gen:schemas` produces no drift.

I also exercised the redaction path directly against `validateWebhooksStrict`, since that is the one place a mistake would leak or destroy a stored credential. All of it holds: the `set` sentinel preserves the stored header across an unrelated edit, resolution is keyed by id so reordering rows does not swap secrets, and `set` with no stored prior now yields `''` — the old `else if` branch stored the literal string `set` as the `Authorization` header, so this quietly fixes a real bug. Row indices appear in messages (`webhooks.1.url: …`), so two rows failing the same rule read differently.

Three non-blocking observations. No correctness bugs found.

1. `normalizeWebhooks` is a sixth restatement of the rules this PR set out to collapse, and it is in the diff (the `WEBHOOK_EVENTS` cast on the `events` line). It independently re-implements the URL scheme regex, the 500-char cap, the event membership filter and the id regex — and it disagrees with the new schema in three ways: it truncates `authHeader` where the schema rejects, re-mints a malformed `id` where the schema rejects, and does not dedupe `events` where the schema does. Because it runs first at load, it silently repairs a hand-edited `settings.json` before the strict schema ever sees it, so the PR body's "a hand-edited `settings.json` carrying such values will now fail import loudly" holds only for backup restore through `update()`, not for the boot path.

https://github.com/perminder-klair/subwave/blob/0750334ac7c047b52396c75a059118c2ba6cd52a/controller/src/settings/normalize.ts#L361-L378

2. The comment explaining the `group: ['*', '!zod']` shape cites `task-2-report.md`, which is not in the repo. This is the config a future reader will open to understand why the rule is written that way, so the dangling pointer costs more here than usual.

https://github.com/perminder-klair/subwave/blob/0750334ac7c047b52396c75a059118c2ba6cd52a/controller/eslint.config.mjs#L27-L31

3. `stripZodImport` throws on a non-canonical *zod* import but copies every *other* import into the mirror verbatim. The module header promises failures land "HERE, at generate time, naming the source file — never downstream as a tsc error inside a generated file nobody is allowed to edit", and that promise currently only covers zod. Verified: a probe module under `src/schemas/` importing `express`, `node:fs` and `../settings/vocab.js` generated a mirror carrying all three imports with no generator error. The ESLint rule does catch it (I confirmed it flags relative, node-builtin and bare-package imports alike), so this is defence-in-depth rather than a hole — but rejecting any non-zod import in the generator too is a two-line change that would make the two guards agree.

https://github.com/perminder-klair/subwave/blob/0750334ac7c047b52396c75a059118c2ba6cd52a/controller/scripts/gen-schemas.ts#L154-L172

Checked and found nothing to flag: the `Pill` split (only `WebhooksPanel` and `ManualTagEditor` pass `onClick`; the pill nested inside a `<button>` in `ThemeSection` has no handler so it stays a `<span>`, no nested-button hydration risk), `mintId`'s output against the new `id` regex, `Field`'s `role="group"` and `FieldLabel`'s `asChild` forwarding behind `fieldAria`'s group variant, and the zod-range lockstep check.

perminder-klair:
All three findings above are fixed in `1f74609`, and the branch is now verified against a running station rather than only in unit tests.

**Fixes.** `normalizeWebhooks` runs the shared schema; id minting + collision de-duplication moved into `resolveWebhookIds`, shared with `mergeWebhookSecrets`, and `WEBHOOK_ID_RE` is exported so load and save test ids against one pattern. Its leniency is unchanged and now explicit. `gen-schemas.ts` rejects any non-zod module specifier at generate time, re-exports included, and cuts the zod import by node position rather than by matching lines. The eslint comment cites the actual reason instead of a missing file.

**Live verification** (dev stack from this branch, real Navidrome library, station on-air throughout):

- Route boundary, 7 rejection cases: each returns the flat `error` string *and* dotted `fieldErrors`, with the row index — `webhooks.1.url`, `webhooks.0.events.0`, `webhooks.0.authHeader`, `webhooks: At most 16 webhooks`.
- The secret path end to end: saved `Bearer REAL-SECRET`, re-saved with `authHeader: "set"` plus an edited URL, and confirmed `settings.json` on disk still holds the real value while the API only ever returns `set`. The listener-gate toggle saves independently without disturbing it.
- The load path, which is what the refactor rewrote: hand-edited `settings.json` to 10 malformed rows and restarted. 6 survived, 4 dropped, station stayed `on-air`. An unknown event name was filtered out leaving `dj.say` (a strict conversion would have deleted that hook), a malformed id was re-minted, a duplicate id was re-minted, `enabled: "yes"` fell back to `true`, a 600-char header clamped to exactly 500, and `Bearer KEEP-THIS` survived the full load → normalise → save cycle. Also checked byte-for-byte against the pre-refactor implementation over 25 load-path inputs before committing.
- Webhooks form in a browser, 19/19: chips are `<button type="button">` with `aria-pressed`, focusable, toggled by both Space and Enter; an invalid URL sets `aria-invalid` and an `aria-describedby` that resolves to the rendered message; a valid one drops the attribute entirely; Save is gated by the resolver; the stored secret renders blank and survives a UI save; no `_rhfKey` in the payload; no page errors.
- `ManualTagEditor`'s capped chip, 13/13 — the item this PR flagged as its weaker link, now driven with real focus and key events rather than read from markup. At the cap the fourth chip is still a `<button>`, carries `aria-disabled` with no `disabled` attribute, is focusable at `tabIndex=0`, refuses Enter without exceeding the cap, and is released the moment a selected chip is deselected with Space.

Follow-ups from this PR are tracked in #1337.

Issue #1286, opened 2026-08-03
No operator setting to keep the clock out of links and idents, the way #471 did for weather


mauriciopaim:
Feature request, not a bug report, and I am not asking to change the default.

## The gap

`SCRIPT_CONTEXT_FIELDS` includes `clock`:

```ts
const SCRIPT_CONTEXT_FIELDS = ['date', 'clock', 'time', 'festival', 'show', 'listeners'];
```

Seven generators read that list: `generateIntro`, `generateStationId`, `generateSignoff`, `generateHandoffGreeting`, `generateAdLib`, `generateLink` and `generateHourlyTime`. Only `generateLink` has any clock conditionality, via `clockIsAirTime`. The other five non-time-check generators receive a real "Local time" line unconditionally, and there is no setting anywhere that stops it. I grepped 1.3.0 for the obvious shapes (`speakClock`, `announceTime`, `timeChecks`, and similar) and there is nothing: the only clock-related settings are the IANA zone and the locale.

## Why this is worth a switch

The comment directly above that array already makes the argument, for a different field:

> Weather is deliberately EXCLUDED (issue #471): ambient weather stapled to every intro/link/ident/time-check made the DJ comically weather-heavy (~50% of all quips). Weather now reaches air only through the dedicated `weather` segment skill, which is cooldown- and change-gated.

That reasoning transfers to the clock unchanged. Once a concrete time is sitting in the context block, the model uses it, and the station opens on the hour repeatedly: "It's just past ten o'clock, and...". No persona wording or `djPrompt` steer reliably outweighs a concrete fact in the prompt, which is the same conclusion #471 reached about weather.

## This is a different axis from the accuracy work

The accuracy side is solved and I am not reopening it. #864 into #872, and later #1282, fixed **whether the spoken time is correct**: air-time resolution for links, the two-minute show-attribution padding, minute-aware hourly checks. None of that addresses **whether the operator wants a time-of-day station at all**.

Worth flagging that #864 was titled "Non-user instigated time call-outs off", but its body and the fix were both about the callouts running 10 to 20 minutes late. The off switch in that title never actually landed, so I do not think this is a duplicate. If you read it as one, say so and I will close this.

## Who this affects

Any format where the operator's wall clock is not part of the show: ambient, focus and study stations, archive and mixtape formats, and any station whose listeners are mostly outside the station's timezone, where the local clock is noise rather than texture.

## Ask

[...]

Issue #471, opened 2026-06-20
Enhancement: Make context quips be optional for each skill


hollowimage:
Pasting a quick dialogue from the discord for tracking:

id almost say, maybe each skill can have a tick-box for additional passed context. e.g. if you're talking about the commute conditions, weather is actually topical
so that when i customize a skill, i can tick boxes for: mention time, mention weather, mention X
this can give you a very neat way to develop tools vs developing contextual subtools
and also saves on tokens and context shuffling because you're no longer passing in as much every time into every tool call, and final product is just a json array building prior to submission.

genesis of this was overly-weather-heavy commentary, in my personal experience hitting easily 50% on all dj actions/quips, sometimes comically unnecessary.

perminder-klair:
Took a crack at this in #482.

You were right that weather was the obvious offender, but the per-skill toggle alone wouldn't have fixed most of it: the heaviest weather chatter actually comes from the generic between-track patter (intros, links, time checks), which aren't skills. Those had the weather stapled into every prompt, and the prompt was nudging the DJ to mention it on top of that.

So the PR does both:

- The generic patter no longer sees weather at all. It goes back to being the weather skill's job, which already has a cooldown and only speaks when conditions actually change.
- Every skill now has a `context:` line (`date, clock, time, weather, festival, show, listeners`) the tick-box idea you described. Default is everything except weather; tick it back on where it's topical (your commute-conditions example made it into the docs). There's a checkbox-per-field editor in /admin/skills too.

That should cut the weather mentions way down. Worth a try once it lands, and let me know if it's still too chatty about anything.

PR #1324, opened 2026-08-05
feat(dj): add a station switch to keep the clock off air

mauriciopaim:
## What

`settings.djSpeakClock: false` keeps the wall clock off air: no time of day in links, idents, hand-overs, ad-libs, banter or programme beats, and the automatic top-of-the-hour time check stands down. Default `true`, absent or non-boolean coerces to on, so an upgrade is byte-identical. Policy lives in one module, `broadcast/clock-policy.ts`, per your note on #1286.

## Why

Closes the gap #471 left. That issue gave weather a per-skill opt-out and its original request asked for "mention time" too, but the hardcoded generators never got it, so today the only ways off are `frequency: 'silent'` or `tts.enabled: false`, which take every other kind of talk with them.

One module rather than a filter at each call site because the clock reaches a model by four independent routes, and dropping `clock` from the context allowlists closes only the first: the "Local time" line; `generateStationId`'s "if you nod to the clock, keep it loose" nudge, which is pushed whatever the context block holds; the agent path's per-turn clause and `say` schema description, which never build context lines at all and are where most links come from with `llm.pickerAgent` on by default; and the hourly cron.

Refs #1286.

## How tested

- `npx tsc --noEmit` clean, `npx eslint .` 0 errors (613 warnings, all pre-existing).
- New `scripts/clock-policy.test.ts` passes, and **fails without the src changes** (stashed `controller/src`, first behavioural assertion trips). It pins the migration path, the tags surviving, the relabel, the hourly stand-down, that idents and banter are NOT gagged, and the agent-side schema rule.
- Full `npm test`: **4 failures with the change, the same 4 on a clean `develop` tree** (`analysis-failure`, `embedding-dim-migrate`, `observatory-scale`, `stem-backfill`). All four are `Could not locate the bindings file` for `better-sqlite3` under Node v26.4.0 here, so that is my environment rather than anything in this branch, and worth a second pair of eyes on a machine where they pass. 90 files before, 91 after, no new failures.
- Not yet run against a live station. I run the two-file hard-ban version of this in production, but that is a different shape from the setting, and the ident half in particular has not been on air here.

## Notes for reviewers

**One open naming question.** You suggested `settings.dj.speakClock`, and I have built `djSpeakClock` instead because there is no `dj` block in `DEFAULTS`: the DJ settings are flat (`djPrompt`, `djPrompts`, `djHouseRules`), and the only `stored.dj` read is `settings.ts`, migrating the legacy `dj.systemPrompt` forward. Say the word and I will move it, it is four lines.

Two deliberate calls worth your eye:

- **The gate sits in `dj-gate.shouldFire()`, not in `generateHourlyTime`.** Manual runners never call `shouldFire`, so the operator's "Time check" pad keeps working and still speaks the time, the same exemption `'silent'` and the voice switch already carry. Gating the generator would have produced a time check that refuses to say the time.
- **The daypart tags move rather than disappear.** `Local time: 21:14 · after dark · weekend` becomes `Vibe: after dark · weekend`. Per your note, and because `isDark` is what stops the DJ describing daylight after sunset.

`llm/internal/prompts/context.ts` gains its first import to read the predicate. That is option 1 from my #1286 comment; if you would rather keep that module dependency-free, the alternatives are still open.


mauriciopaim:
Heads up that nothing has actually been validated here yet: this is a fork PR, so `Lint` is parked at `action_required` waiting for a maintainer to release the run, and `check-runs` on `63897ad7` is empty. That is also what has `mergeStateStatus` on `BLOCKED` rather than any conflict.

Standing in for it until someone clicks approve, I ran what that job runs, `npm run lint` (`eslint . && tsc --noEmit`), in `controller`: 0 errors, 613 warnings, all of them pre-existing on `develop`. Every changed file is under `controller/`, so the `web` and `mcp-subwave` legs of the matrix are untouched.

One difference worth knowing: the workflow pins Node 22 and I ran on 26.4.0. It should not move an eslint or tsc result, but it is also the likeliest reason the four unrelated `better-sqlite3` binding failures I mentioned in the description show up here and probably will not on CI.

mauriciopaim:
Merged `develop` in (merge commit `56e4ed55`). The conflicts are cleared and the branch is mergeable again.

Two of the three were textual, both from the comment-trimming pass on `develop`: `settings/defaults.ts` and `broadcast/dj-agent.ts`. I took your shorter prose in both cases and re-applied the clock-policy paragraph and the `clockOff`/`airAt` lines on top of it.

The third is worth a look, because git did not flag it and it would have shipped the switch dead on the API path.

#1348 landed `settings/patch-registry.ts` after this branch was opened. `SETTINGS_PATCH_KEYS` now rejects any unlisted top-level key at the route, so adding a settings key takes a fourth edit beyond the three in `controller/CLAUDE.md`. `djSpeakClock` predates the registry and was missing from it, so `POST /settings` would have answered:

```
unknown settings key: djSpeakClock
```

while `update()` still carried a working handler for it. No admin panel posts this key today, so the practical blast radius is the API path plus any toggle added later. A hand-edited `settings.json` is unaffected, since `load()` coerces.

Your own test catches it. With the registry entry removed:

```
✖ every settings key with defaults is in the patch inventory
    actual: [ 'djSpeakClock' ], expected: []
```

I registered the key and converted it, rather than leaving it unconverted. "The SIX unconverted keys are the ones CLAUDE.md documents as resistant" is a claim about keys whose rules depend on state the patch does not carry, and a plain boolean is not one of those, so listing it there would have made that test say something untrue. `djSpeakClockSchema` is `z.boolean()` carrying the branch's own message, and deliberately not `settingsBoolLike()`: this key has always taken the strict `typeof` refusal, and `!!value` would loosen it, which is exactly the behaviour change `schemas/settings.ts` warns against smuggling in with a conversion.

The cost is that two inventory tests move and `web/lib/schemas.generated.ts` is regenerated. If you would rather the conversion were not part of this PR, say the word and I will split it out. Registering the key is not optional, but converting it instead of listing it as unconverted is a judgement call, and it is yours.

**No CI has run on this head:** zero check-runs and zero workflow runs on `56e4ed55`, which I believe is the usual gating on fork PRs. So nothing here has been validated by your pipeline. Locally, in `controller/`:

- `npm run lint`: 0 errors (634 pre-existing `no-explicit-any` warnings)
- `npx tsc --noEmit`: clean
- `npm run gen:schemas` then `git diff --exit-code ../web/lib/schemas.generated.ts`: clean
- `npm test`: 546/551

The 5 failures are `better-sqlite3` native bindings missing in my checkout (`airing`, `analysis-failure`, `embedding-dim-migrate`, `observatory-scale`, `stem-backfill`). They fail identically on a clean `develop` checkout, so they are not from this branch.

controller/src/broadcast/clock-policy.ts
```diff
@@ -0,0 +1,70 @@
...
+
+// May the AUTOMATIC top-of-the-hour time check fire? Same predicate, named for
+// the call site in broadcast/dj-gate.ts. Manual runners must NOT call this —
+// they bypass the gate entirely, which is what preserves the operator's pad.
+export function autoTimeCheckAllowed(): boolean {
+  return clockEnabled();
+}
+
+// Snapshot for the admin /debug surface, alongside voiceStatus().
+export function clockStatus() {
```
perminder-klair, line 68: This is never called. The comment says it's for `/debug` alongside `voiceStatus()`, but `routes/debug.ts` isn't in the diff. `voiceStatus()` is wired at `routes/debug.ts:241`; this sits next to it as one more line in that same object:

```ts
clock: (() => { try { return clockStatus(); } catch (err: any) { return { error: err.message }; } })(),
```

Either wire it or drop the export. A comment describing wiring that doesn't exist is worse than no comment, and this is the surface an operator hits when they're asking why the DJ went quiet about the time.

controller/src/settings/defaults.ts
```diff
@@ -161,6 +161,15 @@ export const DEFAULTS = {
...
   djHouseRules: '',
+  // Station clock switch. false = the wall clock stays off air: no time of day
+  // in links, idents, hand-overs, ad-libs, banter or programme beats, and the
+  // automatic top-of-the-hour time check stands down. Daypart colour survives
+  // ("after dark", "weekend", "late night", and the Period line) because that
+  // is atmosphere rather than a clock reading. Manual /dj/segment triggers stay
+  // exempt, so the operator's "Time check" pad still fires and still speaks the
+  // time: off means "stop doing this unprompted". Policy lives in exactly one
+  // place — broadcast/clock-policy.ts. Applies live; no restart.
+  djSpeakClock: true,
```
perminder-klair, line 172: There's no way to reach this from the admin UI. Its closest sibling, `djHouseRules`, ships a control in the System prompt modal on Personas. Right now an operator turns this off with a `POST /settings` or by hand-editing `settings.json`, which means most people will never find it.

A switch next to House rules would be the obvious home, since it's the same "how the DJ talks" question. Happy for that to be a follow-up PR rather than growing this one, but let's open the issue before merge so it doesn't sit as an API-only setting.

perminder-klair (commented):
Read through the whole thing. The design holds up, and I checked the two parts that could have bitten.

`dj-gate` placement is right. I traced `hourlyCheck()` in `scheduler.ts`: `rollSessionNow()` runs before the `shouldFire('hourly')` guard, so the session roll, persona handoff and programme intro are all unaffected. Only the time check stands down. Gating inside `generateHourlyTime` would have given us a time check that refuses to say the time, so this is the correct side of that line.

No import cycle from `prompts/context.ts` reaching into `broadcast/`. `settings.ts` doesn't import anything under `llm/internal/prompts`, so the new edge is safe. Worth knowing it's a one-way door: if `clock-policy.ts` ever imports anything under `llm/`, it closes the loop.

Agreed on all four of your judgement calls. Flat `djSpeakClock` over `dj.speakClock` (there's no `dj` block in `DEFAULTS`, and it sits with `djPrompt`/`djHouseRules`). Strict `z.boolean()` over `settingsBoolLike()`. Tags surviving under a moved heading. And converting the key in the registry rather than listing it unconverted, since a plain boolean isn't one of the state-dependent six.

Three things before this goes in, all wiring rather than design. Two are inline. The third is the root `CLAUDE.md`: `voice-policy.ts` and `dj-budget.ts` each get a "policy lives only in X" bullet, and this is a third module of the same shape with nothing written down. That file is what the next person reads before touching the prompt paths.

One process note: CI has never run on this head. I'll release the workflow before merging so we get the Node 22 leg. Your 5 local failures are `better-sqlite3` bindings and they reproduce on a clean `develop` here too, so I'm not worried about those.

mauriciopaim:
All three addressed. Pushed as `22f614c1`.

**1. `clockStatus()` is wired.** It sits next to `voice` in `routes/debug.ts`, same shape and same error guard, exactly as you sketched. You were right that the comment was the worse half: it described wiring that did not exist.

**2. Root `CLAUDE.md` has a "Station clock switch" bullet**, beside the `voice-policy.ts` and `dj-budget.ts` ones. It records the four routes the clock reaches a model by, the two deliberate exemptions (daypart colour, manual `/dj/segment`), and your one-way door note about `clock-policy.ts` never importing from `llm/`.

**3. The admin control is opened as #1380**, kept out of this PR as you suggested.

One thing that came out of writing #1380 and is worth flagging here: #471 already shipped a per-skill **Clock time** chip, and `djSpeakClock` silently overrides it. In `prompts/context.ts` the station switch sits inside the `on('clock')` branch, so a skill with Clock time ticked still drops to the daypart line when the station switch is off. That is the right precedence, but it is currently undocumented in either direction, so I noted it in the new bullet and built the proposed help copy around saying which wins.

#1380 proposes Settings › Voice rather than next to House rules, and gives the reasoning plus the places your suggestion is the better call. It is a proposal, not a disagreement I am attached to: say the word and I will build it under House rules instead.

Local checks after the merge and these two commits: lint `0 errors` (warning count 634 to 635, from the `err: any` in the line you gave me, matching `voice` and `budget` above it), `tsc --noEmit` clean, schema mirror clean, `npm test` 546/551 with the same 5 `better-sqlite3` binding failures you also see on a clean `develop`.

perminder-klair:
Reviewed the merge and the two follow-up commits — all three of my notes are addressed, and the #1348 resolution is right. `update()` goes through `parseSettingsPatchKey`, the key sits in branch order, and the regenerated mirror matches. Converting it rather than listing it unconverted was the correct call: a plain boolean isn't one of the state-dependent six, and strict `z.boolean()` keeps the existing refusal.

I checked the two things most likely to bite and both hold. `rollSessionNow` runs before the `shouldFire('hourly')` guard, so the session roll and handoff are untouched. And the manual Time check still speaks the time, because `generateHourlyTime` reads `context.clock.spokenTime` directly rather than off the context block.

I found one thing and pushed the fix to your branch as `8340cf73` — shout if you'd rather own it.

`runTrackEvent` nulls the clock stamp when the switch is off (`linkAirAt: airClock ? airAt : null`), but `runPickCycle` stamped `linkClockAt: airAt` unconditionally. With `djSpeakClock: false` the pool link is written under the flat ban, so it can't contain a time — but the stamp still let `linkClockDrifted` drop it at the seam when the real air moment landed more than 90s from the forecast. Turning the clock off then cost whole links, not just their times. Not a rare path either: it's what `llm.pickerAgent` off uses, and what `budgetMode: soft` forces. I gated the stamp rather than `airAt` itself, so `linkAirContext` keeps stepping the daypart tags to air time.

One thing left open, and I'm happy to merge without it. The test pins the context line, the gate and the schema, but not the ident nudge, `generateLink`'s clause, or the agent clause — the three routes the module exists for. The ident nudge is the one your own header calls out as the invented-time trap, and it's one prompt edit from regressing quietly.

Full suite is green here: 551/551, `tsc` clean, eslint 0 errors. Your 5 `better-sqlite3` failures don't reproduce on this box. Releasing the workflow now for the Node 22 leg.

perminder-klair:
Closed the test gap I flagged, pushed as `3a2550ea`.

My earlier fix left both pick paths deciding the same thing in two ternaries 170 lines apart, which is how they diverged to begin with. The decision now lives in `linkClockStampFor(airAt, clockOffered)` in `queue/pure.ts` and both sites call it. The rule it enforces is the one `linkClockDrifted`'s own comment already states: only an item that was actually given a clock carries a stamp.

`clockOffered` is a separate argument from `airAt` on purpose. A forecastable air moment isn't the same thing as a clock the model was handed, and the two paths withhold it for different reasons — the agent path also needs a clock in its context, and either can be overruled by the station switch.

Four cases in `link-clock.test.ts`, the last of which asserts the property end to end rather than as two separate facts: a withheld stamp survives the drift guard at any distance, while an offered one at the same distance is still dropped, so the first can't pass by the guard being toothless. I checked they actually bite by mutating the helper body to `airAt ?? null` — 2 of the 4 fail.

Local: 551/551, `tsc` clean, eslint 0 errors. CI Lint is green on this head across all three legs.

That leaves the two prompt clauses and the ident nudge still unpinned, which needs a different seam than this one and isn't worth holding the PR for. Happy with this as-is.

PR #1328, opened 2026-08-05
fix(broadcast): stop the station looping jingles when the music source starves (#1300 bug 7)

perminder-klair:
Closes the last confirmed listener-facing playback bug on the #1300 roundup.

**Draft: not yet smoke-tested on a live station.** Everything below is verified against the pinned `savonet/liquidsoap:v2.4.5` image and faithful full-chain probes, but never against real Navidrome. Given what that gap hid once already (see below), that test should happen before merge.

## The reported bug

When the music source starves — Navidrome unreachable, or `auto.m3u` empty/exhausted — the station plays jingles back to back forever, with no alert and no log escalation. It just sounds broken.

`rotate` skips unavailable sources, so with `music` starved the jingle rotate has no track boundary to return to. The `emergency` fallback below it never fires, because `radio` is still "available": it is producing jingles.

## What actually turned out to be wrong

The first fix gated the jingles source off after 30s of starvation, expecting the existing emergency fallback to take over. **It could not.** That fallback sat below three stacked `smooth_add` ducking layers, and the stdlib builds `smooth_add` as:

```liquidsoap
add(normalize=false, [normal, fallback(track_sensitive=false, [special, blank()])])
```

The inner fallback falls through to `blank()`, so it is always ready, and `add` is ready whenever any input is. The duck stack therefore reported ready even with a starved music chain, the bottom fallback always selected index 0, and **`emergency` has been unreachable dead code since the ducking layers landed**.

So the guard as first written replaced "stingers forever" with **silence forever** — measured at -91.0 dB, every sample at the floor. A strictly worse failure: a personal radio station is far more likely to notice a jingle loop than dead air.

Two positions had to change, each measured with position as the only variable:

| Guard position | Starved output | Result |
|---|---|---|
| Bottom of chain (as it lived for years) | **-91.0 dB** | digital silence |
| On `music_bus`, above the ducking | **-24.1 dB** | loop airs |
| Below the studio-bed `add`, `bed_enabled=true` | **-57.1 dB** | never selects `emergency` — the bed alone airs |
| On `music`, above the bed too (shipped) | **-9.0 dB** | loop airs in every configuration |

That third row is the nastier one: with the bed on, the mount keeps serving audio and the mixer *looks* alive while the dead-air loop never plays. It was one documented `bed_enabled = true` flip away.

## What's in here

**Mixer** (`liquidsoap/radio.liq`)
- Samples the pre-rotate music chain (`music_meta`) every 1s. After `STARVE_GRACE_SEC = 30.0` sustained, gates the jingles source off so the rotate empties. The window doubles as "a jingle may cover a brief hiccup" — sized off `SKIP_COMMIT_WAIT_MS = 20_000`, the measured bound on a request resolving.
- Moves the emergency guard onto `music`, above both the bed `add` and the ducking layers, and removes the dead bottom fallback rather than leaving a second decorative guard behind.
- ~1s fade-in on the emergency branch, deliberately **not** a debounce — a guard that delays firing is a fallback that doesn't fire, which is the bug being fixed. Needs two fades: `transitions=` does not apply to a switch's *first* selection, and cold boot is that selection.
- Writes `music-starved.json` (`{starved, since, at}`, unix seconds), heartbeat throttled to 10s against the reader's 60s staleness threshold. The write is inside `try/catch`: an uncaught exception in a `thread.run` callback kills the thread permanently, which would freeze the guard.

**Controller**
- `broadcast/music-starve-pure.ts` — pure marker logic, every ambiguous input resolving toward *not* starved. Pinned by `scripts/music-starve.test.ts`.
- `broadcast/music-starve.ts` — the IO shell, 2s sync memo.
- `musicStarved` / `musicStarvedSince` on `GET /state`; `music-starved.json` added to the station-switch IPC drain.

**Web** — `MusicStarvedBanner`, suppressed while the Navidrome banner is showing so one outage raises one alert.

**Docs** — the invariant is now recorded in `CLAUDE.md` and `liquidsoap/CLAUDE.md`: the dead-air guard must sit above every `add` in the music path, because `add` is ready if any input is. That rule being unwritten is what cost this.

## Verification

- `liquidsoap --check` clean against v2.4.5 (nothing in CI validates `radio.liq`).
- Full-chain probes including the ducking layers, measuring actual output level — the harness that missed the original defect omitted those layers, which is precisely why it missed it.
- Starve → latch → emergency → refill → automatic recovery in one tick, no restart.
- Heartbeat `at` deltas measured at 10.02s, not 1s.
- 91/91 controller test files pass; controller and web lint clean.

## Before merge

- [ ] On-air smoke test: confirm the emergency loop is audible on `/stream.mp3` at a real starve, that CPU is sane in the starved state (#660 removed `blank.skip` for a 100% spin in this exact "no music in the chain" condition), that recovery needs no restart, and that a slow cold boot doesn't flap the banner.

## Known and deferred

- The emergency fade registers callbacks on the long-lived `emergency` child that are not collected — one closure per switch, CPU only while the loop airs. Documented in-file; the static `fade.scale`+`mkfade` idiom is the escape hatch if it ever matters.
- The starve latch is suppressed while the idle gate is on, so the banner is dark when nobody is listening.
- `since` records when the latch fired, i.e. 30s after starvation began.

**Worth its own issue:** the emergency dead-air guard has been broken in production this whole time, independent of this change. That may explain #660's "`/stream.mp3` 404s when the chain has no music" rather than the emergency loop playing.


perminder-klair:
## On-air smoke test: passed

Run against a live dev stack built from this branch. The pre-merge checkbox is now closed.

**The bug reproduced live first**, which is worth recording — with the music source unreachable the station did exactly what the issue describes:

```
[jingles:3] Prepared "station_ident_default.wav" (RID 45)
[jingles:3] Prepared "station_ident_default.wav" (RID 51)
[jingles:3] Prepared "station_ident_default.wav" (RID 52)   ← the reported loop
[jingles:3] Prepared "station_ident_default.wav" (RID 58)
[jingles:3] Prepared "station_ident_default.wav" (RID 64)
```

Then the guard did its job:

| Check | Result |
|---|---|
| Latch | `MUSIC SOURCE STARVED for 30.0s - jingles gated off, emergency loop takes over` @ 12:59:33 |
| Jingle loop stopped | prepares froze at 14 and never grew again |
| **Emergency audible** | `/sounds/emergency.mp3` looping (RID 0); captured `/stream.mp3` at **-19.5 dB mean / -3.6 dB max** — not the -91 dB silence the pre-fix build produced |
| CPU while starved | **7.20%** — closes the #660 100%-spin concern for this newly-reached state |
| Marker heartbeat | `at` deltas of **10.02 / 10.00 / 10.01 / 10.01s** — the throttle holds on real hardware |
| Controller contract | `/state` → `musicStarved: true`, `musicStarvedSince: 12:59:33`, exactly the mixer's latch time |
| Recovery | `music source recovered - jingle rotate re-armed` @ 13:02:02, marker cleared, `/state` false, music back on air — **no mixer restart** |

That third row is the one that matters most: "the emergency loop actually airs" was inferred by every offline probe and is the assumption whose failure caused the Critical earlier in this PR. It is now measured on a real stream.

### Two behaviours confirmed real, both new

**Cold boot airs ~5s of the emergency loop** before the first track resolves (`music_safety → fade_in.2` at 12:54:14, → music at 12:54:19). Longer than the ~100-200ms an offline probe showed, because a real first fetch takes about five seconds. It fades in, so it reads as a swell rather than a stab — but every mixer restart now makes a sound where it previously made silence.

**Seam transients happen**, as the review predicted: the guard fired and returned within ~1s at 12:59:07 and 12:59:11. This is what the 1s fade-in exists for, and why a debounce would have been the wrong fix.

### Method note

Emptying `auto.m3u` does **not** starve the mixer — liquidsoap will not reload a playlist to empty, and keeps serving its existing list. The starve was induced by making the fetches themselves fail inside the broadcast container. Worth knowing for anyone reproducing this.

### Unrelated observation

`GET /now-playing` returned `(none)` while the mixer log showed a track playing. Observed **before** any starve was induced, so it is pre-existing and not from this branch — but it may be worth its own look.

Issue #1327, opened 2026-08-05
[BUG] OpenAI-compatible repeat penalty setting is ignored when the SDK supplies repeat_penalty: 1.0


Jaz666:
### Description
With the LLM provider set to OpenAI-compatible/llama.cpp and Repeat Penalty configured as 1.15, llama.cpp logs showed every request using `repeat_penalty = 1.000`.
`appliedRepeatPenalty(cfg)` correctly returns the configured value, but `openAICompatibleFetch()` only injects it when `body.repeat_penalty === undefined`:

```
if (penalty != null && body.repeat_penalty === undefined) {
  body.repeat_penalty = penalty;
}
```

The outgoing AI SDK request already contains `repeat_penalty: 1.0`, so the configured Subwave value is never applied.
Adding `--repeat-penalty 1.15` to the llama.cpp server command changes the effective sampler value to `1.150`, confirming the discrepancy.

### Expected behaviour
An explicitly configured Subwave repeat penalty should override the SDK’s default/no-op value.

### Possible fix
Within controller\src\llm\internal\provider\registry.ts
`if (penalty != null) {
  body.repeat_penalty = penalty;
}`
Or, more conservatively:
`if (
  penalty != null &&
  (body.repeat_penalty === undefined || body.repeat_penalty === 1.0)
) {
  body.repeat_penalty = penalty;
}`

perminder-klair:
Thanks for the detailed report, and for the `--repeat-penalty 1.15` datapoint. That's what confirmed the value wasn't reaching the sampler.

The guard turned out to be a red herring. `@ai-sdk/openai` has no `repeat_penalty` field at all, and its `providerOptions` schema drops one you try to pass, so that key is on the body only when we put it there. I drove a real `generateText` through the wrapper against a capturing fetch and the body came out with `repeat_penalty: 1.15`.

The value was getting lost earlier than that. `settings.load()` composes the `llm` block field by field and never listed `repeatPenalty`, so your 1.15 was validated, clamped, written to `settings.json`, and then dropped on the next restart. `appliedRepeatPenalty()` read undefined and injected nothing, so llama.cpp fell back to its own 1.0. Broken since the field landed in #918, on both the primary and fallback legs.

Fixed in #1329, which also keys the penalty into the model cache signature so an edit in the admin UI applies without a restart. Keep `--repeat-penalty` on your server command until it ships.

Issue #1317, opened 2026-08-04
[Feature Request] Allow more generation parameters for open-ai compatible TTS


CtznSniiips:
Currently I have a remote Chatterbox server set up for TTS under "Cloud provider > OpenAI-compatible". With this setup I can choose from a list of voices or specify a custom voice-ID. However I am missing the ability to specify other speech generation parameters like "Temperature" and "Seed".

PR #1336, opened 2026-08-05
Make the LLM layer more agentic: per-provider loop shape, per-file tools, authored prompts

perminder-klair:
Four independent changes to the `llm/` layer, one per commit, reviewable in order.

The starting observation was that the AI SDK usage felt inflexible. It isn't the SDK usage — `strategy/agent.ts` is careful work. The problem was that **the whole agent harness was calibrated to the weakest supported provider**, and that calibration was global rather than per-provider. These commits move the provider-specific parts into the place that already encodes provider differences, and clean up two structures that had grown past their shape.

### 1. `89593ec` — discovery budget becomes a capability

`COMMIT_AFTER_STEPS = 1` was a module constant: one discovery call, then `done` is forced. That ceiling is real for cloud Ollama (ignores `toolChoice` with several tools visible) and local GGUF models (emit schema-valid objects without exploring), but it was applied to Claude and GPT-5 too.

It's now `discoverySteps` in `ProviderCapabilities`, beside `objectStrategy`. Forced-tool providers keep 1 — **byte-for-byte unchanged**. Native-strategy providers get 3.

The step cap is *derived* (`gatedMaxStepsFor` = budget + 1) rather than taken from the caller, which preserves the invariant the GLM finding depends on: however tall discovery gets, the main run makes exactly **one** forced-`done` attempt before the recovery cascade. Extra `done` steps grow an "I already declined" trail and make compliance worse.

### 2. `4555169` — one file per picker tool, one scope object

`picker-tools.ts` was 712 lines: 17 tools in one closure over 13 lock params. The three parallel lists (call site → destructure → forward) were a real defect class, not clutter — a lock named in one and forgotten in another fell to a `null` default and silently stopped being enforced on the agent path while the pool picker still honoured it. That happened on #1300 FR 13 with `vocalLock`, and could only be caught by scraping source text for matching identifier names.

Constraints now travel as one `PickerScope` value that is never unpacked, so there are no two lists left to disagree. Tools are one file each, gating declared with the tool. `defineAgent` is generic so the hand-off is typechecked.

`picker-lock-forwarding.test.ts` is rewritten: instead of matching identifiers across files, it drives each lock through `collect()` and asserts it actually filters.

### 3. `f6aa261` — sharpen tool descriptions; **skip `outputSchema`**

The plan called for adding `outputSchema`. It doesn't earn its place: in ai@7.0.48 it is neither sent to the provider (the spec carries name/description/inputSchema/inputExamples/providerOptions/strict) nor validated on execute — the only reader is `validateUIMessages`, which this codebase doesn't use. It would buy a union schema per tool with no effect on the model. `src/mcp/tools.ts` keeps its output schemas, since there the schema *is* the published contract.

Descriptions do reach the model, so that half was done properly — targeting four observed failures (searchLibrary used for mood asks, empty `similarSongs` read as an empty library, off-vocab mood words, no-seed tools picked first-returned). Cost is +396 input tokens per pick, in the cacheable prefix; trimmed back from a first pass at +549.

### 4. `8dcb894` — prompt prose into `llm/instructions/*.md`

Static prompt blocks move to markdown; conditional assembly stays in TS. Source, not operator config — they ship in the image and the picker prompt is coupled to the harness. Loader validates at boot; a missing section or unsubstituted placeholder throws rather than shipping `{topic}` to a model.

Also fixes an inconsistency commit 1 introduced: the prompt still said "you get ONE discovery round" after native providers got three. The variant is selected by `promptDiscoverySteps()` — the **minimum** across primary and fallback, since the prompt is built before failover picks a leg and over-promising a second look is the expensive direction to be wrong.

## Verification

- `npx tsc --noEmit` clean; `npm run lint` 0 errors.
- Full suite: 95/96 files pass. The one failure is `analyzer-python.test.ts` (`vocal_gate_test.py`), **pre-existing** — it fails identically with these changes stashed, and is unrelated to the LLM layer.
- Prompt extraction verified lossless: `PICKER_CRITERIA` diffed byte-identical against `HEAD~`, and each moved block is pinned verbatim in `scripts/instructions.test.ts`.
- New tests: `picker-registry.test.ts`, `instructions.test.ts`; `llm-pure.test.ts` gains 5 budget cases; `picker-lock-forwarding.test.ts` rewritten.

## Behaviour change to be aware of

Stations on a **native-strategy provider** (openai, anthropic, google, deepseek, openrouter, requesty, gateway) will see picks make up to 3 discovery rounds instead of 1, with a correspondingly wider prompt. That is the intended change, and it costs more tokens per pick. Stations on ollama / openai-compatible / locca are unaffected on the loop shape.

https://claude.ai/code/session_01UwC4vpCzMRJUtib2NCthY5


perminder-klair:
### 5. `678e3a4` — the discovery budget is now operator-tunable

Follow-up on the review point above: the budget was capability-only, so an operator whose model didn't match its provider's general profile had no lever.

`settings.llm.discoverySteps` — per-leg, default `0` = follow the capability table, `1`–`5` overrides. Admin controls under **LLM → Agent** for both the primary and the backup leg.

**Why an override at all.** The descriptor keys off the *provider* and can't know which *model* it's serving, and the two failure directions are opposite: a capable model behind a self-hosted openai-compatible endpoint gets one cornered round it doesn't need (and if its seed tool returns empty, nothing to commit), while a cloud model that wanders across three rounds wants fewer.

**Per-leg**, same as `toolChoice`/`numCtx` — the backup may be a different provider running a different model. The override rides on the leg's cfg rather than being read from settings inside `capabilities.ts`, which keeps that module a pure function of its argument.

**Safety.** `0` is the sentinel, so an upgraded install is byte-identical to before the setting existed. Both the sentinel path and the clamp keep the resolved value in `[1, 5]` — a 0-round budget would force `done` at step 0 with an empty `seen` map, where the model can only fabricate an id. Raising it still buys no extra `done` attempts; the cap stays derived as `budget + 1`.

**Testing.** `scripts/llm-discovery-steps.test.ts` is a **cold-load** round trip, not an in-process check. `settings.load()`'s llm block composes explicitly and doesn't spread `DEFAULTS`, so a field missing there validates, saves, and works for the rest of the process before silently vanishing on the next restart — the bug that shipped twice (#1317, #1327) and that an in-process assertion passes on. I verified the test catches it: deleting the `load()` line fails 6 of its 9 cases.

Controller 96/97 test files pass (same pre-existing `analyzer-python.test.ts` failure); 0 lint errors in controller and web.

perminder-klair:
### Code review

Found 1 issue:

1. The per-provider discovery budget also widens the segment director's step cap, not just the picker's. `discoverySteps`/`gatedMaxSteps` are read off the leg with no scoping by agent kind, and both loop paths apply them over the caller's `maxSteps` — the native path via `stopWhen: [isStepCount(Math.max(maxSteps, gatedMaxSteps))]` and the gated path via `effectiveMaxSteps = useGatedDiscovery ? gatedMaxSteps : maxSteps`. On a native-strategy provider (`NATIVE_DISCOVERY_STEPS = 3` → `gatedMaxSteps = 4`), `directorAgent`'s `maxSteps: 2` is silently widened to 4, but that value is documented as load-bearing: "a taller budget just grows an increasingly 'I already declined' trail … and was the direct cause of a run burning the FULL agentTimeoutMs internally (45002ms observed)". The PR body scopes the widening to picks ("Stations on a native-strategy provider … will see picks make up to 3 discovery rounds"), so the director's cap changing looks unintentional — it likely needs kind-scoping or a director-side pin.

Budget computed per leg, kind-agnostic:
https://github.com/perminder-klair/subwave/blob/678e3a412d1f119e5c68a2aeddd188e878ca9990/controller/src/llm/internal/strategy/agent.ts#L250-L255

Native path overrides the caller's cap:
https://github.com/perminder-klair/subwave/blob/678e3a412d1f119e5c68a2aeddd188e878ca9990/controller/src/llm/internal/strategy/agent.ts#L311-L317

Gated done-tool path overrides it too:
https://github.com/perminder-klair/subwave/blob/678e3a412d1f119e5c68a2aeddd188e878ca9990/controller/src/llm/internal/strategy/agent.ts#L375-L380

The director's documented 2-step invariant:
https://github.com/perminder-klair/subwave/blob/678e3a412d1f119e5c68a2aeddd188e878ca9990/controller/src/skills/_agent.ts#L271-L280

perminder-klair:
Review issue addressed in e9950a72: the per-provider discovery budget is now opt-in per agent (`providerDiscoveryBudget`, resolved by `runDiscoverySteps`). The pick/request agents opt in; the segment director keeps its documented 2-step cap on every provider, pinned by new assertions in `llm-pure.test.ts` and `llm-discovery-steps.test.ts`. Also moved `schemas.ts`'s `instruction` import onto the `llm/dj.js` barrel and refreshed the stale `COMMIT_AFTER_STEPS` comment in `pick-seed.ts`.

PR #1350, opened 2026-08-07
feat: Test corrections button in Moods → Speech tab

drachenhort:
## What

Adds a Test Button to Moods Speech Correction Window

## Why

It should make it easier to test the Speech Correction in any Language by using more than Guesswork for the Phonetics.
I am German, we have quite a few Words where Syllabels are not spoken Out. This should help alot.

## Summary
- Adds a "Test corrections" control to admin → Moods → Speech: type a sentence, hear it synthesized through the station's default voice with the tab's current (including unsaved) correction rules applied.
- The Test-corrections window also gets a language dropdown for auditioning the sample sentence in another language — leave the test sentence blank and pick a language to hear a localized sample (reuses the sample-sentence lookup table other "Play sample" buttons already use elsewhere in admin).
- `POST /settings/tts/preview` gains an optional `corrections` override (sanitized via the existing `normalizeTtsCorrections`).

Note: an earlier revision of this PR scoped individual correction rules to a language (per-row tagging, filtered at speak time). That was reverted after review — corrections stay global/language-agnostic, as they always were; only the Test-corrections preview's sample-sentence language picker survived.

## Test plan
- [x] `cd controller && npm test` (full suite; only the 5 pre-existing unrelated failures, nothing new)
- [x] `cd controller && npm run lint`
- [x] `cd web && npm run lint`
- [x] Manual: server-side round trip proved the corrections override changes synthesized audio
- [x] Manual: settings save/load round-trips the plain `{from, to}` correction shape correctly
- [x] Manual: empty test text + a picked language plays a localized sample
- [x] Whole-branch code review before merge


drachenhort (commented):
tested lint,
otherwise tested using my own native language and some of the more special words we use around here.

perminder-klair:
Thanks for this, the core feature is solid and I'd merge it. A few things first.

**It needs a rebase on `develop`.** `web/components/admin/MoodsPanel.tsx` conflicts, and it won't be a textual fixup: #1359 moved the admin editors to react-hook-form, so the panel now holds `corrections` in a `useZodForm` field array with its own `correctionsSchema`, not the `useState` list your branch builds on. The `effectiveCorr` you pass to the preview button no longer exists in that shape. #1360 and #1365 landed on top of it since.

**Drop the two docs files.** `2026-08-08-corrections-language-scoping-design.md` and the matching plan describe the `{from, to, language}` shape and `correctionAppliesToLanguage`, both of which you reverted later in the branch. They're 1,549 of the 1,949 added lines, and sitting in `docs/superpowers/` they read as accepted design.

**Drop the language dropdown too.** It's the last surviving piece of the reverted work, and it doesn't test corrections: the 29 sample sentences are fixed strings, so none of them contains a word you've corrected. `synthesizeSample` also hardcodes `language: ''` into `speakWith`, and the card pins the engine to the station default, so picking Japanese hands a Japanese sentence to your English Piper voice. Without it the PR is about 100 lines and exactly its title, which also makes the rebase much smaller.

Smaller ones:

- `normalizeLanguage` is exported now only for the reverted predicate. Make it module-private again.
- `LanguageSelect.tsx`'s header comment still lists per-correction scoping as a call site.
- The new test file uses the old plain-script shape; CLAUDE.md asks for `node:test` on new files.
- The card hint prints the raw engine id, so operators see "(pocket-tts)".

What I liked: routing the preview through `normalizeTtsCorrections` instead of keeping a second sanitizer (d9a7861) is the right call, `corrections !== undefined` correctly separates "no override" from "override with an empty list", and reusing `effectiveCorr` means the preview auditions exactly what Save persists. `tts-corrections.test.ts` also adds the first coverage those two normalizers have ever had.

Ran the branch locally: both touched test files pass, controller lint is clean.

drachenhort:
Believe i changed it to everyones expectations now, let me know if theres anything i have missed

perminder-klair:
Thanks for sticking with this. Nice feature, and the corrections override path is clean.

I pushed the four follow-ups to your branch rather than send you round again: dropped the leftover `PREVIEW_LANGUAGES`, moved the new test to `node:test`, made the card hint show "PocketTTS" instead of the raw engine id, and pulled the engine→voice ladder into one shared helper (the copy in MoodsPanel had dropped `lang`).

Lints clean, tests pass. Merging.

Issue #156, opened 2026-05-26
CLI install issues


WhatKateDoes:
Trying the CLI install, unsuccessfully:

kate@azuracast:~$ curl -fsSL https://cli.getsubwave.com | sh    # installs, then offers to init + start
subwave setup                              # connect Navidrome + LLM
==> SUB/WAVE CLI installer
  target:  linux-x64
  version: v0.1.24
  source:  https://github.com/perminder-klair/subwave/releases/download/v0.1.24/subwave-linux-x64
  dest:    /usr/local/bin/subwave
  (elevating with sudo to write to /usr/local/bin)
[sudo] password for kate: 

✓ installed subwave to /usr/local/bin/subwave

Run `subwave init` now to scaffold the install? [Y/n] 

  ███████╗██╗   ██╗██████╗     ██╗██╗    ██╗ █████╗ ██╗   ██╗███████╗
  ██╔════╝██║   ██║██╔══██╗   ██╔╝██║    ██║██╔══██╗██║   ██║██╔════╝
  ███████╗██║   ██║██████╔╝  ██╔╝ ██║ █╗ ██║███████║██║   ██║█████╗  
  ╚════██║██║   ██║██╔══██╗ ██╔╝  ██║███╗██║██╔══██║╚██╗ ██╔╝██╔══╝  
  ███████║╚██████╔╝██████╔╝██╔╝   ╚███╔███╔╝██║  ██║ ╚████╔╝ ███████╗
  ╚══════╝ ╚═════╝ ╚═════╝ ╚═╝     ╚══╝╚══╝ ╚═╝  ╚═╝  ╚═══╝  ╚══════╝
  install

  · Scaffolds a fresh install directory, writes the compose file + .env, and records the home so future commands know where to look.
  After this, run `subwave start` then `subwave setup` to finish configuration.

│
◇  Install directory
│  /home/kate/subwave
│
◇  Deployment shape
│  prod — bundled Caddy on :7700
│
◇  Admin username (gates /admin and /onboarding)
│  admin
│
◇  Admin password (leave blank to generate a random one)
│  ▪▪▪▪▪▪▪▪▪▪▪▪▪▪
│
◇  Public site URL (https://radio.example.com — blank to defer)
│  https://subwave.drearburh.uk

━━ Scaffolding install ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
  ● created /home/kate/subwave/ (state/, state/logs/)
[...]

PR #1357, opened 2026-08-08
refactor(cli): trim comments to load-bearing rationale + add README

perminder-klair:
## What

`cli/` carried ~1130 comment lines over ~5200 lines of source (22%). A lot of it restated the code directly beneath it. This cuts that down to ~525 lines, touching comments only.

**Removed:**
- File headers narrating the flow step by step — `setup.ts` opened with 41 lines including a numbered 12-step list that the function below already reads as, and whose "persistence layers" section was stale after the `/onboarding/save` refactor.
- Numbered step banners (`// --- 4b. Heavy TTS sidecar ---`) sitting over one self-describing call, usually duplicating the doc comment on the function being called.
- ASCII section dividers (`// --- individual sections ---`).
- Doc comments restating a signature (`// Read the pid file (best-effort). Returns 0 if absent`).
- One comment that was literal thinking-out-loud: *"so let's check whether headers were available... we can't, so just report fail"*.
- Speculative dead weight: `// Future fields: lastUpdatedAt, cliVersionAtInit, telemetryOptIn`.

**Kept, compressed:** the rationale that isn't recoverable from the code — the Bun/macOS stdin bug and why `/dev/tty` is opened by hand, Compose's `$VAR` interpolation trap behind `envEscape` (#156), the loopback→`host.docker.internal` swap, the version-pin drift story, the `smooth_add`/burst-size style invariants, and every issue reference.

**Two stale comments fixed:**
- `docker.ts` said `composeExec` was "not used in the v1 doctor" — `doctor.ts` calls it for the `radio.log` fallback.
- `setup.ts`'s header described writing `setup-config.json` + `secrets.env` + `POST /settings` from the host, which is the exact flow the file's own body says was replaced.

## Code changes

Two, both behaviour-identical:
- `writeEnvFileAt` was a pure delegating wrapper whose stated purpose was *"kept as a separate function so future init-only quirks have somewhere obvious to land"*. Removing the comment would leave an unexplained no-op function, so its one call site now calls `writeEnvFile` directly.
- One single-statement `if` lost its braces.

`git diff` filtered to non-comment lines is those two changes plus trailing-comment moves — nothing else.

## Verification

- `npm run typecheck` clean.
- `subwave --version` / `--help` run.
- Exercised the one changed path: `init --yes` into a throwaway home writes `.env` with the template's comments and key order intact (`ADMIN_USER` line 15, `TZ` line 252) and the `SUBWAVE_VERSION` pin replacing the commented example line — same as before.
- `embed-assets` regenerates `assets.generated.ts` with no drift, so the CI verify job stays green.


perminder-klair:
Added `cli/README.md` (the package had none) in 0b16c3b0.

Operator half is the install flow, a command table, and the `SUBWAVE_HOME` resolution cascade. Contributor half documents the things that bite and aren't visible from any single file:

- **Embedded assets** — regenerate after any root compose / `.env.example` edit, because `verify-cli-assets.yml` gates on it.
- **CI does not typecheck `cli/`** — `lint.yml` covers `controller/`, `web/` and `mcp-subwave/` only, so `npm run typecheck` is on you before pushing.
- **Prompts and piped stdin** — why `tty.ts`, `patch-clack.mjs` and the `ui.ts` watchdog exist, framed as one mechanism rather than three optional pieces, since removing any one re-opens the macOS hang.
- **Version pinning** — the pin's purpose and the `x-release-please-version` marker that must stay on the version line.

Two claims I drafted from the code comments turned out to be wrong and were corrected against the source before committing: the installer defaults to `/usr/local/bin` with sudo escalation (`~/.local/bin` is only via explicit `--dir`, not an automatic fallback), and typecheck is not a CI gate.

Issue #1348, opened 2026-08-07
Design: /settings mega-endpoint on the shared zod form schema (#1337)


perminder-klair:
Split out of #1337, which converted every single-feature admin form onto the shared zod schemas (webhooks, stations, shows, skills, playlists, imaging, onboarding probes, the public request boxes and — in the same PR — the weekly schedule). `/settings` is the one surface left, and it was deliberately held back because it needs a design rather than a straight conversion.

## Why the six-step recipe doesn't apply

The recipe on #1337 assumes a route that owns one shape. `/settings` owns forty-four.

`controller/src/routes/settings.ts` is a 22-line mount table. The handler is `controller/src/routes/settings/core.ts:218`, and there is **no middleware at all** between `requireAdmin` and the body:

```ts
router.post('/settings', requireAdmin, async (req, res) => {
  try {
    const result = await settings.update(req.body || {});
    …
  } catch (err) {
    res.status(400).json({ error: err.message });
  }
});
```

All validation lives inside `settings.update()` (`controller/src/settings.ts:1023`), a ~1,100-line chain of **44 `if ('<key>' in patch)` branches**: `jingleRatio`, `crossfadeDuration`, `maxTrackSeconds`, `archive`, `stream`, `loudness`, `weather`, `station`, `timezone`, `locale`, `theme`, `moods`, `moodSchedule`, `weatherMoods`, `festivals`, `djPrompts`, `activeDjPromptId`, `djPrompt`, `djHouseRules`, `personas`, `shows`, `schedule`, `scheduleOverride`, `activePersonaId`, `tts`, `llm`, `search`, `embedding`, `skills`, `audio`, `transitions`, `sfx`, `beds`, `ui`, `privacy`, `requests`, `webhooks`, `webhooksPolicy`, `scrobble`, `likes`.

Four properties of that shape are what make this a design problem:

1. **The body is a partial patch, not an object.** `z.object` strips unknown keys. A schema over the whole settings object would silently delete whatever a form learns to send next — the exact failure the `skills` open-tail note and the `onboarding` pass-through note on #1337 both document. `schemas/onboarding.ts:14-16` states it directly:

   > Most of `POST /onboarding/save` forwards partial llm / tts / personas / weather patches to `settings.update()`, and a schema in front of that is the `/settings` mega-endpoint problem in miniature — worse than nothing, because `z.object` would strip whatever the wizard learns to send next.

[...]

Issue #1337, opened 2026-08-05
Follow-up tracker: remaining form conversions from the zod-forms foundation (#1323)


perminder-klair:
#1323 landed the zod + react-hook-form foundation and proved it end to end on Webhooks. The mechanism is in place; what remains is applying it to the other admin forms. Filing this so the migration order and the one open design question don't live only in a merged PR description.

## What the foundation gives a converting form

- A shared schema in `controller/src/schemas/<feature>.ts`, mirrored into `web/lib/schemas.generated.ts` by `npm run gen:schemas` and drift-checked in CI. Never hand-edit the mirror.
- Rules that need server state (redaction sentinels, id minting, cross-item awareness) go in a non-mirrored `<feature>-server.ts` sibling.
- `middleware/validate.ts`'s `validateBody()` at the route boundary, returning `{ error, fieldErrors }` where `fieldErrors` is keyed by dotted path and maps straight onto inputs.
- `settings.update()` stays the authoritative chokepoint — `validate*Strict` keeps its name and signature and calls the same schema.
- The lenient load path (`settings/normalize.ts`) runs the same schema too, so load and save can't restate a rule differently. It keeps only its leniency: repair or drop, never throw.
- `lib/form.ts`'s `useZodForm`, the vendored shadcn `Field` primitives, `fieldAria()` for ARIA, and `<Pill onClick pressed>` for chip multi-selects.

## Converting one form

1. Add `controller/src/schemas/<feature>.ts` (zod-only imports — enforced by eslint and by `gen-schemas.ts`).
2. Move stateful rules to `<feature>-server.ts`.
3. Point `validate<Feature>Strict` and `normalize<Feature>` at the schema.
4. `cd controller && npm run gen:schemas`.
5. Add `validateBody()` to the route.
6. Convert the panel to `useZodForm` + `Field` + `fieldAria`.

## Remaining conversions, cheapest first

All six landed conversions ship in **PR #1346** (`feat/zod-forms-stations`). The last three were lost uncommitted before ever reaching the branch and were **redone from the notes below** (see the status-correction comment) — the notes are now the record of a spec that has been executed twice.

- [x] `stations` — done, PR #1346
[...]

PR #1358, opened 2026-08-08
feat: finish the zod form conversion — personas, blocklist, library, and the fieldErrors channel (#1337)

perminder-klair:
Picks up the #1337 tracker, which was closed with its checklist ticked but had only ever covered the *shared-schema* half. This does the rest of that half and the browser half it deferred.

## Where coverage stood

Two axes, and only one was close to done:

- **shared schema** — 26/42 `/settings` keys; `personas`, the blocklist, and manual tagging had no schema at all, and three of those routes had **no validation middleware whatsoever**.
- **browser** — `POST /settings` had been answering with `fieldErrors` since #1348 and *nothing consumed it*. `lib/notify`'s own header says a toast is the wrong surface for field-level validation; a toast is what every one of those failures got.

## What this does

**Three new schema modules.** `schemas/persona.ts` (+ `persona-server.ts`) covers the roster and the DJ prompt library, and absorbs the `{engine, voice, cloudProvider}` **voice slot** that the station-wide rescue slot shares with every persona — so the per-engine voice rules live once. `schemas/blocklist.ts` and `schemas/library.ts` cover the never-play rules, the id-entry body, and the manual tagger. `settings/vocab.ts` now aliases 16 constants rather than declaring them.

**`/settings` registry: 26 → 36 keys.** Each new one keeps whatever half is genuinely stateful in `update()` — id minting, the async theme-exists fallback, the crossfade-derived floor, membership in a roster the same patch may be replacing. The six that stay unconverted are now pinned by a test naming each reason, so "why isn't `llm` done" has an answer in code.

**The `fieldErrors` channel actually lands on inputs.** `saveSettings` keeps the payload, scoped by top-level key so a fixed field clears while an unrelated section's error survives. Ten controls render inline; the five bulk-save sections get it through `SaveBar` in one edit.

## One real bug, found by the conversion

`personaValid()` was a hand-written reimplementation of the whole persona validator, and it had drifted:

```js
if (e === 'cloud') return v.length >= 1 && v.length <= 100;
```

The controller has always allowed an **empty** voice when the provider is `openai-compatible` — a self-hosted server picks its own default and there is no id to type. That predicate gates the Save button *and* paints the roster's validity dot, so an operator on such a server saw their persona marked invalid and **could not save the roster at all**, with nothing naming the field.

Verified by running the old predicate and the new one over the affected persona: refused before, accepted now, while plain `openai` with an empty voice is still refused.

A sweep for the same pattern — a bare literal beside a "keep in sync with the controller" comment — turned up three more (bitrate vocabularies, mood caps, festival bounds). None had drifted yet. That is the argument for doing them: the persona one had, and nothing was watching either.

## Notes for review

- **`validateBody` gains `messages: 'verbatim'`.** These schemas' messages already name their own field (`rule.label is required`), so the default prefixing would answer `label: rule.label is required` — and worse, the route would disagree with the store's own chokepoint about the same body. Same narrow exception the settings registry already documents; `fieldErrors` carries the dotted path either way.
- **`SETTINGS_PATCH_ROOTED_KEYS`.** An array key's messages are not self-locating, so the route must root them exactly as the strict validator does, or the two name different rows for the same body.
- **Zero behaviour change was the contract**, and most of the work was preserving accidental leniency: `String()` coercion on name/soul/tagline, an absent key reading as a default (zod 4 treats a bare `z.unknown()` as **required**), a malformed `skills` entry DROPPED where a malformed `language` is REFUSED, and a numeric skill slug surviving because `String(42)` is a legal slug.

## Deliberately not done

The big editors (`ShowEditor`, `SkillEditModal`, `PlaylistBuilderPanel`, imaging) keep their own state machines and already run the schema for their save gates — the call #1337 made for each of them. Binding them to react-hook-form is a rewrite that buys no new correctness guarantee, and belongs in its own PR.

## Testing

`controller`: lint clean (0 errors), **115/115 test files pass**. `web`: lint clean, `tsc` clean.

New: `persona-schema.test.ts` (41), `library-schema.test.ts` (21), `validate-wire-shape.test.ts` (8), plus additions to `settings-patch-schema.test.ts` (73 total).

`validate-wire-shape.test.ts` is worth a look — it asserts that the dotted paths hard-coded in `SettingsPanel`'s JSX are the paths the controller really sends. Without it, dropping or re-rooting `fieldErrors` would leave every inline error silently blank and read as "validation stopped working".

**Still to do before this leaves draft:** browser verification that the inline errors render where intended. The payload contract is covered by CI; the rendering is not.

### DIFF
--- a/controller/scripts/library-schema.test.ts
+++ b/controller/scripts/library-schema.test.ts
@@ -1,0 +1,305 @@
+// The never-play blocklist and the manual tagger moved onto shared zod schemas
+// (controller/src/schemas/{blocklist,library}.ts), mirrored into
+// web/lib/schemas.generated.ts. These tests pin what the conversion promised:
+// the STORE's chokepoint (blocklist-rules.validateRulePatch) and the ROUTE
+// boundary now run one implementation, and every message stayed byte-identical.
+//
+// The message half is the sharp end here. Both schemas' messages already name
+// their own field ('rule.label is required'), so their routes mount
+// validateBody's `messages: 'verbatim'` posture — without it the operator reads
+// 'label: rule.label is required'. scripts/blocklist-rules.test.ts keeps the
+// MATCHING half; this file covers validation only.
+//
+// Run: npx tsx scripts/library-schema.test.ts (auto-discovered by npm test).
+import assert from 'node:assert/strict';
+import { mkdtempSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import test from 'node:test';
+
+process.env.STATE_DIR = mkdtempSync(join(tmpdir(), 'subwave-library-schema-'));
+
+const {
+  BLOCK_TYPES,
+  RULE_FIELDS,
+  RULE_TEXT_MAX,
+  RULE_VALUES_MAX,
+  blockEntrySchema,
+  blockRuleSchema,
+  normText,
+} = await import('../src/schemas/blocklist.js');
+const {
+  MANUAL_TAG_ENERGIES,
+  MANUAL_TAG_MOODS_MAX,
+  manualTagSchema,
+} = await import('../src/schemas/library.js');
+const rules = await import('../src/music/blocklist-rules.js');
+const { validateBody } = await import('../src/middleware/validate.js');
+
+const rule = (o: Record<string, unknown> = {}) => ({
+  label: 'Christmas songs',
+  field: 'genre',
+  values: ['Christmas'],
+  ...o,
+});
+
+// --- the rule schema --------------------------------------------------------
+
+test('a minimal rule parses and defaults its optional halves', () => {
+  const r = blockRuleSchema.parse(rule());
+  assert.equal(r.label, 'Christmas songs');
+  assert.equal(r.field, 'genre');
+  assert.deepEqual(r.values, ['Christmas']);
+  assert.equal(r.season, null, 'no season = always blocking');
+  assert.deepEqual(r.showIds, [], 'no scope = station-wide');
+});
+
+test('label is required, trimmed and capped', () => {
+  assert.equal(blockRuleSchema.parse(rule({ label: '  Xmas  ' })).label, 'Xmas');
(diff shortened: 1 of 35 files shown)


perminder-klair:
## Browser-verified

Ran against an isolated controller (`:7791`, temp `STATE_DIR`) + the worktree web dev server (`:7793`). The operator's real station was never touched.

**The payloads, from the running server:**

```
POST /settings {"crossfadeDuration":999}
  {"error":"crossfadeDuration must be number in [0, 30]",
   "fieldErrors":{"crossfadeDuration":"crossfadeDuration must be number in [0, 30]"}}

POST /settings {"stream":{"bitrate":7,"opusBitrate":1}}
  fieldErrors: {"stream.opusBitrate":"…one of: 96, 128, 192, 256, 320",
                "stream.bitrate":"…one of: 64, 96, 128, 160, 192, 320"}

POST /settings {"personas":[{"name":"", …}]}
  {"error":"personas.0.name: name must be 1-40 chars",
   "fieldErrors":{"personas.0.name":"name must be 1-40 chars"}}

POST /library/blocklist/rules {"label":"", …}
  {"error":"rule.label is required", "fieldErrors":{"label":"rule.label is required"}}
```

Note the last one: `error` is the message **verbatim**, not `label: rule.label is required`, while `fieldErrors` still carries the path. That is the `messages: 'verbatim'` posture doing its job.

**The persona bug, end to end:** `openai-compatible` + empty voice now returns **200** (the exact case the UI was blocking), while plain `openai` + empty voice still returns 400 — so the fix removed the drift without loosening the rule underneath it.

**The rendering,** driven with Playwright — 4/4:

- the message is in the DOM as `role="alert"`
- it sits **inside the crossfade field, next to the input** — not only in a toast
- it **clears** once the value is fixed and saved

On screen: `999` in the input, with `crossfadeDuration must be number in [0, 30]` in red directly beneath it, above the existing hint line, and the toast still firing.

(Those UI checks were a throwaway harness, not committed — the durable half is `controller/scripts/validate-wire-shape.test.ts`, which pins the dotted paths the panel's JSX hard-codes.)

PR #1366, opened 2026-08-10
chore(controller): replace five hand-rolled parsers/subsystems with libraries

perminder-klair:
Each swap replaces code whose failure mode was **silent**, in a place that takes **operator input**. No behaviour change is intended anywhere else.

## RSS parsing → `fast-xml-parser`

`skills/news.ts` matched shallow RSS 2.0 `<item>` blocks with three regexes. An Atom feed (`<entry>`/`<summary>`), an RDF/RSS-1.0 feed (items at the document root, no `<channel>`), a namespaced tag or a nested CDATA section all returned **zero items** — and the feed URL is whatever an operator types into the news skill's `feed:` field. The failure presents only as the news beat quietly never airing.

`removeNSPrefix` collapses the dialects onto plain tag names; `parseTagValue: false` keeps a headline of `2026` a string. `parseFeed(xml, cap)` is split out as a pure, network-free seam.

Verified against three live feeds — BBC (RSS 2.0), Hacker News, github.blog (WordPress, `content:encoded`) — plus `scripts/news-feed.test.ts`.

## SKILL.md frontmatter → `yaml`

The flat `key: value` line splitter mangled most of what an operator would reasonably write: a list parsed as an empty value and dropped every entry, an inline `# comment` survived into the value, and a block scalar became the literal text `|-`.

That last one is not hypothetical — our own `docs/custom-skills.md` example writes `cooldown: 6h   # hard min gap…`, the comment stayed in the value, `parseCooldownMs`'s `^(\d+)\s*([smhd]?)$` failed, and it silently fell back to 60m. The docs have said "YAML frontmatter" for a while.

Two contracts hold it together:

- **`data` stays `Record<string, string>`.** The tool's 4th `config` arg, `config-fields` coercion, `preservedFrontmatter` and `parseTags` are all built on strings. Lists flatten to the comma-joined form those already accept, so `tags: [a, b]` and `tags: a, b` now mean the same thing. Nested maps have no consumer and are dropped.
- **A block YAML refuses falls back to the legacy parser** and reports `malformed` so `loadSkillDir` warns. The sharp case is an unquoted colon (`label: News: Today`) — the old parser accepted it, so it is on operators' disks today, and a skill must not vanish at boot over it.

The writer is the other half of one round trip: `yamlScalar` quotes a value **only when it would not survive being emitted bare**, and tests that by actually round-tripping it rather than by a dangerous-character list. `6` needs no quotes (it parses as a number and the loader flattens it straight back to `"6"`); `007` does. Being minimal matters — this rewrites files operators hand-edit, and over-quoting would churn every SKILL.md on the first save after the upgrade.

All 18 real `SKILL.md` files (state + builtins) parse to byte-identical data.

## Env config → zod (`src/util/env.ts`)

Every numeric var was read as `parseInt(process.env.X || '<default>', 10)`, which yields **NaN** for anything non-numeric, and the NaN then travels: `NEWS_MAX_ITEMS=ten` removed the headline cap entirely, a stray character in `PORT` bound a random port, a bad `PIPER_SPEED` reached ffmpeg as `atempo=NaN`. Every one of those is silent at the point of the mistake.

Posture is **warn and fall back, never throw** — the same reasoning as the state bootstrap in `docker/broadcast-entrypoint.sh`. A station that refuses to boot over one malformed convenience var is strictly worse than one that boots on the default and says so. Absent and *empty* both mean "not set", so `ANALYZE_URL=` stays an ordinary compose line. Issues collect in `envIssues()` and `server.ts` repeats them into the **booth log**, since container stdout is not a surface operators read.

`envUrl` parses inside its own predicate rather than chaining `z.url().refine(…)`: zod runs every check even after one fails, so the refine saw unparseable input and `new URL()` threw straight out of the reader — breaking the exact never-throw contract the module exists to hold. The test caught that.

## Test runner → `node:test`

`scripts/run-tests.ts` keeps its CLI (`npm test`, `npm test -- <filter>`) and its discovery, and now hands the file list to `node --import tsx --test --test-concurrency=1 --test-reporter=spec`.

**No existing test file changed.** Two shapes coexist deliberately: the original plain-script shape reports as one pass/fail keyed on its exit code — byte-for-byte the old contract — while a file using `node:test`'s `test()` reports per assertion, with the failing one named. New tests should use `node:test`. Concurrency is pinned to 1 because files here reach for shared ground (a temp state dir, the library DB, env vars).

## `helmet`

Applied above every route, with three deliberate overrides:

- `crossOriginResourcePolicy: 'cross-origin'` — **required**. Helmet's `same-origin` default tells the browser to refuse the response to any other origin, which would blank `/cover/:id` artwork, persona avatars and the sfx/jingle/bed previews in every deployment where the player is not same-origin with the controller. That is exactly what `middleware/cors.ts`'s wide-open `*` exists to allow.
- CSP off — no document is served from here, so a CSP on JSON or an image is inert; the web app ships its own.
- HSTS off — TLS is not this process's to assert. Cloudflare/Caddy terminate it, and helmet's default carries `includeSubDomains`.

What is left is what an API wants: `nosniff`, no `X-Powered-By`, `X-Frame-Options: SAMEORIGIN`, `Referrer-Policy: no-referrer`.

## Verification

- Full suite **536/536 passing**; `eslint` + `tsc --noEmit` clean (0 errors; the 633 warnings are pre-existing `no-explicit-any`).
- Header set checked against a **booted controller**, not just a harness.
- `package-lock.json` regenerated under node 22 in a temp dir and confirmed with `npm ci --dry-run` in a clean dir, so CI's install step won't reject it.

New tests: `news-feed.test.ts`, `skill-frontmatter.test.ts`, `env-config.test.ts`, plus a round-trip block in `skill-scaffold-preserve.test.ts`.

## Note for whoever runs the suite next

`observatory-scale.test.ts` builds a 200k-track SQLite DB under `os.tmpdir()`, which is a 12 GB tmpfs on a default Arch box — it dies with `SQLITE_IOERR_WRITE` when that fills. Running with `TMPDIR` pointed at real disk passes. Documented in the root `CLAUDE.md`; worth fixing properly in its own PR so the test targets real disk itself.

## Not in scope

`pino` for the 235 `console.*` calls and `kysely` over the raw `better-sqlite3` queries were both discussed and deliberately deferred — each is a large enough diff to deserve its own PR.


perminder-klair:
Pushed a fifth swap of the same shape: **`state/secrets.env` → `dotenv`, through one parse** (`fd120b14`).

This one is about *credentials*, and it failed the same way the others did — silently, on operator input.

**What was wrong.** The file holds every cloud API key and is deliberately hand-editable (the module header invites it). The splitter understood `KEY=value` plus one pair of surrounding quotes, so ordinary `.env` syntax produced the **wrong secret**:

- `export OPENAI_API_KEY=sk-…` — everything before the first `=` was taken as the key, matched nothing in the allow-list, dropped entirely.
- `OPENAI_API_KEY=sk-… # personal account` — the comment became part of the key.

Neither points anywhere near this file; both surface as a provider 401.

**The parse also existed twice** — once on the boot path, once inside `saveSecrets`' merge. Those copies have to agree *by construction*, because `saveSecrets` reads then rewrites the whole file: a value the merge misreads is a stored secret **destroyed on disk**, not merely one that fails to load. There is now one reader.

**Nothing about the on-disk format changes.** `envEscape` is untouched — it single-quotes anything outside a conservative safe set precisely so the reader gets it back verbatim, and refuses a newline rather than corrupting the file. No existing `secrets.env` needs rewriting, and no shell script reads this file (only `icecast-secrets.env` is `source`d, and that is a different file and module).

**One honest trade, handled rather than hidden.** dotenv truncates an unquoted `#` — correct `.env` semantics, and what makes the inline-comment fix work, but it is the one place this swap is *less* forgiving than the splitter it replaces. Going quiet about it would just trade one silent wrong secret for another, so the case is detected and the warning names the fix. The writer never emits that shape. Same treatment for a multi-line quoted value, which `envEscape` cannot persist and which would otherwise make the *next* save throw on a key the operator never touched.

Warnings reach both stdout and the booth log, like the env-var warnings.

Suite is now **550 passing**. The load-bearing new pin is the save→read round trip over awkward secrets (spaces, `#`, `=`, `&`, empty).

---

For the record, on what a sweep for more of these turned up: not much. Most hand-rolled parsing in the controller reads files *we wrote* (`recent-plays.json`, `events-*.jsonl`, `auto.m3u`, the `liquidsoap_*.txt` markers) — both sides ours, so the format can't surprise us. The only other genuine instance is `parseListClients` (`broadcast/listeners.ts`), a regex over Icecast's admin XML; small blast radius, since the listener *count* comes from `status-json.xsl` and a miss only empties the admin Listeners table. Left alone deliberately.

PR #1367, opened 2026-08-10
fix(tts): distinguish cloud fallback providers

### LINKED ISSUE #1345: [BUG] Fallback TTS Engine choice not being respected
on v1.6.0, with fall back set to Cloud > Elevenlabs, using a known working voice, which passes the Play Sample button test.
I've confirmed the new settings are persisting in settings.json

Failed TTS calls are still falling back to Piper :(

Jaz666:
## Summary

- distinguish Cloud TTS rescue targets by provider, not only by the shared `cloud` engine name
- allow a configured ElevenLabs fallback after an OpenAI-compatible/Fish provider fails
- preserve the existing engine-level deduplication between fallback rungs
- add regression coverage for both cross-provider rescue and same-provider suppression

## Root cause

The TTS rescue chain excluded any fallback whose engine matched the failed primary engine. OpenAI-compatible Fish and ElevenLabs both use Subwave's `cloud` engine, so a healthy configured ElevenLabs fallback was mistaken for the failed Fish target and skipped. The chain then continued to Piper.

The dispatcher now treats a Cloud target as the combination of engine and provider. Other engines retain their existing engine-only identity.

## Verification

- `tsx scripts/tts-fallback.test.ts`
- `npm run lint` (0 errors; existing repository warnings only)
- live Docker test: stopped the configured OpenAI-compatible Fish service and triggered `generateHourlyTime`; the request used the configured ElevenLabs fallback voice instead of Piper

Fixes #1345

### DIFF
--- a/controller/scripts/tts-fallback.test.ts
+++ b/controller/scripts/tts-fallback.test.ts
@@ -10,7 +10,7 @@
 
 import assert from 'node:assert/strict';
 import {
-  configuredSlot, fallbackTextFor, orderedFallbacks, type RescueSlot,
+  configuredSlot, fallbackTextFor, orderedFallbacks, sameTtsTarget, type RescueSlot,
 } from '../src/audio/tts-fallback.js';
 
 const allUsable = () => true;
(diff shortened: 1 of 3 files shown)


perminder-klair:
Thanks for this — the diagnosis was exactly right, and your commit went in unchanged.

Merged via #1378 (544a3fca), which carries `fbefedec` verbatim plus a few follow-ups from review, so you're credited as co-author on the squash:

- `resolveEngine`'s last resort had the same engine-only comparison, so a persona on a keyless Fish target still fell past a healthy station Cloud provider to Piper whenever `defaultEngine` was also `cloud` — the pre-flight half of the same bug. That hop is probed, unlike the lenient branch above it.
- `recordTts` and `describeRouting` reported a cloud→cloud reroute as `fellBack: false`, hiding the misconfiguration from Stats and `/debug`. Both now share one `rerouted()` predicate with `speak()`.
- `primaryText` follows the same predicate, so a provider reroute gets the cue-stripped rescue text rather than speaking `[laughs]` literally.
- Added coverage for the hardcoded default-engine cloud rung — what an operator with no configured fallback actually gets, and previously untested.

Closing in favour of #1378. Appreciate the live Docker verification in the description, that's what made this easy to land.

PR #1371, opened 2026-08-10
feat(web): sort persona roster by on-air status and name

Jaz666:
## What changed

- pins the persona currently on air to the top of the Personas roster
- sorts the remaining personas alphabetically by display name
- applies the same ordering to both card and table views
- preserves each persona's original form index so editing, validation, and deletion continue to target the correct record

## Why

Persona rosters currently follow creation order, which becomes difficult to navigate on stations with many DJs. Putting the live presenter first and the rest in alphabetical order makes the page easier to scan without changing stored persona order.

The order is calculated from the page's existing on-air snapshot, so it updates naturally on page refresh and adds no polling.

## Validation

- web ESLint and TypeScript checks
- focused ordering checks covering on-air priority, case-insensitive alphabetical order, stable duplicate names, blank names, missing on-air IDs, and preserved form indices
- `git diff --check`

### DIFF
--- a/web/components/admin/PersonasPanel.tsx
+++ b/web/components/admin/PersonasPanel.tsx
@@ -35,6 +35,7 @@ import {
 import { PersonaHero } from './personas/PersonaHero';
 import { SystemPromptModal } from './personas/SystemPromptModal';
 import { PersonaRoster } from './personas/PersonaRoster';
+import { orderPersonaRoster } from './personas/roster-order';
 import { PersonaEditor } from './personas/PersonaEditor';
 
 // The RHF resolver is the two shared schemas the controller validates against,
@@ -466,6 +467,11 @@ export default function PersonasPanel() {
   const onAirPersonaId = data?.onAir?.personaId || activePersonaId;
   const onAirPersona = personas.find(p => p.id === onAirPersonaId) || activePersona;
   const onAirShow = data?.onAir?.show || null;
+  // Display order for the roster AND the editor's counter, so both name the
+  // same slot. Sits below the loading guards rather than in a useMemo, and at
+  // PERSONA_MAX rows the sort per render costs nothing.
+  const roster = orderPersonaRoster(personas, onAirPersonaId);
+  const focusedPosition = roster.find(e => e.index === safeIdx)?.position ?? safeIdx + 1;
   const focusedOk = !isPersonaInvalid(safeIdx);
   // Drives the confirm on ×/Escape: closing the editor keeps edits pending in
   // `personas`, which is how unsaved state used to ride along on the next save.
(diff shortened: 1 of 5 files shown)


perminder-klair:
Pushed a follow-up commit (2429e351) — the sort was solid, but the editor header still counted positions in the form array, so clicking the second visible card could open "5 of 8". The unnamed-persona placeholder ("Persona 5") had the same split across cards, table and editor.

`orderPersonaRoster` now stamps a 1-based `position` beside `index`: `index` stays the RHF field path for editing/validation/deletion, `position` is what human-facing labels read. The panel owns the ordering so the roster and the editor name the same slot.

Also retitled to match the repo's conventional-commit convention. `web` lint + tsc pass.

perminder-klair:
Browser-tested this against an isolated controller, and it turned up a crash my first commit had introduced — fixed in 2b0415d8.

**Setup:** four personas stored as Zara/Hale/Marlowe/Bex with Marlowe on air, so display order (Marlowe, Bex, Hale, Zara) differs from storage order at every slot.

**The counter fix, confirmed against this branch's original head running side by side:**

| card | before (9538e50) | after |
|---|---|---|
| Marlowe (on air, storage 2) | `3 of 4` | `1 of 4` |
| Bex (storage 3) | `4 of 4` | `2 of 4` |
| Hale (storage 1) | `2 of 4` | `3 of 4` |
| Zara (storage 0) | `1 of 4` | `4 of 4` |

**The crash.** Removing a persona hands `watch('personas')` one frame where the dropped index is still counted but holds only the editor's registered fields — no `skills`, no `id`. Ordering the roster on that frame renders a card for it and throws on `p.skills.length`, blanking the panel with "PANEL FAILED TO RENDER". Moving the ordering into the panel is what exposed it: the `useMemo` inside `PersonaRoster` had been re-using the last good order rather than recomputing, so the partial entry never reached the map. The ordering stays in the panel (one source for roster and counter) but is memoised on the same deps, with a comment saying the memo is load-bearing rather than an optimisation.

**Also fixed:** `personaCount` now comes from the same roster. Deleting one of four left the header reading "of 4" against three visible cards — on this branch and before it.

**Verified after the fix:** `1..4 of 4`, then `1..3 of 3` after deleting Bex, no crash, the delete removed `p_bex`, and editing the card shown 4th wrote to storage slot 0 (`p_zara`) — so `index` still drives the RHF field path. Card and list views share the ordering. `web` lint + tsc clean.

One thing I could not exercise: an unnamed persona's "Persona N" placeholder. The schema rejects a blank name, so it only exists on a freshly added row, which is appended last and also sorts last — its display position always equals its storage position. That part of the change is consistency, not an observable fix.

PR #1372, opened 2026-08-10
feat: show the next queue transition in the admin header once it's armed

Jaz666:
## What changed

- replaces the opaque queue transition indicator with a readable text label
- shows the upcoming transition type in the Queue header
- makes clear that the label describes the transition from the currently playing track into the next queued track

## Why

The existing icon does not communicate its meaning unless the user already knows what it represents. A text label makes normal, special-effect, and stem-blend transitions immediately understandable while monitoring the station.

## Validation

- web lint and TypeScript checks
- locally built and deployed the web image
- monitored the admin dashboard during live playback for several hours
- confirmed the label updates correctly as the queue changes


Jaz666:
My first experiment using Github and changing code, all changes made by ChatGPT 5.6 Sol.

perminder-klair:
Nice catch on the precedence: the function matches `radio.liq:287-291` and `:381` exactly, including the bit where washout suppresses dissolve/chop but still allows sweep/blend. Tests pass, both packages typecheck and lint clean.

Two things before this goes in.

**The stem-seam badge disappears from the row that carries it.** `stemSeam` is stamped on the *successor* at pair drain (`queue.ts:1126`), so in the normal steady state the drained item sits at `upcoming[0]` and the armed blend is on `upcoming[1]`. The header only describes `current → upcoming[0]`, so it reports "Normal" while the blend sits one row down with no icon anymore. Either look one further ahead, or keep the per-row icon next to the new header.

**The label is a prediction, and the icon it replaces wasn't.** `applyMixTransition` runs at exactly one call site, inside the drain loop (`queue.ts:1068`). Flags on an unsent `upcoming[0]` are still raw agent proposals, and the pair-drain hold keeps them unsent for minutes. They can be stripped by the variety anti-streak, `effectAllowedFor`, a predecessor loop, the fade veto on `chop`, or a stem seam. The auto-washout on a capped exit goes the other way: it isn't armed until drain, so it never shows. Gating on `upcoming[0].sent` and showing `—` otherwise would fix both directions.

Also worth updating the stem-blend bullet in `CLAUDE.md`, which still says the admin queue badges rendered-blend seams.

perminder-klair:
Also fixed the title to `feat: show upcoming transition in queue header` — release-please reads the squash subject off the PR title, so a bare sentence would land in the changelog uncategorised. Left the scope off since the change spans controller and web.

Jaz666:
Cheers for the feedback, I'll get my LLM to give your comments a look so it can learn for future PRs.
We shall make the required changes shortly.

Jaz666:
All ready for your consideration.

perminder-klair:
Both fixes hold up. I traced the drain and cancel paths: the `sent` gate keeps the header quiet until the pick reaches the mixer, and the badge stays on the row that actually carries `stemSeam`.

Pushed four small follow-ups rather than sending them back to you:

- converted the test to `node:test`, so the eleven cases report individually instead of one exit code
- tooltip on the header naming which pair it describes, which was the point of dropping the icon
- tooltip on the em dash, since under pair drain that is the reading for most of every track and otherwise looks like a broken panel
- listed the test in CLAUDE.md's "Pinned by" tail

Verified in a browser against an isolated controller. `/state` carries `nextTransition`, and with a blend armed one row down the header reads `WASHOUT + SWEEP` while `STEM BLEND` stays on row 02.

Retitled again to cover the gating that landed in your second commit.

Good first PR.

PR #1379, opened 2026-08-11
feat(skills): optional per-skill cron timer

DMCK96:
## Summary

Adds an optional `cron:` field to a skill's `SKILL.md` frontmatter. When set, the scheduler registers a dedicated timer that fires the skill directly on that schedule, bypassing the cooldown and frequency-gate entirely — the same posture as the admin "Run now" override. Lets an operator pin a segment to a real-world moment (e.g. `cron: 0 8 * * *` for a daily 8am weather check) instead of relying on the cooldown/frequency probability gate alone.

- `schemas/skill.ts`: `SKILL_CRON_RE` (structural 5-field check) + `skillCronSchema`, wired into `builtinSkillFileSchema` / `skillFieldsFrom`. Deeper per-field range validation happens where `node-cron`'s own `validate()` runs, at registration.
- `skills/loader.ts`: reads `data.cron` into `cronExpression` on the loaded capability.
- `skills/scaffold.ts`: `writeSkillFile` emits the `cron:` frontmatter line.
- `broadcast/scheduler.ts`: new `syncSkillCrons()` — stops and rebuilds a `Map<kind, ScheduledTask>` from every loaded capability's `cronExpression`; an invalid expression is logged and skipped rather than refused at save time. Runs in the station's own timezone. Called at scheduler start and after every skill mutation.
- `routes/dj.ts`: calls `syncSkillCrons()` after rescan, community install, zip import, create, edit (custom + built-in), reset, and delete; exposes `cron` in the skill file GET/PUT payloads.
- `web/components/admin/skills/SkillEditModal.tsx`: a "Cron timer" field in the skill edit form, with a note on the leave-blank default and station-timezone behavior.
- `web/lib/schemas.generated.ts`: regenerated via `npm run gen:schemas` from the schema change above.
- `docs/custom-skills.md`: documents the field alongside the other frontmatter keys.
- `controller/scripts/skill-schema.test.ts`: extends the existing `skillFieldsFrom` round-trip case with `cron`.

A skill can carry both `cooldown:` and `cron:` — the cooldown still governs autonomous firing from the segment tick, the cron timer ignores it.

### Follow-up fix: `cronOnly` — withhold a cron-timed skill from random autonomous picks

Found in real operation: `cron:` was only ever an *additional* trigger. `availableCapabilities()` (the autonomous segment director's eligibility filter) never checked for a cron timer, so a skill with `cron: 0 10,13,19 * * *` etc. stayed fully eligible for the normal random selection too, off-cooldown — e.g. a skill written for a specific 7:10 moment fired at 13:45 as an ordinary random pick, not through its timer.

- `schemas/skill.ts`: new optional `cronOnly: boolean` field (default `false`), same validation posture as persona `djMode` (present must be a real boolean).
- `skills/_agent.ts`: `availableCapabilities()` now skips any capability with `cronOnly` set — withheld from random selection entirely. `runCapability()` (the cron timer's own call path, and the admin "Run now" override) is untouched, since it reads `allCapabilities()` directly and never goes through `availableCapabilities()`.
- `skills/loader.ts` / `scaffold.ts`: read/write the new frontmatter line.
- `routes/dj.ts`: GET/PUT skill-file routes carry `cronOnly` through.
- `web/components/admin/skills/SkillEditModal.tsx`: new "Only fire on the cron timer" switch under the Cron timer section.
- `docs/custom-skills.md`: documents `cronOnly` and the "cron is a second trigger, not a replacement" gotcha.
- `controller/scripts/skill-schema.test.ts`: updated the `skillFieldsFrom` round-trip for the new field, added a `cronOnly` validation test.

`cronOnly` on its own, without a `cron:` expression, means the skill never fires autonomously at all (still runnable via "Run now").

## Test plan

- [x] `controller`: `npm run lint` (eslint + tsc) — 0 errors
- [x] `controller`: `npm test` — 543/551 pass; the 8 failures are pre-existing on `main` and unrelated to this change (Windows-only `spawnSync tsx` path issues, ffmpeg-dependent voice-library assertions, bash-script-driven `state-bootstrap`/`aio-*` tests) — reproduced identically on a clean checkout before this branch's commits
- [x] `web`: `npm run lint` (eslint + tsc) — 0 errors
- [x] `controller/scripts/skill-schema.test.ts` — all 23 cases pass, including the `cron` round-trip and the new `cronOnly` cases

I have been testing this in my private fork for the last 2-3 weeks and it has been working consistently.


perminder-klair:
### Code review

Found 3 issues:

1. `cron` / `cronOnly` were not added to `OWNED_FRONTMATTER_KEYS`, so every re-save duplicates them — cron edits are silently ignored and a cron can never be cleared. `writeSkillFile` now emits both as typed fields, but `preservedFrontmatter()` only skips keys listed in `OWNED_FRONTMATTER_KEYS` (`controller/src/skills/config-fields.ts` L24-27, untouched by this PR), so it also carries the previous file's values through. That set's own comment states the rule: "the preserve pass below must not carry one through (the form is authoritative for them)". Reproduced against this branch:

```
=== save 2 (cron changed 8am -> 9am) ===
name: dabbers
cron: 0 9 * * *
cronOnly: true
cron: 0 8 * * *
cronOnly: true

parse2 => {"cron":"0 8 * * *", ... "malformed":"Map keys must be unique at line 4, column 1:"}

=== save 3 (operator clears cron + cronOnly) ===
cron: 0 8 * * *
cronOnly: true
```

The duplicate key makes YAML parsing throw, so `parseFrontmatter` falls back to the legacy line parser and resolves the **old** value, and `loadSkillDir` logs "SKILL.md frontmatter is not valid YAML" on every scan. Fix is to add `'cron', 'cronOnly'` to that set (which also closes the `RESERVED_CONFIG_KEYS` half, since a `tool.mjs` can currently declare a config field named `cron`).

https://github.com/perminder-klair/subwave/blob/7ef4dd3762e9df707499d60582c334dcaa417a6c/controller/src/skills/scaffold.ts#L112-L117

2. The cron callback applies none of the four gates every other autonomous talk site applies (CLAUDE.md: "Manual `/dj/segment` command routes ... are intentionally **not** budget-gated — an explicit operator action always fires"). `skillsTick` checks `autoVoiceAllowed()`, `programme.onAir()`, `djCallsAllowed()` and `optionalSegmentsAllowed()` before running; the cron task calls `runCapability()` straight away. A scheduled timer is not an explicit operator action, so as written a cron skill still speaks and spends tokens with the station voice switch off (CLAUDE.md: `tts.enabled: false` makes the station "music only — no ... skill segments"), past the daily LLM hard cap ("makes **no** model call"), with zero listeners, and mid-programme-episode.

https://github.com/perminder-klair/subwave/blob/7ef4dd3762e9df707499d60582c334dcaa417a6c/controller/src/broadcast/scheduler.ts#L904-L913

Compare the gates on the adjacent autonomous tick:

https://github.com/perminder-klair/subwave/blob/7ef4dd3762e9df707499d60582c334dcaa417a6c/controller/src/broadcast/scheduler.ts#L685-L690

3. Registered crons keep a stale timezone after a Settings change. `getStationTimezone()` is read once and baked into `cron.schedule(..., { timezone: tz })`, and `syncSkillCrons()` is only called at scheduler start and from the skill CRUD routes. `settings.update()` applies a timezone change live (`controller/src/settings.ts` L2041 `setStationTimezone`) but never re-syncs the crons, so they keep firing on the old zone until an unrelated skill edit or rescan — contradicting the new doc line that a `cron: 0 8 * * *` "fires at 8am local time wherever the station is configured to be".

https://github.com/perminder-klair/subwave/blob/7ef4dd3762e9df707499d60582c334dcaa417a6c/controller/src/broadcast/scheduler.ts#L903-L905

perminder-klair:
### Verification

Booted an isolated controller from this branch (spare port, temp `STATE_DIR`) plus the worktree web dev server, and drove the feature end to end. The unit tests pin the pure helpers, but until now no test in the PR had ever caused a cron to fire.

Verified at `bcc4930e`:

| Check | Evidence |
|---|---|
| Cron on a **disabled** skill | `cron "minutely" stood down — skill is disabled` |
| Persona allowlist | `cron "minutely" stood down — the on-air persona does not run this skill` |
| Cron fires when eligible | segment traces at `11:50:00 / 11:51:00 / 11:52:00 / 11:53:00`, `ok=true`, reaching TTS and airing an SFX |
| Out-of-range cron | 400 with `fieldErrors.cron`, and no skill folder written |
| 6-field cron | accepted; 4-field refused by the schema |
| Hand-edited bad cron | scheduler logs and skips, skill still loads, `cronInvalid: true` in the file GET |
| Timezone change | `station timezone changed — re-registering skill crons`, both crons re-registered at `Asia/Kolkata`, and not re-fired on unrelated saves |
| Community install | `cron: 0 7 * * *` + `cronOnly: true` survive install (stub catalog); an unfirable one refused, nothing written |
| Re-save round trip | cron 7am→9am wins, clearing removes both lines, no duplicate keys, no YAML warning |
| Deregistration | after clearing, only the other cron re-registers |
| Admin UI (Playwright) | 6/6: field loads from disk, `cronInvalid` warning renders, the server field error lands on the input, bad value not persisted, valid one saves, warning clears |

`npm test` 565/565 on Linux, so the 8 failures you saw look environment-specific rather than real.

**One gap the run found**, in `bcc4930e` rather than in your commits: the eligibility checks name their reason in the booth log, but the four station gates returned silently. A registered cron sat mute for four ticks and the only way to learn that `djCallsAllowed()` was closed was to read `listeners.ts`. Fixed in a follow-up commit landing shortly: `skillCronStandDownReason()` names which gate closed, and `skillCronAllowed()` is defined as "no reason" so a gate added to one can't be forgotten in the other.

Still unverified: behaviour across a DST boundary in a non-UTC zone. That is node-cron's own territory rather than this PR's code, but it is the obvious next question for a feature whose selling point is "8am local".

perminder-klair:
Checked the DST question I left open.

Simulated node-cron's scheduler loop with its own `TimeMatcher` against real transitions in `Europe/London` and `America/New_York`. The docs claim holds — `0 8 * * *` fires once at 08:00 local on an ordinary day, on the spring-forward day, and on the fall-back day.

Two edges come from node-cron 3.0.3, not from anything in this PR:

- **Fall-back skips the repeated hour.** A `*/10` cron fires 19 times instead of 25, an hourly one misses a beat. `scheduler.js` guards on `lastExecution.getTime() < date_tmp.getTime()`, and `date_tmp` is the wall-clock-shifted date, which moves backward at the transition.
- **Spring-forward silently skips** a cron pointed inside the hour that does not exist that day.

Nothing double-fires, and that is the same guard, so a naive fix would trade a missed segment for a repeated one.

Documented in `0e2ce0ed` instead of fixed. node-cron 4.x rewrote scheduling, but the upgrade touches every cron in `scheduler.ts` (playlist refresh, hourly, segment tick, nightly doctor), so it wants its own PR.

Thanks for this, and for running it in your fork for a few weeks before sending it. The `cron` / `cronOnly` split was the right shape to land on; everything found since has been about the edges around it.

perminder-klair:
One more docs pass, prompted by a mistake worth writing down.

I was about to add `cron: 0 22 * * *` + `cronOnly: true` to the `moon-phase` example as a worked demo. It would have been wrong. A cron fires through `runCapability()`, the same forced path as **Run now**, where `forcedSchema()` has no `air` field and `runCapability` throws on empty text. The model cannot decline. The autonomous director's `segmentSchema()` does carry `air`, and tells the model that silence is often the best answer.

A skill's own `{ available: false }` is therefore inert on the cron path. `moon-phase` would have announced the moon every night at 22:00, against its own brief. Both examples keep their `cooldown` and gain no cron, and `docs/custom-skills.md` now states the rule, which is more use than the example would have been.

`379da212` also brings the manual's Custom Skills page up to date, since it had no mention of the feature at all. It now carries `cron` / `cronOnly` in the frontmatter block and the three-way split the gate work established: **Run now** ignores every gate, a **cron** ignores the frequency gate and cooldown but still obeys the enabled toggle, the persona allowlist and the four station gates, and the ordinary tick obeys everything.

PR #1386, opened 2026-08-12
fix(tagging): calibrate audio moods per label and derive energy for propagated tracks

### LINKED ISSUE #1362: fix(tagging): derive energy from audio_mood_scores_json for source: "propagated" tracks instead of trusting the propagated guess
Context

tracks.moods/tracks.energy (the fields the picker and the strict-filter locks actually read — music/library-db/browse.ts line 32, music/picker.ts) come from an LLM tagging pass over text metadata (genre/lastfm_tags/lyrics). When a track has thin or misleading metadata, the tagger falls back to source: "propagated" — on our ~11k-track library that's 4,507 tracks (41%), and per the code these are effectively inherited/guessed values, not a real per-track judgement.

Separately, tracks.audio_moods/tracks.audio_mood_scores_json come from a genuine audio-content classifier (CLAP-style, listens to the actual track) — this data exists for the whole library already but isn't used by the strict-filter path at all (browse.ts line 32 only checks tracks.moods), and is only OR-combined (not preferred) in the discovery-tool lookup in library-db/queries.ts lines 55-58.

Concrete example that surfaced this: Loop Da Loop – What I Like About You (a Fatboy Slim/Propellerheads-era big-beat track on a "Switched On" club compilation) had genre: "Soundtrack" (wrong), moods: ["calm","night"], energy: "low", source: "propagated", confidence: 0.48 — and was getting picked for a strict instrumental+low-energy overnight ambient show. Its audio_mood_scores_json correctly scored energetic: 0.42, workout: 0.43 vs calm: 0.003, night: 0.015 — the real signal was sitting right there, unused.

What we tried, and what we learned

We backfilled energy for all 4,507 propagated tracks locally, deriving it from audio_mood_scores_json: avg(energetic, workout, driving, celebratory, festival) − avg(calm, reflective, spiritual, focus, night), bucketed into low/medium/high. We calibrated the two thresholds against the ~6,300 tracks with source: "llm" (a real per-track judgement, used as ground truth) — the three energy buckets showed a clean, monotonic separation on this diff (medians 0.076 / 0.131 / 0.218 for low/medium/high), so the signal is real and worth using. 2,147 of the 4,507 propagated tracks (48%) got a different, evidence-backed energy value out of this.

We deliberately did NOT also copy audio_moods into moods — checking label-firing rates across the whole library first: energetic fires on 61.7% of all tracks and rainy on 43.5%, while calm fires on only 0.9% and spiritual on 0.2%. That's not a plausible real distribution for a mixed library — the raw top-label extraction on audio_mood_scores_json looks miscalibrated (a single global threshold across labels with very different baseline score ranges, most likely), so blindly trusting the label list would trade one systematic bias for another. The underlying continuous scores are more usable than the label list they get thresholded into.

Proposed change

1. For energy specifically: when source = 'propagated', derive it from audio_mood_scores_json (the high-minus-low dimension diff above, or something better-calibrated) instead of carrying forward the propagated guess. Happy to share our calibration script/thresholds — they're a reasonable starting point, not claimed to be optimal.
2. Separately, worth a look regardless of (1): the audio_moods label-extraction thresholds seem to need per-label calibration rather than one global cutoff, given the 61.7%/43.5%-vs-0.9%/0.2% firing-rate spread above. This affects anything reading audio_moods directly (the OR-match in queries.ts), not just our energy proposal.
3. Longer term, the strict mood/energy locks (browse.ts line 32, gated behind filtersStrict per show) could weight audio_mood_scores_json directly for tracks where source = 'propagated' or confidence is low, rather than only reading the text-derived moods column.

Not attached to a specific implementation — mainly want to flag that the audio analyzer's data is under-used relative to how good (energy-wise, at least) it turned out to be once calibrated, and that the raw audio_moods labels need a calibration pass before they're safe to lean on directly.

perminder-klair:
Closes #1362 (items 1 and 2).

## The problem

`tracks.moods`/`tracks.energy` come from an LLM reading each track's **text** metadata. When that metadata is thin, the tagger inherits a value from embedding neighbours and stamps `source = 'propagated'` — a guess built on guesses. On the reporter's 11k library that is 41% of the catalogue.

Meanwhile the analyzer had already scored those same tracks from their **audio**, and the answer was sitting in `tracks.audio_mood_scores_json`, read by nothing on any playback path.

The reported case: a big-beat dance track with genre wrongly tagged `Soundtrack` propagated to `energy: 'low'`, `moods: [calm, night]` — which then satisfied the strict energy lock on an overnight ambient show and aired there. Its audio scores had it right.

## What changed

**1. Per-mood calibration (`music/audio-calibration.ts`)**

CLAP cosines are not comparable across moods — each prompt sits at its own baseline in the shared space, so ranking a track's raw cosines ranks *prompts*, not tracks. Measured spread on that library: `energetic` 61.7% / `rainy` 43.5% vs `calm` 0.9% / `spiritual` 0.2`%`.

> Note: the issue diagnosed this as "a single global threshold across labels". It isn't — selection was already per-track relative top-K. That makes the skew *worse*, not better, and means per-label thresholds wouldn't fix it. The fix is centring each mood against its own library-wide distribution: `z = (score − mean) / sd`, then the same relative top-K on the z axis.

Baselines come from **every stored score map, library-wide** — a distribution built only from the tracks the tagger found hardest to read would be a biased yardstick. Below 200 scored tracks it falls back to raw selection, byte-for-byte the old path.

**2. Energy correction for propagated tracks (`music/propagated-energy.ts`)**

Mean z of the high-arousal moods minus mean z of the low-arousal ones. Where that is decisive, it overrules the propagated guess.

**Deliberately two-sided, not a three-way bucketing.** `audioEnergy` returns `high` / `low` / **null**, and the null is the design: bucketing the ambiguous middle into `'medium'` would replace one guess with another and call it evidence. Returning null leaves the existing value alone — so a mis-tuned threshold costs *coverage*, not correctness.

Scope is `source = 'propagated'` only; an `llm`/`manual`/`uncertain-llm` energy is a real per-track decision and is never touched. The write touches the `energy` column alone, so a corrected row still reports honestly that its **moods** are inherited.

## Two design points worth review

**The pass is now three ordered phases** — score → baselines → labels. Labels can't be picked during scoring, because on a full re-score the baselines are a property of the cosines that loop is still writing. A crash between phases leaves `audio_moods` NULL, which the existing incremental scope re-picks up, so the split is resumable rather than lossy.

**`mood_vocab_hash` now carries the calibration version** (`vocabhash:2`). The two invalidate very different amounts of work — a vocabulary change makes the *cosines* stale (full CLAP re-score), a calibration change makes only the *labels* stale (re-derive from disk, no analyzer). Folding the version into the vocab hash would strand every install not running `ANALYZER_HEAVY` — most of them — on stale labels. A legacy bare hash parses as version 0, so **upgrading relabels rather than re-scoring**. Composite string, so no schema migration.

## Not included

Issue item 3 — weighting scores inside the strict `moodLock`/`energyLock`. Those locks are hard with no never-starve on the agent path, so introducing a continuous score there changes starvation behaviour on thin shows. That wants its own spec.

## Calibration honesty

`ENERGY_HIGH_Z`/`ENERGY_LOW_Z` (±0.35) and `DEFAULT_MARGIN_SD` (0.5) are defensible starting points, **not measured optima** — they want a pass against a real library's label distribution. The two-sided design above is what keeps that safe: being slightly off means fewer corrections, not wrong ones. @MYBLtd offered their calibration script on the issue and it would be worth checking these against it.

## Testing

- `scripts/audio-calibration.test.ts` — 21 pure tests. The load-bearing one reproduces #1362: on a synthetic library where one mood has a higher baseline, raw selection has it winning **every single track**; calibrated selection tracks the actual music.
- `scripts/propagated-energy.test.ts` — real SQLite. Every query added here is new SQL, which pure tests over the selection maths would pass straight through. Pins the reported case (low → high), that `source = 'llm'` is never touched, that only the energy column is written, and that re-running is idempotent.

`npm run lint` 0 errors, `tsc --noEmit` clean, full suite **612/612 pass**.


perminder-klair:
## Tested against a real 1,405-track library

Ran both passes over a copy of a real `library.db` (1,413 tracks, 1,405 with CLAP audio scores, 162 propagated). The stored `mood_vocab_hash` was a bare legacy hash, so this exercised the upgrade path — **relabel from stored cosines, no analyzer involved**.

### This caught a real bug

`relabelFromStoredScores` wrote its batch **while the read cursor was still open**, which better-sqlite3 refuses:

```
TypeError: This database connection is busy executing a query
    at setTrackAudioMoodLabelsBulk (library-db/audio-moods.ts)
    at relabelFromStoredScores (audio-moods.ts)
```

It worked for any library that fit in one 500-row write batch (the flush landed after the loop) and threw on every library larger than that — so it would have broken on essentially every real install while passing the 300-track fixture. Fixed by keyset pagination (`193abebf`); fixture sized past a batch, and I verified the new test fails with exactly this error on the streamed version.

### Label distribution, before → after

The reported skew, reproduced and fixed. Note `calm`, `focus` and `festival` fired on **literally zero** tracks before — even more extreme than the 0.9% in the issue.

| | before | after |
|---|---|---|
| `energetic` | **72.4%** | 13.7% |
| `rainy` | 40.2% | 12.9% |
| `workout` | 39.6% | 28.6% |
| `calm` | **0.0%** | 11.9% |
| `focus` | **0.0%** | 17.9% |
| `festival` | **0.0%** | 14.3% |
| `spiritual` | 0.1% | 17.1% |
| `romantic` | 0.1% | 12.0% |
| spread (max/min) | 72.4% / 0.0% | 28.6% / 5.8% |
| moods that ever fire | 14 of 17 | **17 of 17** |

Labels per track moved 1.83 → 2.34.

### Energy correction

162 propagated tracks: **38 corrected, 70 confirmed, 54 left alone** (audio not decisive). The split moved from `{low 61, medium 45, high 56}` to `{low 72, medium 23, high 67}` — the ambiguous `medium` bucket is where propagation parked what it couldn't judge, and that is what drains.

More conservative than @MYBLtd's 48%, which is the two-sided design working as intended: only decisive cases are touched.

### Caveat worth naming

I can't verify individual corrections by ear, and a couple look questionable — `studyBreak — Waves` (moods: `calm`) went `low → high`, which for a track that sounds like it should be lo-fi is either a genuine catch or a threshold that's slightly hot. The aggregate is clearly right; the per-track threshold (±0.35 z) still wants a second opinion. That's the one number in this PR I'd most like checked against your calibration data.

perminder-klair:
### Code review

Found 1 issue:

1. A library scored below `MIN_BASELINE_TRACKS` (200) is stamped with the calibrated `mood_vocab_hash` even though its labels came from the raw-cosine fallback, so it never gets re-labelled once it grows past the floor. `runAudioMoodPass` calls `db.setAudioMoodVocabHash(composeMoodStateHash(hash))` unconditionally — the same stamp a calibrated pass writes — while the log line right below it already knows the difference (`uncalibrated — library too small`). On a later pass with a grown library, `moodPassAction` sees a matching vocab hash and a matching `CALIBRATION_VERSION`, returns `'none'`, and `relabelFromStoredScores` never runs; `idsNeedingAudioMoods` skips those rows too since `audio_moods` is non-NULL. Only tracks scored after the crossing get calibration, so a station that started small stays permanently on the selection issue #1362 was filed to fix. Encoding "uncalibrated" in the stamp (or re-labelling when `baselines` flips from null to usable) would close it.

https://github.com/perminder-klair/subwave/blob/193abebfc42de847066dd9b2cb14e487d7192137/controller/src/music/audio-moods.ts#L231-L239

https://github.com/perminder-klair/subwave/blob/193abebfc42de847066dd9b2cb14e487d7192137/controller/src/music/audio-calibration.ts#L279-L295

perminder-klair:
Fixed in ec163da5.

- **Uncalibrated stamp** (the review finding): a pass that couldn't calibrate now stamps `UNCALIBRATED_VERSION` (0) via `moodStateHashFor`, so the relabel stays owed and a library that grows past the floor re-derives its labels instead of reporting `'none'`.
- **`baselinesUsable` gated on the maximum per-mood `n`**, so one established mood admitted the whole set — including a vocabulary entry scored on a handful of tracks, whose near-degenerate sd makes its z explode and win every track. `prunedBaselines` drops the thin moods and keeps the rest (requiring every mood to clear the floor would refuse calibration library-wide over one new entry). `propagated-energy` takes the same path, where a thin mood on either arousal list would swing the axis.
- **N+1 read**: phase 3 reuses phase 1's score maps rather than re-`SELECT`ing each one, bounded to the incremental scope — a full re-score still relabels from disk, paged.
- Dropped the now-unused `setTrackAudioMoodsBulk`; moved `propagated-energy.test.ts` onto `node:test`.

Suite 626/626, lint 0 errors.

MYBLtd:
Great fix, and thanks for the detailed writeup — the per-mood z-centring is exactly the right move, and I like that the energy correction is two-sided (high/low/null) rather than forcing a medium bucket.

Our numbers, for what they're worth against your ±0.35 z / 0.5 sd starting points:

We derived a high-minus-low arousal diff per track — `avg(energetic, workout, driving, celebratory, festival) − avg(calm, reflective, spiritual, focus, night)` on the raw (uncalibrated) `audio_mood_scores_json` — and calibrated a threshold against the ~6,300 `source: "llm"` tracks on our library (treated as ground truth, not touched by our backfill). The three energy buckets on that ground-truth set showed a clean, monotonic separation on this diff:

- low: median 0.076
- medium: median 0.131
- high: median 0.218

We only used this to backfill `energy` for `propagated` tracks (2,147 of 4,507 got a different, evidence-backed value), and deliberately left `moods` alone given the 61.7%/43.5% vs 0.9%/0.2% firing-rate skew — sounds like that's exactly what #1 in this PR now fixes properly at the source.

Happy to share the actual backfill script if useful, though it's a one-off SQL+Python pass against raw scores, not built against your new z-calibrated pipeline — probably more useful as a sanity-check data point than as code to reuse.

Agreed on holding off on item 3 (weighting inside the strict locks) — didn't have starvation behaviour in mind when writing that up, and "wants its own spec" is the right call.

PR #1402, opened 2026-08-13
feat: add Never Play Again track exclusion

solgate56:
## What

Adds an admin action, **Never Play Again**, for the currently playing track.

The action:

- blocks the track from future SUB/WAVE selection
- removes queued duplicates
- optionally excludes the underlying file from Navidrome via a root `.ndignore`
- triggers a Navidrome scan
- skips the current track immediately
- adds a matching admin dashboard control
- reverses the Navidrome exclusion when the track is later unblocked

Navidrome catalogue exclusion is optional and only enabled when `NEVER_PLAY_LIBRARY_PATH` is configured with a read/write mount of the Navidrome music-library root.

## Why

Sometimes an operator doesn't just want to skip a track — they don't want the station to select it again.

SUB/WAVE already had the pieces needed for immediate skipping and persistent blocklisting, so this adds a single operator action that combines them safely.

The Navidrome integration is optional so existing deployments remain unchanged unless explicitly configured.

## Safety / behaviour

- admin-authenticated only
- verifies the confirmed track is still on air before making any changes
- block happens before skip to prevent immediate reselection
- `.ndignore` updates are atomic and serialized
- Navidrome failures degrade with a warning rather than undoing a successful SUB/WAVE block
- Liquidsoap skip takes priority over the best-effort auto-playlist refresh

## Testing

- focused Never Play Again tests: 31/31 passing
- blocklist tests: 2/2 passing
- controller typecheck/lint clean
- web typecheck/lint/build clean
- CLI typecheck clean
- live-tested against SUB/WAVE + Navidrome, including `.ndignore` paths containing square brackets


perminder-klair:
@kanrei-subwave review this pr

solgate56:
Both blocking review findings are fixed and pushed.

- Bulk unblock now captures entries before removal, reverses any owned `.ndignore` rules, and triggers a single shared Navidrome scan for the batch.
- Never Play Again now only records `libraryPath` / triggers a scan when `ignoreAdd()` actually adds a new rule, so pre-existing/manual `.ndignore` entries are not claimed or later removed.

Regression coverage was added for both cases. Focused Never Play Again / blocklist / ndignore tests pass, controller typecheck is clean, and lint has 0 errors.

The branch has also been reconciled with current `develop` and GitHub reports the PR mergeable again.

Ready for re-review.

perminder-klair:
@kanrei-subwave review this pr again

solgate56:
All four re-review blockers are addressed in `aa028760`:

- re-check the confirmed track after metadata lookup and again immediately before skip; the optional Navidrome scan is deferred until after a successful skip attempt
- reject CR/LF and other unsupported control characters in library paths before `.ndignore` mutation
- root-anchor generated `.ndignore` patterns, while retaining reversal compatibility with legacy unanchored patterns
- sent duplicates now go through the existing mixer-aware queue cancellation path, and any copies that can no longer be recalled are reported

I also surfaced bulk-unblock reversal warnings in the Blocked tab and clarified the header tooltip that Navidrome exclusion is conditional on configuration.

Regression coverage has been added for the new cases. Focused suite: 73/73 passing. Controller/web/CLI typechecks pass, controller/web lint have 0 errors, and `git diff --check` is clean.

Ready for re-review.

PR #1407, opened 2026-08-15
fix(broadcast): level-match loop transitions

Jaz666:
## Summary

- capture the outgoing Loop bar from the raw source instead of the logarithmic outgoing fade
- release the completed loop branch with the linear `1 - elapsed / duration` envelope as the incoming track rises
- remove the incorrect global `x2` comb-cascade makeup that doubled the direct capture pass
- add a deterministic real-image `scripts/fx-render-test.sh loopcheck` regression check and document the measured invariant

## Root cause

Loop previously kept the outgoing track hot under the rising incoming track, then held its repeated bar near full level until a late master release.

The first revision corrected that envelope but retained `amplify(2.0, looped)`, based on the assumption that the comb cascade had a fixed -6 dB loss. In Liquidsoap, `comb(feedback=0.0)` emits a 0 dB delayed copy. The one-bar gate plus power-of-two delays place those copies in non-overlapping bar slots at unity, so the global makeup instead doubled the direct path during the capture pass, before any delayed tap existed.

The loop now keeps the raw capture and its repeats at unity, then applies the outer linear release. The ride-out low-pass still deliberately darkens the repeated bar as the incoming track takes over.

## Impact

Loop transitions enter at the current programme level without the measured approximately +6 dB jump. Sweep, Washout, Dissolve, Chop and Blend behaviour is unchanged.

## Validation

- `scripts/fx-render-test.sh loopcheck` using `savonet/liquidsoap:v2.4.5` and deterministic pink noise:
  - before final fix: dry `-38.369543 dB`, Loop `-32.454259 dB`, delta `+5.92 dB` — expected regression failure
  - after final fix: dry `-38.369543 dB`, Loop `-38.382377 dB`, delta `-0.01 dB` — pass
- `docker run ... savonet/liquidsoap:v2.4.5 liquidsoap --check /radio.liq`
- `bash -n scripts/fx-render-test.sh`
- `git diff --check`


Jaz666:
Improved loop transition now heard on my live test station, fix looks good.

Something to note that's related, a loop on the tail end of a track that's already fading out causes a noticeable audible oddity (quiet-loud-quiet-loud) - maybe look at disqualifying loops from track that have a notable fade out.

perminder-klair:
Fixed and pushed in `cf52e016`.

The `×2` gain after the comb cascade was the problem. The repeats were already at the correct level, so that extra gain doubled the first captured bar before any delayed taps started.

I removed the gain and added `scripts/fx-render-test.sh loopcheck` using the real Liquidsoap image. The measured difference from a normal crossfade went from **+5.92 dB** to **−0.01 dB**. All CI checks pass.

PR #1410, opened 2026-08-15
fix(broadcast): don't start a jingle while a DJ clip is on air

obareau:
### The gap

#997 guarded one direction: a voice handoff waits out a jingle that is already feeding, via `jingle-playing.json` and `waitForJingleClear` in the controller.

The other direction was unguarded. The jingle rotate can hand a stinger to the music chain **while a DJ clip is on air**, and the stinger simply plays over the voice. The gate already knew about beds:

```liquidsoap
jingles = source.available(jingles, {not bed_on_air() and not music_starved()})
```

A bed on air makes jingles unavailable; a voice on air did not.

### Why it may not have shown up

It is only audible on stations that actually rotate jingles. Ours ran with `jingleRatio: 0` for months, so nothing ever collided. Raising it to 12 surfaced the overlap within a day — most often on shows with chatty personas and `extended` scripts, where a clip can hold the voice chain for 15-60s and a track boundary lands inside that window.

### Implementation — a deadline, not a boolean

`source.is_ready(voice_queue)` looks like the obvious predicate and is a trap: on a `request.queue` it latches `true` forever after the first push. Measured on Liquidsoap 2.4.5, one 6.4s clip pushed at t=2s, queue drained and empty afterwards:

```
t=2s   is_ready=false  queue=0
t=6s   is_ready=true   queue=0
t=12s  is_ready=true   queue=0
t=24s  is_ready=true   queue=0   <-- still
```

Gating the rotate on that would starve jingles permanently after the first segment — worse than the bug it fixes. A timestamp cannot latch: at worst it is already in the past and the gate opens.

`voice_until` is armed from `voice_marker`, which already fires on every clip and carries the filename. `request.duration` is documented as expensive, but that hook is `synchronous=false` and runs off the streaming thread. The 30s fallback keeps the guard useful when a clip cannot be measured, and stays bounded by construction.

### Verification

- `liquidsoap --check` passes
- Running in production on a 24/7 station since 2026-08-14, no dead air, jingles still rotating (`jingleRatio: 12`)
- The `is_ready` measurements above are reproducible with a short script against any `request.queue`

Happy to adjust naming or comment density to taste.

### DIFF
--- a/liquidsoap/radio.liq
+++ b/liquidsoap/radio.liq
@@ -869,6 +869,30 @@ bed_on_air = ref(false)
 # the future cannot latch — at worst it is already in the past.
 voice_until = ref(0.)
 
+# Hard ceiling on how far ahead voice_marker may push that deadline. The gate
+# fails toward SERVING jingles, deliberately: a measured duration that comes back
+# wrong (corrupt header, a filename that resolved to some other media) would
+# otherwise silence the rotate for as long as that number says, and unlike the
+# collision the gate fixes, that failure is inaudible — nothing in the log says
+# why the stingers stopped.
+VOICE_GATE_MAX_SEC = 120.
+
+# Two things this gate deliberately does NOT do:
+#
+#   - It arms at the clip's on_metadata, i.e. after the 0.5s say.txt poll and the
+#     silent lead-in. waitForJingleClear (voice-io.ts) covers a jingle already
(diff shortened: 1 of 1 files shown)


perminder-klair:
Thanks for the clear write-up and testing. I reviewed the final diff and it looks good to merge.

PR #1430, opened 2026-08-19
fix(dj): reach banter and guest exchanges with the station house rules

### LINKED ISSUE #1420: Banter not getting house rules?
On the basis that the house rules I use have TTS guidance in them, the banter text does not seem to be using the house rules, as none of the TTS guide rules are being applied.

perminder-klair:
Closes #1420.

## The issue

House rules with TTS guidance in them were applied to links, idents and segments but never to banter.

`settings.djHouseRules` is documented as *"the one operator block appended to **every** spoken-output prompt"*, but it had exactly two append sites — `renderDjPrompt` (scripted talk) and `agentPersonaPreamble` (picker / request / segment agents), both in `settings/persona.ts`.

The multi-voice cast prompts go through neither. `generateBanter` and `generateProgrammeExchange` write a whole exchange in one structured call, so each hand-rolls a system prompt around a per-call speaker enum. That made them the one class of spoken output the rules never reached.

## The fix

`castHouseRulesBlock()` is the third append site, wrapping the existing private `houseRulesBlock(scope)`.

Both call sites share **one** scope wording rather than each framing it themselves: the two prompts emit the same `{speaker, text}` shape, and a rule like *"write numbers out in words"* mangles a persona id as readily as it mangles a track id on the agent path — so the framing exempts `speaker` explicitly, the same way the agent framing exempts ids/reasons/kinds.

`banter.ts`'s system prompt becomes an exported pure `banterSystem`, matching the precedent `exchangeSystem` already set for the grounding rule.

## Test

`scripts/house-rules.test.ts` gains the cast paths — on, off, and the byte-identical round trip after clearing.

It asserts on the **rendered** prompts, not on `castHouseRulesBlock()`. This bug *was* a builder that existed and was never appended, so a test on the builder passes on the broken code. Verified: reverting just the two appends fails the test with `AssertionError: banter path carries the block`.

## Also

- Comments in `defaults.ts` / `vocab.ts` / `persona.ts` and `docs/internals/broadcast.md` updated from "both paths" to all three.
- Admin System-prompt modal copy — it already promised *"appended to everything the DJ speaks"*, which is now true; the surrounding text now names the guest exchanges alongside the tool-using agents.

## Verification

- `npm test -- house-rules` passes; fails as expected with the appends reverted.
- Full controller suite 681/686 — the 5 failures (`aio-log-link`, `analyzer-python`, `state-bootstrap`, `observatory-scale`) are pre-existing environment failures, confirmed at baseline with the changes stashed.
- `controller` and `web` lint clean (0 errors); `gen:schemas` produces no mirror diff.


perminder-klair:
## Live verification

Ran this against an isolated controller booted from the worktree (own port, throwaway `STATE_DIR`) — the real station was never touched. Guest show pinned on air (host Marlowe + guest Wren, banter on), with house rules set to two checkable rules:

```
Never use digits — spell every number out in words.
Begin every line you speak with the word "Listen," before anything else.
```

The controller was pointed at a local stub provider that logs the request body, so these are the system prompts actually put on the wire — not a rendering of the pure builder.

### Before vs after

| | banter system prompt |
|---|---|
| without the fix | 1358 chars — `Station house rules` **absent** |
| with the fix | 1680 chars — block present, rules verbatim |

The appended block, as sent:

```
Station house rules — follow these in every line's spoken text (the words the
listener hears on air); they do not apply to the "speaker" field, which stays
an exact persona id from the cast list:
Never use digits — spell every number out in words.
Begin every line you speak with the word "Listen," before anything else.
```

### Scope check

Firing `programme-intro` on the same guest show produced 8 LLM calls:

- the **guest-show exchange** prompt (1653 chars) — block present ✅
- the scripted **`djSystem`** prompt (1901 chars) — block present, as it already was pre-fix ✅
- 4× **`generateProgrammePlan`** (432 chars) — block correctly **absent**. That call writes the producer's notes *for the host*, not spoken text; the host then speaks through a prompt that does carry the rules. No over-reach.

### Not covered

Whether a real model **obeys** the rules is untested — the Ollama cloud endpoint this install points at is out of quota (`requires a subscription` / `reached your weekly usage limit`) across every available model, so a stub was used rather than spending on another provider. Worth stating plainly that this was never a compliance bug: the rules simply weren't in the prompt, and now they are.

TTS render failed throughout (`piper ENOENT` — not installed on this machine). That is downstream of the LLM call and unrelated to the change.

Issue #842, opened 2026-07-04
Era Selection and Compilation Albums


Jaz666:
It's standard practise in MusicBrainz to tag Albums with

- Date (of this release)
- Original Release Date
- Original Release Year.

This covers regular releases and re-issues, but causes problems for Compilation Albums, as it isn't standard tagging practise to tag each individual tracks in compilations with their original release year.

So, "100 Hits: 70s Chartbusters", with it's release date of 2013 gets played when you are trying to limit a show to the music of the 2010s.  Similarly the compilation gets missed when you are trying to play songs from the 70s.

We could either exclude albums by "Various Artists", which isn't ideal for everyone, or come up with a way for Sub/Wave to tag these files.  Either on the fly, or in a library scan.

Ideas?

brandenespinoza:
> it isn't standard tagging practise to tag each individual tracks in compilations with their original release year.

I want to double down on this observation That said, on compilation albums I manually (or via llm) adjust the Original Release Date and Originial Release Year to the song's actual originial release year.  I use Claude for this (and a skill I've been slowly evolving). It is about 85% correct.  So not as dependable as I wish, but still a better listening experience when listening to eras. 

I would advise against using "Various Artists" as the Album Artist tag (and I admit that you acknowledge it wouldn't be great for everyone).  Many of us do not include ANY "Album Artist" metadata on compilation albums with songs from many artists.

perminder-klair:
There's a fix for this up in #1059. How it works:

Era filtering now judges tracks by original release year instead of the plain year tag. Three sources feed it:

1. If the album has an Original Release Date tag, Navidrome exposes it and the library scan picks it up for free. So reissues tagged the MusicBrainz way just work now.
2. For compilation tracks, SUB/WAVE asks MusicBrainz for each song's earliest release date during the library scan (rate limited to 1 request/sec, resumable, results cached so it only happens once per track). On my library it resolved about 70% of compilation tracks.
3. Anything it can't resolve is left out of era-bounded shows rather than played in the wrong decade. Regular shows still play them fine.

@brandenespinoza took your advice on detection: it uses Navidrome's compilation flag (from the COMPILATION tag), not the Album Artist field, so comps with no album artist are handled. And if you've already fixed original years per track with your Claude skill, those tags won't be read directly (Navidrome doesn't expose per-song original dates over Subsonic), but the MusicBrainz lookup should land the same years without the 15% error rate.

[...]

PR #1431, opened 2026-08-19
fix(music): era resolution — route the listener-facing surfaces, add an operator override, and trigger on unflagged anthologies (#1418)

### LINKED ISSUE #1418: Era resolution rarely fires: original-year lookup is gated on Navidrome's compilation flag, which reissue anthologies do not set
Follow-up to #842 / #1059. The era pipeline built there is sound, but on my library it
almost never runs: it is gated entirely on Navidrome's compilation flag, and the reissue
anthologies that motivated #842 are not flagged. Separately, three surfaces read the raw
`year` and bypass the resolution entirely, so even a correct `originalYear` would not
change what the DJ says on air.

SUB/WAVE 1.8.0, Navidrome 0.63.2, 27,860 tracks.

## 1. The gate almost never opens

`musicbrainz.ts:131-135`:

```ts
): boolean {
  if (t.isCompilation !== true) return false;
  if (t.originalYear != null) return false;
  return reEnrich || !t.originalYearCheckedAt;
}
```

On my library:

| | count |
|---|---|
| tracks | 27,860 |
| `is_compilation = 1` | **37** |
| `is_compilation = 0` | 27,823 |
| `original_year_source = 'musicbrainz'` | **27** |
| `original_year_source = 'album-tag'` | 19,487 |
| `original_year IS NULL` | 8,346 |
| `original_year <> year` | 995 |

The MusicBrainz recording lookup — the part that actually recovers a true original year —
has run on 27 tracks out of 27,860. Everything else took its `original_year` from the
album tag, which on a reissue *is* the reissue year, so `original_year = year` for 18,492
of them.

## 2. Navidrome does not flag these as compilations

`getAlbum` for a Light in the Attic anthology of 1964–65 Stax/Volt singles:

```json
"name": "After Laughter Comes Tears: Complete Stax & Volt Singles + Rarities 1964–65",
"year": 2012,
"isCompilation": false,
"originalReleaseDate": { "year": 2012, "month": 2, "day": 7 },
"releaseDate":         { "year": 2012, "month": 2, "day": 7 },
"recordLabels": [ { "name": "Light in the Attic Records" } ]
```

Same signature on a second one (Allen Toussaint, *The Atco/Atlantic Singles 1968–1974*):
`year: 2015`, `isCompilation: false`, `originalReleaseDate: 2015`.

Note `originalReleaseDate == releaseDate == year`. This is the release *group's* first
release date, i.e. the reissue's own date — not the recordings'. `@brandenespinoza`'s point
in #842 holds one level up as well: many anthologies carry no `COMPILATION` tag either, so
keying on that flag misses exactly the records the issue was about.

## 3. Two consequences

**(a) `original_year` is populated with the wrong number.** `tag-library/flags.ts:116-119`:

```ts
      // compilation's original date is the compilation's own, so its tracks
      // stay unresolved here and phase-0 asks MusicBrainz per track instead.
      originalYear: song.albumIsCompilation ? null : song.albumOriginalYear ?? null,
      isCompilation: song.albumIsCompilation ?? null,
```

With `albumIsCompilation` false, the album's `originalReleaseDate.year` (2012) is copied in
as the track's original year, and the comment's intended path is never taken.

**(b) The "leave it out rather than play it in the wrong decade" rule never fires.**
`show-filter.ts:242-252`:

```ts
export function resolveEraYear(
  year: number | string | null | undefined,
  originalYear: number | null | undefined,
  isCompilation: boolean | null | undefined,
): number | null {
  const oy = Number(originalYear);
  if (Number.isFinite(oy) && oy > 0) return oy;
  if (isCompilation) return null;
  const y = Number(year);
  return Number.isFinite(y) && y > 0 ? y : null;
}
```

`originalYear` is 2012 (not null), so it returns 2012 on the first branch. The `isCompilation`
guard on line 249 is unreachable for these albums. A 1964 Stax single is treated as 2010s
material: it is skipped by a sixties show and eligible for a 2010s one — the exact behaviour
#842 set out to fix. The same resolved year feeds the `Era:` line in the embedding text, so
the sound-similarity index inherits it.

## 4. Independently: three surfaces bypass the resolution

These read `song.year` directly and never consult `resolveEraYear`:

- `subsonic.ts:878` — the Liquidsoap annotation:
  ```ts
  if (song.year) fields.push(`year="${escAnnotate(song.year)}"`);
  ```
- `llm/internal/prompts/scripts.ts:95` — the DJ's intro context, so the DJ can say the
  reissue year on air.
- `llm/internal/tools/picker/slim.ts:18` — `year: s.year || null`, what the picker sees.

So fixing §1–3 alone would correct era filtering but still leave the DJ announcing "2012"
for a 1964 record.

## 5. The file tags cannot be used as a fallback

`ffprobe` on the track in question:

```
TAG:DATE=2012-02-07
TAG:ORIGINALDATE=2012-02-07
TAG:ORIGINALYEAR=2012
TAG:MUSICBRAINZ_RELEASEGROUPID=f2e3047f-5f23-4763-bd55-86cec9c2f3cf
TAG:LABEL=Light in the Attic Records
```

`ORIGINALDATE`/`ORIGINALYEAR` equal the reissue date. Whatever tagger wrote these used the
release group's first-release-date. There is no better field on disk to read.

## Suggested direction

1. **Stop treating `isCompilation` as the only trigger.** A cheap additional signal, available
   without any extra request: treat an album as unresolved when
   `originalReleaseDate == releaseDate == year` *and* the album credits more than one distinct
   track artist. Both example albums match. A stricter variant is to compare the release
   group's first-release-date against the release date via the MBIDs already in the tags.
2. **Do not let the album tag masquerade as a resolved original year.** Consider recording
   `original_year_source = 'album-tag'` as provisional, so `resolveEraYear` can decline it and
   fall through to the "leave it out" branch rather than asserting a reissue year. Right now
   `album-tag` and `musicbrainz` are indistinguishable to every consumer.
3. **Route the annotation, the DJ prompt and the picker `year` through `resolveEraYear`**, so a
   resolved original year reaches the surfaces the listener actually perceives.
4. **A manual per-album override** would cover the long tail either way. `@brandenespinoza`'s
   LLM-tagged original years can't be read over Subsonic, and an override would let that work
   count for something.

Happy to run any query against my library — 27,860 tracks, heavy on catalogue soul, reggae,
funk and blues, so it exercises this hard.

perminder-klair:
Closes #1418 — all four parts.

The station announced, displayed and filed a 1964 Stax single as a 2012 record. Three defects stacked:

1. **The lookup almost never ran.** Gated on Navidrome's `isCompilation` flag, which reissue anthologies do not set — 27 tracks out of 27,860 on the reporter's library.
2. **The album tag masqueraded as a resolved answer.** The walk recorded `originalReleaseDate.year` even when it merely echoed the release year (18,492 rows), which both stated the wrong year confidently and hid the track from the lookup that could fix it.
3. **Four surfaces ignored the resolution anyway**, so even a correct original year never reached the listener.

Three commits, readable in order.

---

## 1. Route the listener-facing surfaces (`1f66ed1e`)

| Surface | Drives |
| --- | --- |
| `subsonic.getAnnotatedUri` | the Liquidsoap `annotate:` metadata |
| `llm/internal/prompts/scripts.ts` | the DJ's "Now starting:" line — what the station **says** |
| `picker/slim.ts` | the year the picker reasons with |
| `routes/public.ts` `/now-playing` | every skin's year display |

The last one isn't in the issue. `now-playing.json` carries no year — `radio.liq` writes title/artist/album/id only — so that line is the *sole* source of the year all five skins render. Listeners were seeing the wrong year as well as hearing it.

**Behaviour change:** an unresolved compilation now omits the year rather than asserting the reissue's. Same "leave it out rather than play it in the wrong decade" rule era filtering already follows. Admin surfaces keep the raw `year` deliberately — the operator is looking at the file.

## 2. An operator override (`43f4db42`)

There was previously **no way to correct a track at all**. `routes/library.ts:788` gates the per-track retag on `needsOriginalYearLookup(t, true)`, and `reEnrich=true` only bypasses the *checked-at* stamp, not the compilation gate — so bulk and manual were both dead on an unflagged anthology.

`POST /library/original-year` writes `original_year_source = 'manual'`, optionally album-wide (the common case — an anthology is wrong an album at a time). Surfaced as an `original year · era` field in the library row editor, which now also says **where the current answer came from**; `album-tag` is labelled as weak, since on a reissue it is just the reissue again.

The load-bearing part is that **both automatic writers must refuse to touch a manual row** — the walk's upsert CASE and `setOriginalYear`'s WHERE. Every walk re-visits every track, so an override the walk can clobber is one the next rescan silently undoes.

`originalYear: null` **clears** rather than pinning "unknown": the track re-enters the automatic pipeline, so "I was wrong about this one" is recoverable without a library reset.

Deliberately does not invalidate the embedding — the `Era:` line is stale until the next tag pass, but dropping the vector to force it would pull the track out of similarity search entirely, a bigger loss than one wrong decade word beside genre, tags, lyrics and acoustics. Era *filtering* reads the column and is correct at once.

## 3. Trigger on anthologies Navidrome doesn't flag (`4cd0cbf1`)

`music/era-suspect.ts` replaces the flag as the trigger. A policy module, not a branch in the walk, because the same judgement is now reached from three places.

Tuned **precision-first** — every signal is an anthology *marker*, never a weak correlate:

- Navidrome's flag (unchanged, still first)
- a Various-Artists album artist
- **three or more** distinct credited artists — two is a duo, a split, or any features-heavy record, and fires constantly
- a date range in the title that **closes before** the album's year — the only signal that catches a single-artist anthology like *The Atco/Atlantic Singles 1968–1974* on a 2015 release

A false positive costs a 1 req/s MusicBrainz call and can drop a good album out of era shows; a false negative just leaves today's behaviour. So the negative cases are what the tests defend. The walk logs the blast radius by reason — an operator whose library shifts should read it in the log, not infer it from a show that stopped picking.

**Three columns, three jobs.** `is_compilation` stays the raw Navidrome fact; `era_untrusted` is the derived verdict; `yearUntrusted` (composed once in the row mappers) is the only one era resolution reads. `resolveEraYear`'s third argument is that composed value, and `needsOriginalYearLookup` deliberately refuses the raw flag so no call site can quietly reopen the gap. Both NULL = not yet re-walked = pre-#1418 behaviour exactly. `browse.ERA_YEAR_SQL` and `idsNeedingOriginalYear` carry the same OR.

**Migration 21** clears the echoed album-tag years. That UPDATE is the one statement here that deletes operator data, so it's extracted as `clearEchoedAlbumTagYears` and tested in both directions: it spares a `musicbrainz` or `manual` value that merely coincides with the file year, and spares any `album-tag` year that *differs* — those carried real reissue information (the 995-row case). Clearing the echoes is behaviour-neutral by construction: a trusted album falls through to the identical `year`.

---

## Note on the issue's suggestion 2

Making `resolveEraYear` decline an `album-tag` year does not work as described — it falls through to `if (isCompilation) return null`, which is `false` for these albums, so it lands on the same reissue year. It's a no-op for the reported case and would discard correct values on the 995 rows where the two differ. The idea underneath it was right, but it belongs at the **walk** (don't record an uninformative tag) rather than at the resolver, which is where §2 above puts it.

## Verification

- `npm run lint` clean in both `controller/` and `web/` (eslint + `tsc --noEmit`), schema mirror regenerated.
- New: `scripts/era-suspect.test.ts` (18, pure), `scripts/manual-original-year.test.ts` (17, real SQLite). `scripts/original-year.test.ts` updated to the new gate contract.
- Full suite: **725 pass, 4 fail** — `aio-log-link`, `analyzer-python`, `state-bootstrap`, `state-tree`, all confirmed failing identically on a stashed clean tree (no state dir / no python analyzer in this worktree).

## Not done

The batched-MBID idea from the issue thread (`rid:(a OR b OR c)`, one request per anthology instead of one per track) is a worthwhile follow-up for large libraries but is a change to the lookup's transport, not its gating — separate PR.


perminder-klair:
Verified end-to-end against an isolated stack (controller on :7791 with a temp `STATE_DIR`, worktree Next dev server on :7793, Playwright) — the live station was not touched.

**Fixtures**, built through the real writers (`albumEraSuspect` → `upsertTrackMeta`), not hand-written rows:

| track | album | file year | outcome |
| --- | --- | --- | --- |
| `anth1-3` | the reported Stax/Volt anthology (VA, 8 artists, `1964–65` in the title) | 2012 | suspect → `various-artists`, album tag **not** recorded |
| `ord1` | *In Rainbows* | 2007 | not suspect, echoed tag suppressed |
| `reissue1` | *Maggot Brain (2020 Remaster)* | 2020 | album tag **kept** — it genuinely knew 1971 |

**26 API checks.** The round trip that matters most is SQL-side era filtering, which exercises `ERA_YEAR_SQL`:

- 1970s show finds `reissue1` by its resolved **1971**, not its 2020 file year
- 2020s show does **not** include it despite `year=2020`
- 2010s show does **not** mis-include the anthology — the original #842 bug
- after a 1964 override the anthology appears in a 1960s show, and disappears again when cleared

Plus validation (non-integer, out-of-window, garbage string, blank id, unknown id → 404, numeric string from an `<input>` accepted), that the override leaves `is_compilation` and `era_untrusted` untouched, that siblings are unaffected without `applyToAlbum`, and that both routes 401 unauthenticated.

**`/now-playing`, the surface not named in the issue** — all four cases correct: ordinary `2007`, reissue **`1971`**, overridden anthology `1964`, unresolved anthology **absent** rather than `2012`.

**Annotation + DJ line via a real library lookup** (bare Subsonic-shaped song, no inline era fields — the actual request/auto-pick path): reissue annotates `year="1971"`, unresolved anthology emits **no** year field, off-library track still falls back to its plain year.

**18 Playwright checks** on the row editor: the source note distinguishes all four provenances, the input is empty rather than prefilled (prefilling would invite "confirming" a number nobody checked, laundering it into a manual answer that outranks every later fix), Set year stays disabled on a partial year, the row updates without a refetch, and clear restores the unresolved wording.

## Two defects this found, both now fixed

1. **The helper text was effectively invisible** (`9a02b12a`). I had used `text-ink-soft`, which is `color-mix(in oklab, var(--ink) 6%, transparent)` — a *surface* token. `ManualTagEditor` was the only file in `web/` using it as a text colour, which was the tell. Now on the house helper classes, with a capped measure since it was setting as one unbroken ~200-character line across the full table width.
2. **Inconsistent apostrophes** — two branches used straight ones against typographic elsewhere in the component.

Neither would have shown up in lint or the unit tests — only in a browser.

## Suite

`npm run lint` clean in both packages. **725 pass / 4 fail** — `aio-log-link`, `analyzer-python`, `state-bootstrap`, `state-tree`, all confirmed failing identically on a stashed clean tree (no state dir, no python analyzer in this worktree).

Still unverified, and honestly so: the detector's precision against a **real** large catalogue. The thresholds are reasoned, not measured. The walk logs the blast radius by reason so it is visible on first run, and I have asked the reporter for those numbers.

perminder-klair:
## Measured the detector against a real catalogue — and it was badly mistuned

I said the thresholds were "reasoned, not measured" and left it there. Measuring it found a serious problem, so this is a correction to `4cd0cbf1`, in `82a8c3b6`.

Ran the shipped `albumEraSuspect` read-only over a real **9,216-track / 3,989-album** library (modern Punjabi and rap heavy):

| | albums | tracks | share of library |
| --- | --- | --- | --- |
| **as shipped** (`>=3 distinct artists`) | 207 | 2,221 | **24.1%** |
| **retuned** | 28 | 548 | **5.9%** |

The 24% was not a near-miss. The flagged sample was `Bangarang`, `Because of You`, `Bhram (Deluxe)`, `AURA`, `BULLY` — ordinary albums with guest verses. On this kind of catalogue "three or more credited artists" is simply what an album looks like.

And the consequence is worse than a wasted lookup: a false positive marks the year untrusted, so every one whose MusicBrainz lookup misses **drops out of era-bounded shows entirely**. The previous tuning would have quietly emptied a quarter of that library out of decade shows.

### Three findings, each now a test

1. **`artist` is the raw per-track string and carries the features.** `Slauson Boy 2` counted 13 distinct artists, `Mr. Morale & The Big Steppers` 9, `Moosetape` 10 — all single-artist records. Counting `artistRootKey` (the lead) collapses them to 1. Biggest single correctness win, and it reuses the project's existing lead-artist primitive rather than duplicating its split rules.
2. **No artist-count threshold separates the two populations** — the distribution is continuous, 132 of 207 albums sat in the middle band. What separates them is whether anyone is at the *front* of the record: an anthology has no dominant lead, an ordinary album always does.
3. **Small releases break any ratio test.** A 4-track EP with four collaborators has no dominant artist by arithmetic alone, and the catalogue is full of them (`Nasha - Single`, `Say Less - EP`).

So the rule is now: **≥5 distinct LEAD artists, ≥8 tracks, top lead under 30%**, plus a new `title-compilation` signal (short, specific word list) that catches the single-artist "Best of" no artist-count signal can see. `albumEraSuspect` now takes raw `trackArtists` and does the lead normalisation itself, so the walk cannot get it wrong on the way in.

What it flags now, in full: 7 DJ mixes, 3 soundtracks (`TRON: Legacy Reconfigured`, `Black Panther: The Album`, `Dhurandhar`), various-artist collections (`Dharma Sounds Of Summer`, `Today's Hits 2021`, `56 Districts`, `NXTAPE Vol. 1`), and 4 title-flagged collections. Both albums from the issue still fire.

### Migration 21 verified on the same library

Populated to its realistic pre-#1418 state (9,083 album-tag rows: 8,223 echoing the file year, 860 carrying real reissue info, 40 musicbrainz):

- v16 → v21 in **279ms** over 9,216 rows
- all 8,223 echoed years cleared, **0 remaining**
- all **860** genuine reissue values kept, all **40** musicbrainz values kept
- idempotent on a second `open()`

### Suite

Lint clean both packages. **730 pass / 4 fail** — the same four environmental failures, confirmed identical on a stashed clean tree.

The `various-artists` signal is still unmeasured here: `library.db` has no `album_artist` column, so it could not be exercised offline. It is the highest-precision signal of the set and only adds recall, so the 5.9% figure is a lower bound on what the detector catches — but it is worth watching in the walk log on a first real run.

PR #1433, opened 2026-08-19
fix(broadcast): space artists out on the agent path, and fold name variants (#1406)

### LINKED ISSUE #1406: Artist spacing: no artist-recency on the agent path, and artistRootKey misses name variants (The X / X Experience)
### Summary

Two related gaps let the same artist dominate a stretch of air on the agent pick path (observed on 1.8.0):

1. **No artist-level spacing beyond the on-air track.** The agent path deliberately has no artist-recency filter (`llm/internal/tools/picker/index.ts` explains an artist strip once gutted the similarity tools). The only artist protection at the point of choice is the back-to-back guard (`dj-agent.ts` ~371), whose memory is `ARTIST_VARIETY_WINDOW = 5` slots and which only fires when the pick equals the **currently playing** artist. Result on my station: the same artist aired 3 times in one morning show (and had featured heavily the previous evening) — every occurrence legal, each 3+ slots apart. The 6h `recentArtistsSince` window exists but is only consulted by the fallback pool picker.

2. **`artistRootKey` misses common name variants.** It strips `feat./ft./featuring` tails and a second credited act after `&/+/and`, but does not strip a leading "The", ensemble suffixes, or fold punctuation. Concrete bypass: "The Jimi Hendrix Experience" followed immediately by "Jimi Hendrix" — the back-to-back guard never fired because the root keys differ. The same comparison shape backs `recentArtistsSince` and `neighbourArtistRoots`, so every artist-level mechanism shares the blindness.

### Suggested changes

- An operator-tunable artist-spacing setting (e.g. `llm.artistRecencyHours` or an artist-count window) enforced **at the point of choice** — extend the existing guard's `recentRoots` from 5 slots to a configurable N, or feed `recentArtistsSince(hours)` roots into the guard's `alternativeCandidates` flow. Keeping it at choice-time (not candidate-filtering) preserves the similarity tools, which was the reason the artist strip was removed.
- Normalize harder in `artistRootKey`: leading "the ", common ensemble suffixes ("experience", "band", "orchestra", "trio", "quartet"), and punctuation folding (curly vs straight apostrophes) — or alias via MusicBrainz artist ids where tags carry them. The queue's raw `title|artist` keys can stay byte-exact as documented; this is matching-side only.
- Exposing the derived recency windows (`trackHours`/`artistHours` in `music/recency.ts`) as settings would also help large-library operators tune spacing without code.

For what it's worth, `llm.noRepeatWindow` works as documented — raising it helped track-level repetition, but nothing operator-facing touches artist spacing today.

perminder-klair:
Closes #1406.

Two gaps let the same artist dominate a stretch of air on the agent pick path, every occurrence legal.

## 1. The guard had no memory beyond the on-air track

`dj-agent.ts` fired the artist guard only on `artistRootKey(song) === curArtist`. The recency window was passed one line lower into `alternativeCandidates`, so it narrowed the **re-pick pool** but never the **entry condition** — a pick three slots after the same artist was never examined and the window never consulted. The reporter saw one artist three times in a morning show, each occurrence 3+ slots apart and all of them legal.

`artistGuardCause()` (in the existing policy module, not inlined at the call site) now names **which** of two causes fired, and they escalate differently on purpose:

| Cause | Meaning | Escalation |
| --- | --- | --- |
| `'onair'` | back-to-back repeat | unchanged — re-pick → `blockedArtists` pool rescue → relax (the full #1187 cascade) |
| `'recent'` | inside the spacing window | **soft** — re-pick from fresher candidates only; if none exist, or the re-pick doesn't land, the original pick stands |

The soft path is the load-bearing design call. Running the full cascade on every spacing miss would add a re-pick call plus a pool rescue to a large share of transitions, and on the default homelab Ollama that is real latency spent chasing a preference. Spacing never costs the station a slot, and never adds a second model call.

## 2. `artistRootKey` handles safe name variants

"The Jimi Hendrix Experience" straight into "Jimi Hendrix" walked past the guard because the root keys differed. It now folds a leading article, curly-vs-straight apostrophes, whitespace runs, and a small set of exact full-name aliases verified as the same act.

A review caught that stripping ensemble words generically was unsafe: "The Beta Band", "Manchester Orchestra", "Unknown Mortal Orchestra", and "Kronos Quartet" are complete act names, not a lead artist plus decoration. Because `blockedArtists` in `filterPickerCandidates` is a hard drop, collapsing those onto Beta/Manchester/Unknown Mortal/Kronos could discard a legitimate rescue candidate. The implementation now keeps those names intact and maps only exact verified aliases (including the reported Hendrix pair, Glenn Miller Orchestra, Dave Matthews Band, and Bill Evans Trio).

Article/alias handling still runs after the join split, so the `X & the Y` exception sees the tail it protects ("Sly & the Family Stone" stays whole).

## 3. Operator dial

`settings.llm.artistVarietyWindow` — slots, 0–25, default 5 (the figure the guard already used internally for its re-pick pool, so an upgrade turns spacing on rather than silently keeping the behaviour this was filed against). Admin UI control sits under the no-repeat window in the LLM section.

**0 disables spacing only.** Back-to-back is not operator-disableable — an artist following itself is a fault, not a taste.

## One correction to the issue's diagnosis

The issue notes that "the same comparison shape backs `recentArtistsSince` and `neighbourArtistRoots`". It doesn't: `queue.recentArtistsSince` builds **raw** lowercase artist strings for the pool picker's relaxable filter, matched raw against raw (documented at `queue.ts:2426`). It's blinder still — it doesn't collapse collaborations either — and the normalisation here does not reach it. Documented in `docs/internals/music.md` rather than changed, since altering it is a behaviour change to the pool path and belongs in its own PR.

The issue's third suggestion — exposing `trackHours`/`artistHours` as settings — is also left out deliberately. Those feed `recencyWindowsForLibrary`'s library-size scaling, so an operator override first needs a decision about whether it replaces or multiplies the boost.

## Verification

- `controller` and `web`: `npm run lint` (eslint + `tsc --noEmit`) clean, 0 errors.
- New `scripts/llm-artist-variety-window.test.ts` — **cold-load** round trip per the settings rule (an in-process assertion passes on the broken code), plus clamping, `0`-is-not-absent, and the absent-key upgrade default.
- `scripts/artist-guard.test.ts` extended with the variant table and the full cause matrix (including "an empty window still guards back-to-back").
- Full suite 686/691. The 5 failures (`aio-log-link`, `analyzer-python`, `state-bootstrap`, `state-tree`) reproduce on the base commit — they're environmental to the worktree used here (no `state/`, no AIO scripts, no python analyzer), not introduced by this change.


perminder-klair:
## Follow-up: the wiring is now tested too

The first push left one honest gap — the two mechanisms were pinned, but their **composition** wasn't. The guard's decisions already lived in `artist-guard.ts`; how they compose sat inline in `pickViaAgent` behind a model call, so the part that actually differs between the two causes was verifiable only by reading it.

**`runArtistGuard()`** now takes that block with the two expensive calls injected — `repick` (the constrained `djObject` over the alternatives) and `poolRescue` (`pickViaPool` with `avoidArtist`). `pickViaAgent` keeps every queue and settings read and hands the module values only, getting back a discriminated outcome: `'none'` / `'kept'` / `'repicked'` / `'rescued'` (rescued = slot filled, caller stops). Behaviour-preserving — same order, same log lines, same events, same relaxation reasons; the pre-existing pure tests pass unchanged.

`scripts/artist-guard-run.test.ts` — 14 cases covering every path. The load-bearing ones **count the injected calls**, because the cost guarantee is the thing that would silently regress if the two causes were ever merged back together:

- `spacing yields when every alternative is also recent` → asserts `{repick: 0, poolRescue: 0}`
- `a failed spacing re-pick keeps the pick and stops` → asserts `repick: 1, poolRescue: 0`
- `a failed back-to-back re-pick still escalates to the pool` → asserts `repick: 1, poolRescue: 1`

Plus: an unguarded pick costs nothing and logs nothing; a `'queued'` rescue reports the slot filled; a re-pick answering with an id it wasn't offered is refused; a name variant of the on-air act is caught as back-to-back; and `window: 0` disables spacing while back-to-back still fires.

## Admin UI, end to end

Driven with Playwright against an **isolated** controller (`:7791`, throwaway `STATE_DIR`) and a worktree Next dev server (`:7793`) — never a live station. Asserted through the controller API after Save, not via toasts:

- field renders in the LLM section under Next-track picker, below the no-repeat window
- loads the controller's current value (pre-seeded to 9 so this is a real assertion, not a default)
- is a `0`–`25` number input, hint states back-to-back stays on at `0`
- saving 12 reaches the controller; survives a reload
- **an over-max value is clamped server-side, not just by the input** — typed 99, `settings.llm.artistVarietyWindow` reads 25

Also checked directly against the API: default `5`, `12 → 12`, `99 → 25`, `-4 → 0`, `7.8 → 7`, persisted to `settings.json`, and **still 12 after a real controller restart** (the cold-load path the settings rule exists for).

## Current state

- `controller` + `web` lint (`eslint` + `tsc --noEmit`): 0 errors
- suite 701/705 — the failures are `aio-log-link`, `analyzer-python`, `state-bootstrap`, which reproduce on the base commit (environmental to the worktree: no `state/`, no AIO scripts, no python analyzer)

Issue #1417, opened 2026-08-16
BPM is double-time on slow material: beat_track runs with no octave correction, and analysis_confidence scores the wrong tempo as fully plausible


spaceshipmike:
`analyze_worker.py` calls `librosa.beat.beat_track` with no tuning arguments and stores the
result unmodified. On slow material librosa's default tempo prior doubles the tempo, and
nothing downstream corrects it. The stored value then drives real crossfade timing, so this
is audible, not cosmetic.

SUB/WAVE 1.8.0, `subwave-analyzer-heavy`, 27,860 analysed tracks.

## Symptom

Curtis Mayfield — "Never Say You Can't Survive" (1977), a slow ballad, reported as **152.0 BPM**
(`analysis_confidence = 0.613`, `analysis_version = 6`). True tempo is about 76.

Not isolated. Songs whose tempo is not in dispute:

| track | reported BPM |
|---|---|
| 'Round Midnight — Thelonious Monk | 143.6 |
| 'Round Midnight — Miles Davis | 143.6 |
| A Change Is Gonna Come — "Brother" Jack McDuff | 152.0 |
| Summertime — Al Green | 143.6 |
| Summertime — The Zombies | 172.3 |
| Summertime — Booker T. & the MG's | 161.5 |

Library-wide, on a collection that is mostly soul, reggae, funk, blues and jazz:

- 7,054 of 27,860 tracks (25%) are stored above 140 BPM
- 2,126 above 170 BPM
- **4,984 tracks are above 140 BPM while tagged `energy` low or medium** — the analyser
  contradicting itself in the direction a doubling error predicts

## The stored values sit on librosa's tempogram grid

The 20 most common BPM values in the library are all exactly `60 * 22050 / 512 / n`:

```
lag 14 -> 184.6    lag 20 -> 129.2    lag 26 ->  99.4    lag 32 ->  80.7
lag 15 -> 172.3    lag 21 -> 123.0    lag 27 ->  95.7    lag 33 ->  78.3
lag 16 -> 161.5    lag 22 -> 117.5    lag 28 ->  92.3    lag 34 ->  76.0
lag 17 -> 152.0    lag 23 -> 112.3    lag 29 ->  89.1
lag 18 -> 143.6    lag 24 -> 107.7    lag 30 ->  86.1
lag 19 -> 136.0    lag 25 -> 103.4    lag 31 ->  83.4
```

[...]

PR #1434, opened 2026-08-19
fix(mix): make tempo-derived timing octave-safe (#1417)

perminder-klair:
Partially addresses #1417.

## What changed

Real-audio validation showed the proposed onset-strength correction did not separate known octave errors from backbeat controls. This revision removes that analyzer heuristic, the confidence/schema changes, and the version bumps that would have forced a library-wide reanalysis. Stored and displayed BPM remain raw.

Instead, `music/mix.ts:timingBpm` folds readings at or above 110 BPM down by octaves when deriving single-track durations. A genuinely fast track stays aligned at half-time; a genuine slow reading is never multiplied into a half-bar.

The shared timing fold covers:

- adaptive and ending crossfade bar snaps
- washout canvas and delay
- loop length and loop crossfade
- chop period

**Every effect period reaches its audible window by halving, never by clamping.** This PR's own first cut clamped *after* folding, which was wrong: `chopPeriodFor`'s `Math.min(0.75, 60/folded)` binds for every folded pulse under 80 BPM — that is every raw reading from 110 to 160, the commonest band in a library. All of them returned a flat 0.75 s gate: 1.38 beats at 110, 1.50 at 120, 1.60 at 128, 1.90 at 152. A gate that is not a whole multiple of the beat walks the grid and cuts mid-note, which is exactly what the function's own contract rules out. `washoutDelayFor`'s 0.45 ceiling bound for every raw reading from 110 to 200 plus everything under 100, so the comb tap stopped being a dotted eighth of anything.

Both now descend the same dyadic ladder `loopBarFor` already used, landing on the single dyadic point in the window. That makes `chopPeriodFor`/`washoutDelayFor`/`loopBarFor` octave-invariant at **any** tempo — they no longer depend on the 110 threshold at all. Only the bar-snapped canvases do, which is why a double landing below the line (a true 52 read as 104) is left alone rather than guessed at.

## Analyzer regression

The earlier shared-onset refactor had converted the pace envelope to a list while `estimate_pace()` still expected ndarray slices with `.size`, which made pace analysis return `None` and would have cleared existing pace curves during reanalysis. That failed refactor and its artifacts are removed from this branch; no Python remains in the diff.

## Tests

`scripts/mix-fx.test.ts` previously asserted only `f(x) == f(2x)`. That is necessary but not sufficient — two equally off-grid values compare equal just as happily as two aligned ones, which is precisely how the clamp-after-fold regression would have shipped green.

It now asserts the property the doc comments actually promise: each stamped period is the track's beat (or its dotted eighth, or its bar) scaled by a power of two, so it lands on the grid at some octave. The unit is derived from the **raw** BPM, so the assertion cannot be satisfied by mirroring whatever `timingBpm` does. Swept over 30–300 BPM in 0.1 steps; the pre-fix implementation fails it at **4,140 of 5,402 points**.

Two smaller pins: the effect periods are octave-invariant below the 110 line too (the #1417 corpus has tracks at 41.7, 55 and 56.2 BPM whose doubles land under it), and the 110–160 chop band keeps a one-beat gate, named as the regression it is.

## Scope

This is the safe, immediately useful part of #1417. The issue remains open for a tempo estimator validated against real audio, and for any later correction of stored BPM, playlist filtering, display values, or embedding text. This PR deliberately triggers no reanalysis or database migration.

The halve-never-clamp rule is now recorded in `CLAUDE.md` and `docs/internals/music.md` with the measured failure, so it is not rediscovered the hard way.

## Verification

- `npm test -- mix-fx` — pass
- `npm run lint` (controller) — exit 0, 0 errors (653 pre-existing warnings); `tsc --noEmit` clean
- full `npm test` (controller) — 796 pass, 1 fail. The failure is `analyzer-python.test.ts` → `ModuleNotFoundError: No module named 'numpy'`, a missing Python sidecar dependency on this host; it fails identically on `develop` and this diff touches no Python.
- `git diff --check` — pass

The branch is merged with the latest `develop`.


spaceshipmike:
Tested against real audio on my library. **Please hold this one — it does not correct anything yet, and there are two separate reasons.** The first is a two-line crash; the second is the part that needs your judgement.

Setup: 29 tracks, the 7 named in #1417 plus controls I picked before running anything — reggae one-drop, fast punk with a hard backbeat, mid-tempo funk that already reads correctly, and very slow readings where `double` could fire. `subwave-analyzer-heavy:latest`, librosa 0.11.0, numpy 2.4.6, same 40 s window and `hop_length=512` as `analyze()`. I called your `onset_envelope` / `tempo_octave_fit` / `apply_tempo_octave` directly rather than reimplementing them.

Control first: `beat_track` reproduced the stored value on **29/29** tracks, to the decimal. So the numbers below are measuring your code, not a decode difference.

## 1. `beat_frames or []` raises on librosa's output, and the failure is silent

`librosa.beat.beat_track` returns an **ndarray**. `tempo_octave_fit` line 983 and `apply_tempo_octave` line 1069 both do `(beat_frames or [])`, and `ndarray.__bool__` raises on more than one element:

```
ValueError: The truth value of an array with more than one element is ambiguous.
```

`corrected_tempo` catches it, so nothing crashes in production — it logs `tempo octave fit failed` and returns the tempo unchanged with `bpm_confidence=None`. Run against the Mayfield track exactly as `analyze()` calls it:

```
beat_frames type: ndarray
old_bpm=152.0  new_bpm=152.0  bpm_confidence=None
[analyze-worker] tempo octave fit failed: The truth value of an array with more than one element is ambiguous.
```

So on a default install the feature is **inert on every track**: same wrong BPM, and `bpm_confidence` NULL, which the PR documents as "not judged — never a low score". It reads as a per-track degrade rather than total failure. `analyzer_tempo_test.py` passes plain lists, which is why it stays green.

`[] if beat_frames is None else list(beat_frames)` at both sites fixes it. `analyze_outro` reaches the same path via line 1186.

## 2. With the crash patched, the halve branch still never fires — the `weak <= mean_env` guard blocks it

I patched those two lines locally and re-ran all 29. Result: **`action` was `keep` on 29/29.** Zero corrections, including all 7 tracks from the issue.

The cause is `alt_decisive`. It requires `weak <= mean_env * OCTAVE_WEAK_MAX_MEAN` (1.0). Measured on these 29 tracks:

```
weak / mean_env    min 2.13    median 3.53    max 9.88     passes: 0/29
```

The minimum is more than double the threshold, so this is not a tuning miss. `mean_env` is the mean over **all** frames, and most frames sit between onsets. Any beat position — including the weaker half of an alternating pair — is well above that mean by construction. The comparison is between two different populations, so on real music the guard is close to unconditional, and `OCTAVE_ALT_MAX` never gets consulted. Both octave tests are currently dead code.

I think the intent was "weak is at background level", but the background reference wants to be the inter-onset floor, not the whole-envelope mean — something like the median of frames away from any beat, or a low percentile of the envelope.

## 3. The part I would not have caught without the controls: `alt` barely separates the two classes here

This is the answer to your "did anything get halved that shouldn't". Sweeping `OCTAVE_ALT_MAX` with the guard removed, against labels I fixed **before** running:

```
alt<=0.60 : fires  1/29 | catches 0/7 issue tracks | halves 1 must-keep
alt<=0.75 : fires  7/29 | catches 3/7             | halves 2 must-keep
alt<=0.80 : fires 11/29 | catches 4/7             | halves 4 must-keep
alt<=0.976: fires 25/29 | catches 7/7             | halves 9 must-keep
```

The 7 issue tracks span `alt` 0.687–0.976, and the must-keep controls span 0.596–0.998. The ranges sit almost on top of each other.

The sharpest single data point: at the **shipped** 0.60 threshold, with the guard removed, exactly one track in 29 corrects — Sly & the Family Stone, "Love City", **123.0 → 61.5**. It has the lowest `alt` in the whole set (0.596), and it is a track that should not move: `beat_track` says 123.0 and a prior-free tempogram peak says 123.0. So the one correction the shipped numbers would make is a false positive, and none of the true positives are anywhere near firing.

That is the failure mode you said you were least able to rule out, and on this material it is not an edge case — it is the leading edge of the distribution.

## Per-track numbers

`alt` and `off` as `tempo_octave_fit` computed them. "expect" was written down before the run.

| group | track | stored | alt | off | expect |
|---|---|---|---|---|---|
| #1417 | Al Green — Summertime | 143.6 | 0.687 | 0.173 | halve |
| #1417 | Curtis Mayfield — Never Say You Can't Survive | 152.0 | 0.692 | 0.319 | halve |
| #1417 | Jack McDuff — A Change Is Gonna Come | 152.0 | 0.733 | 0.332 | halve |
| #1417 | Miles Davis — 'Round Midnight | 143.6 | 0.796 | 0.537 | halve |
| #1417 | Thelonious Monk — Round Midnight | 143.6 | 0.918 | 0.411 | halve |
| #1417 | The Zombies — Summertime | 172.3 | 0.961 | 0.371 | halve |
| #1417 | Booker T. & the MG's — Summertime | 161.5 | 0.976 | 0.310 | halve |
| one-drop | Toots & The Maytals — Broadway Jungle (dub) | 143.6 | 0.719 | 0.576 | halve |
| one-drop | Ken Boothe — Mr. Wind | 143.6 | 0.741 | 0.821 | halve |
| one-drop | The Heptones — Love Won't Come Easy | 152.0 | 0.900 | 0.266 | ? |
| one-drop | Bob Marley — Mix Up, Mix Up | 172.3 | 0.895 | 0.218 | ? |
| one-drop | Augustus Pablo — Memories of the Ghetto | 143.6 | 0.993 | 0.404 | halve |
| one-drop | The Police — Walking on the Moon (live) | 152.0 | 0.998 | 0.404 | halve |
| fast+backbeat | Minor Threat — Steppin' Stone | 184.6 | 0.800 | 0.630 | keep |
| fast+backbeat | Fugazi — Argument | 215.3 | 0.816 | 0.505 | ? |
| fast+backbeat | Fugazi — Strangelight | 184.6 | 0.826 | 0.292 | keep |
| fast+backbeat | Minutemen — One Reporter's Opinion | 184.6 | 0.838 | 0.562 | keep |
| fast+backbeat | The Smiths — Please, Please, Please… | 184.6 | 0.880 | 0.623 | halve |
| fast+backbeat | Fear — Responsibility (rough mix) | 184.6 | 0.913 | 0.709 | keep |
| mid-tempo OK | **Sly & the Family Stone — Love City** | 123.0 | **0.596** | 0.743 | **keep** |
| mid-tempo OK | Yeah Yeah Yeahs — Zero | 123.0 | 0.871 | 0.649 | keep |
| mid-tempo OK | James Brown — Papa's Got a Brand New Bag | 117.5 | 0.913 | 0.809 | keep |
| mid-tempo OK | Stevie Wonder — Contusion | 123.0 | 0.998 | 0.726 | keep |
| already slow | Thelonious Monk — 'Round Midnight | 63.0 | 0.685 | 0.305 | keep |
| already slow | Thelonious Monk — Round Midnight | 69.8 | 0.779 | 0.390 | keep |
| very slow | Aretha Franklin — Mr. Ugly | 55.0 | 0.755 | 0.393 | ? |
| very slow | Bibio — Light Up The Sky | 56.2 | 0.984 | 0.269 | ? |
| very slow | OutKast — Mamacita | 41.7 | 0.925 | 0.254 | ? |

`off` never reached `OCTAVE_OFF_MIN` (0.85) either — max was 0.821, on Ken Boothe, a track that wants halving rather than doubling.

Two notes on how far to trust the "expect" column. It is my judgement, so treat the sweep in §3 as indicative, not as a score. §1 and §2 do not depend on it at all — the crash and the 0/29 guard result are measurements. As a cross-check I ran `beat_track` a second time with `start_bpm=70` and a prior-free tempogram peak; on the clean cases they agree with halving (Mayfield 152.0 → 73.8, Toots 143.6 → 71.8 on both), but both estimators carry their own octave ambiguity on the punk tracks, so I did not use them as ground truth.

## What might help

The three guards are each reasonable in isolation, but on this material the two measurements underneath them do not carry enough signal to separate a doubled grid from a backbeat. Before re-tuning thresholds, it may be worth checking whether `alt` and `off` can separate the classes **at all** on real audio — on this sample they cannot, and a threshold change alone would buy true positives at roughly one false positive each.

If it would help, I can run any branch over a much larger, genre-stratified sample and hand back the same table — the harness is ~60 lines and reads audio only. I can also dump `alt`/`off` for a few thousand tracks if a distribution would be more useful than 29 rows.

Happy to test again whenever you want another pass. And thank you for the depth of the fix — the beat-grid-travels-with-the-tempo point and the `strongOffset` parity trap were both things I would not have thought to check.

perminder-klair:
Your test changed the direction here. I removed the analyzer octave-correction path completely rather than retuning thresholds that do not separate doubled grids from backbeats. The PR now does one bounded thing: `mix.ts` folds readings at or above 110 BPM down only when deriving transition timing. Stored and displayed BPM, filtering, and embeddings remain untouched. Tests pin half/double invariance across every timing consumer.

I also benchmarked Beat This on 998 usable GTZAN tracks. Neither checkpoint met the 95% precision and 80% coverage gate; the best result was 93.24% precision at 24.67% coverage. The 698-track held-out split was not scored, and none of that estimator work is included here.

There is no analyzer correction left for your 29-track harness to retest. Latest `develop` is merged, the PR is mergeable, and all three CI jobs pass. This is ready as the bounded timing defence, while #1417 should remain open for the underlying stored-BPM problem.

PR #1442, opened 2026-08-22
feat(shows): add matching-track availability diagnostic

Jaz666:
## Summary

- Add an on-demand matching-track count to Edit Show, positioned with excluded playlists.
- Validate the current unsaved show draft before calculating.
- Distinguish the strict selection pool from the soft-filter preference pool.
- Show pinned-playlist and excluded-playlist effects without presenting the broad soft-filter fallback as the useful count.
- Follow the existing Show editor’s operator-facing copy and layout conventions.

## Validation

- Controller focused candidate tests passed.
- Controller type check passed.
- Web lint and type check passed.

## Station assessment

Tested on the live station. The currently on-air show reports 3,727 matching tracks; with Strict filter enabled this is its post-exclusion selection pool, while with it disabled it remains the DJ’s preferred pool.

### DIFF
--- a/controller/scripts/show-candidate-display.test.ts
+++ b/controller/scripts/show-candidate-display.test.ts
@@ -1,0 +1,27 @@
+import assert from 'node:assert/strict';
+import { test } from 'node:test';
+
+const displayModulePath = '../../web/components/admin/shows/' + 'candidate-diagnostic.js';
+const display = await import(displayModulePath).catch(() => null);
+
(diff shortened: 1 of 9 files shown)


perminder-klair:
Reviewed and pushed a follow-up fix in `ac549d81`. The candidate preview now handles vocal status and audio-derived moods correctly, avoids a full library stats scan, and shows the true selection pool for strict shows. I also added regression tests for projection, exclusions, coverage, and the displayed counts. All CI checks are green.

PR #1444, opened 2026-08-22
fix(llm): give the script generators the track's actual feel

### LINKED ISSUE #1443: DJ link prompt asks the model to "capture its feel" but passes no tempo, energy or mood — it guesses from the title
## What happened on air

A link aired 1.4s after *Balaclava* by Arctic Monkeys started:

> "Sometimes the best way to find your footing is to **slow down and let a steady groove take over**."

*Balaclava* is 143.6 BPM — fast and frantic. The line describes the opposite of the record.

The library had it right. This is not #1417; the tempo is correct:

| field | value |
|---|---|
| `bpm` | **143.6** (correct) |
| `audio_moods` | `["workout","festival","celebratory"]` (correct) |
| `moods` | `["reflective","night"]` (wrong — text tagger) |
| `energy` | `"medium"` (understated) |

## Why

`generateLink` (`llm/internal/prompts/scripts.ts:214`) instructs:

> "Write a short DJ link to carry into the track now starting — set it up, **capture its feel**…"

and the only description of the track it passes is:

```js
if (current?.title) ctxLines.push(`Now playing: "${current.title}" by ${current.artist || 'unknown'}`);
```

`SCRIPT_CONTEXT_FIELDS` is `['date','clock','time','festival','show','listeners']` — no per-track musical fields. So the model is asked to characterise music it has been told **nothing** about beyond a title and an artist, and infers from the words. "Balaclava" reads wintry and brooding, so it wrote a wintry, brooding line. It will be right when the title happens to be evocative of the sound and wrong when it isn't.

## The part that makes this cheap to fix

`generateLink` **already has the tempo**. Three lines above the prompt:

```js
const prevAK = bpmKeyFor(previous);
const curAK  = bpmKeyFor(current);
const patterClause = (djMode && (prevAK.bpm || prevAK.key) && (curAK.bpm || curAK.key))
  ? ` You may nod to the mix if it feels natural — e.g. easing into something a touch faster or slower, or how it sits in key — but never say raw numbers.`
  : '';
```

`curAK.bpm` is fetched and then used only to decide whether a **relative** nod is permitted. Its value never reaches the model, and a relative nod carries no information about how fast this track actually is. The same function also calls `introMsFor` and `firstVocalMsFor`, so measured analysis is already in hand at that point — no new plumbing, no extra query.

## Suggested direction

Pass a short qualitative feel line for the track now starting, e.g.:

```
Now playing: "Balaclava" by Arctic Monkeys — fast, high-energy
```

Two notes on which fields to use, both learned the hard way here:

1. **Prefer `audio_moods` over `moods`.** On this track the acoustic analysis was right and the text tagger was wrong. Wiring in the obvious field would have kept the same failure.
2. **Keep it qualitative.** "fast / mid / slow" derived from BPM rather than the number itself, so it composes with the existing "never say raw numbers" rule and degrades to nothing on un-analysed tracks — the same posture `introBudgetPhrase` already takes.

`generateIntro` (`scripts.ts:75`) has a milder version of this: it passes album and year (`Now starting: "…" by … from … (…)`) but likewise no tempo, energy or mood. Whatever shape you pick probably belongs in both.

## Confidence

The code path is certain — the track facts are provably absent from the prompt.

The **rate** is not measured. `djLog` holds only the last ~50 events and `events-*.jsonl` doesn't retain spoken text, so I have one confirmed miss rather than a frequency. I'm reporting it because the mechanism guarantees the model is guessing every time, not because I can show how often the guess lands badly. If there's an easy way to log link text alongside the track it introduced, I'm happy to gather a real sample on a 29k-track library and report back.

spaceshipmike:
Closes #1443.

## Why

`generateLink` and `generateIntro` both ask the model to "capture its feel", and both hand it only a title and an artist. There is nothing in the prompt about how the track sounds, so the model infers the sound from the words in the title. That is a coin flip.

It lost on a 143 BPM Arctic Monkeys track called *Balaclava*, which the DJ introduced with:

> "Sometimes the best way to find your footing is to slow down and let a steady groove take over."

The library had it right the whole time: `audioMoods` on that track is `["workout","festival","celebratory"]`.

The same gap makes the constraint un-fixable from the operator side. House rules banning mood-talk end up arguing with a live instruction that sits closer to the task, and the model settles in the middle — vague mood-writing, which is worse than either extreme. An operator can counterweight the prompt but never overrule it.

## What

One adjective — `high-energy` or `low-key` — appended to the track line, plus a clause telling the model it is a steer rather than copy:

```
Now playing: "Balaclava" by Arctic Monkeys — high-energy
```

New pure module `llm/internal/prompts/track-feel.ts`, wired into both generators.

## Three decisions worth reviewing

**It reads `audioMoods`, never `moods`.** On the reported track the editorial field says `["reflective","night"]` — the exact wrong steer, and reading it would have preserved the bug. `library.ts` already draws this line ("so consumers can tell 'the LLM read the metadata' from 'the audio actually sounds like this'"); this is a consumer that needs the second one.

**It deliberately leaves `bpm` out.** #1417 has `beat_track` reading slow material an octave high, so a tempo band would be confidently wrong on exactly the slow tracks this is meant to protect. Concretely, on a real 29k-track library Leonard Cohen's *If It Be Your Will* — a hushed acoustic ballad — stores as **136 BPM**. A tempo band calls that fast; the audio-mood route calls it low-key. Once #1417 lands, tempo is a natural second signal.

**It classifies with the existing `HIGH_ENERGY_MOODS` / `LOW_ENERGY_MOODS` split** from `audio-calibration.ts` rather than a new vocabulary, so a renamed or deleted mood drops out of the count instead of flipping the answer. A tie returns null — two ends pulling equally is not a feel.

## The upgrade contract

No-op when a track carries no audio moods: the suffix is `''` and the prompt is byte-identical to before, matching `introBudgetPhrase`'s posture and the "absent or malformed coerces to pre-existing behaviour" rule in `CLAUDE.md`. The `FEEL_CLAUSE` is likewise only added when a feel was actually resolved.

## Verification

Measured against a real 29,240-track library (catalogue soul, reggae, funk, punk):

| | tracks | share |
|---|---|---|
| high-energy | 14,696 | 50.7% |
| low-key | 8,565 | 29.6% |
| tie → no feel | 5,707 | 19.7% |
| no audio data → no-op | 7 | 0.0% |

So ~80% of the library gets a usable steer and the rest degrades to silence rather than a guess. Spot checks:

```
Balaclava — Arctic Monkeys           high-energy   (moods said ["reflective","night"])
Move On Up — Curtis Mayfield         high-energy
Zombie — Fela Kuti                   high-energy
If It Be Your Will — Leonard Cohen   low-key       (stored bpm 136 — see #1417)
```

- New `scripts/track-feel.test.ts` — 8 cases, all passing: the reported track, the moods-vs-audioMoods disagreement, mixed and tied moods, out-of-vocabulary labels, case-insensitivity, and every malformed shape returning `''`.
- `npm run typecheck` clean. `npm run lint` 0 errors (the `any` in the resolver matches `bpmKeyFor` and the surrounding prompt modules).
- Full suite: this branch showed 6 failures, clean `develop` showed 23, both under Node 22 and both exiting 0 — `voice-air-events.test.ts` and friends appear flaky under the parallel runner. That suite passes alone on both. Nothing here touches it.

## Not in this PR

The clock half of #1443 — `buildContextLines` injecting `Local time:` into the same prompt, which is the same defect in a second place (the model is handed the clock, then told not to say it). Kept separate to stay focused; happy to open it once you've got a view on the shape.

### DIFF
--- a/controller/scripts/track-feel.test.ts
+++ b/controller/scripts/track-feel.test.ts
@@ -1,0 +1,85 @@
+// Unit tests for the track feel resolver (llm/internal/prompts/track-feel.ts)
+// — the one adjective the script generators get about how a track actually
+// sounds (issue #1443).
+//
+// Two families of pins: the arousal split must read the AUDIO moods (never the
+// editorial `moods`, which is what produced the reported bug), and every
+// unknown case must degrade to '' so an un-analysed track builds a prompt
+// byte-identical to the pre-change one.
+//
+// Run: `npm test -- track-feel` (tsx scripts/track-feel.test.ts).
+
+import assert from 'node:assert/strict';
+import { trackFeel, trackFeelSuffix } from '../src/llm/internal/prompts/track-feel.js';
+
+let failures = 0;
+function test(name: string, fn: () => void | Promise<void>) {
+  return Promise.resolve()
+    .then(fn)
+    .then(() => console.log(`  ✓ ${name}`))
+    .catch((err) => { failures++; console.error(`  ✗ ${name}\n      ${err?.message || err}`); });
+}
+
+await test('the reported track reads high-energy from its audio moods', () => {
+  // Arctic Monkeys — "Balaclava", 143.6 BPM. The DJ introduced this with
+  // "slow down and let a steady groove take over".
+  const balaclava = { audioMoods: ['workout', 'festival', 'celebratory'] };
+  assert.equal(trackFeel(balaclava), 'high-energy');
+  assert.equal(trackFeelSuffix(balaclava), ' — high-energy');
+});
+
+await test('the editorial moods field is ignored, even when it disagrees', () => {
+  // Same track: `moods` says ["reflective","night"] — the tagger reading the
+  // metadata. Reading that field instead would keep the bug.
+  const balaclava = {
+    moods: ['reflective', 'night'],
+    audioMoods: ['workout', 'festival', 'celebratory'],
+  };
+  assert.equal(trackFeel(balaclava), 'high-energy');
+});
+
+await test('a genuinely quiet track reads low-key', () => {
+  assert.equal(trackFeel({ audioMoods: ['calm', 'reflective', 'night'] }), 'low-key');
+  assert.equal(trackFeelSuffix({ audioMoods: ['calm', 'night'] }), ' — low-key');
+});
+
+await test('mixed moods resolve to the dominant end', () => {
+  assert.equal(trackFeel({ audioMoods: ['workout', 'driving', 'night'] }), 'high-energy');
+  assert.equal(trackFeel({ audioMoods: ['calm', 'focus', 'energetic'] }), 'low-key');
+});
+
+await test('an even split is not a feel', () => {
+  assert.equal(trackFeel({ audioMoods: ['workout', 'calm'] }), null);
+  assert.equal(trackFeelSuffix({ audioMoods: ['workout', 'calm'] }), '');
+});
+
+await test('un-analysed and malformed tracks are a no-op', () => {
+  // Every one of these must produce '' — the prompt is then byte-identical to
+  // the pre-change one, which is the upgrade contract in CLAUDE.md.
+  for (const track of [
+    {}, null, undefined,
+    { audioMoods: [] },
+    { audioMoods: null },
+    { audioMoods: 'workout' },      // not an array
+    { moods: ['reflective'] },      // editorial moods only, no audio scoring
+  ]) {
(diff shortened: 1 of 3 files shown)


perminder-klair:
Reviewed this and found no blocking issues. The `audioMoods` split and no-data fallback look right, including the ID-only library lookup used by queue projections. Focused tests, lint/typecheck, and diff checks pass; the remaining suite noise is unrelated and reproduces on `develop`. I’d merge once GitHub runs the required controller and web lint checks.

PR #1453, opened 2026-08-23
fix(analyzer): accelerate CUDA sounds-like backfills

### LINKED ISSUE #1426: Analyzer set to Cuda falling back to CPU
I have a dual GPU machine with:

- 1x RTX 3080 
- 1x RTX 3090, 
-  Driver Version: 580.159.03 
- CUDA Version: 13.0 
-  NVIDIA Container Toolkit installed with Docker properly configured

I am using the `ghcr.io/perminder-klair/subwave-analyzer-cuda` image and have run with the default settings and also tried setting `ANALYZE_DEVICE=cuda `. Both ways I can use a console in the container to run an `nvidia-smi` command and both of my cards shows up.

When I run a ` Audio fingerprint · sounds-like`  scan on my library, every CPU core maxes out and I get maybe 4% GPU usage on the 3090. Here is a little bit from the log:

```
[INFO] analyzer - [analyze] [analyze-worker] analysis device: cuda (NVIDIA GeForce RTX 3090)
[INFO] analyzer - [analyze] [analyze-worker] CLAP transformers model loaded: laion/clap-htsat-unfused
[INFO] analyzer - [analyze] [analyze-worker] idle model release armed — models unload after 300s without heavy use
INFO:     172.20.0.6:38404 - "POST /analyze HTTP/1.1" 200 OK
```

The last line repeats. 

Running `nvidia-smi` with the scan running does show `/opt/analyzer/venv/bin/python` using `992MiB` or VRAM, but the processor doesn't seem to get used at all, just the CPU cores.

What am I missing? Where should I look to see why the GPU isn't being used exclusively?

perminder-klair:
## What

- Add an embedding-only worker path for already-analysed tracks missing only their CLAP sounds-like vector.
- Batch each track start/middle/late CLAP windows into one CUDA transformers forward, while keeping CPU and ONNX sequential and falling back to sequential CUDA inference if the batch fails.
- Keep new, stale, re-analysis, vocal, and stem work on the full acoustic path, and clarify the CPU/GPU split in the operator docs.

## Why

The CUDA analyzer was selected correctly, but a sounds-like backfill still repeated CPU baseline analysis and ran each CLAP window as a separate model call. That saturated the CPU and left long idle gaps between short GPU bursts.

Closes #1426

## How tested

- `cd controller && npm run lint`
- `cd web && npm run lint`
- `cd mcp-subwave && npm run lint`
- `cd controller && npm run gen:themes` plus generated-mirror diff check
- `cd controller && npm run gen:schemas` plus generated-mirror diff check
- `cd controller && npm test -- analyze-mode`
- `cd controller && PATH=<venv-with-numpy> npm test -- analyzer-python`
- Python bytecode compilation for the worker, regression suite, and sidecar server
- Full controller suite: 728 passed; 7 existing `voice-air-events.test.ts` cases were cancelled because their pending promise outlives the unref timer. The focused file reproduces the same 8 passed / 7 cancelled result and this PR does not touch that path.

## Notes for reviewers

No NVIDIA GPU is available in this checkout. The regression suite exercises the CUDA processor/model batch boundary with a fake tensor surface, verifies one processor/model call for three windows, and verifies the sequential fallback after a simulated CUDA OOM.

### DIFF
--- a/controller/scripts/analyze-mode.test.ts
+++ b/controller/scripts/analyze-mode.test.ts
@@ -1,0 +1,13 @@
+import assert from 'node:assert/strict';
+import test from 'node:test';
+import { analysisModeForTrack } from '../src/music/analyze-capability.js';
+
+test('an already-current track missing only its sounds-like vector uses embedding-only work', () => {
+  assert.equal(analysisModeForTrack('audio-only', new Set(['stale', 'vocal', 'stems']), true), 'embedding-only');
+});
+
+test('baseline, vocal, and stem scopes retain full acoustic analysis', () => {
+  const full = new Set(['stale', 'vocal', 'stems']);
+  for (const id of full) assert.equal(analysisModeForTrack(id, full, true), 'full', id);
+  assert.equal(analysisModeForTrack('ordinary', full, false), 'full');
+});
--- a/controller/scripts/analyze_worker.py
+++ b/controller/scripts/analyze_worker.py
@@ -681,6 +681,15 @@ def estimate_sections(y, sr, librosa, chroma=None):
 # so a worker with embeddings DISABLED never needs torch/transformers/onnx and
 # the librosa-only venv keeps working.
 # ---------------------------------------------------------------------------
+def clap_batch_enabled(mode, device):
+    """Only the published CUDA transformers path benefits from window batching.
+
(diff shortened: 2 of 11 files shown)


joshp23:
Nice work, thanks!

PR #1459, opened 2026-08-24
refactor(web): extend TanStack Query across admin

### LINKED ISSUE #1368: web: extend TanStack Query beyond /admin/library (177 hand-rolled fetch sites)
## What

`@tanstack/react-query` is already a dependency, but it is mounted in exactly one place — `components/admin/AdminQueryProvider.tsx`, inside `LibraryPanel`. Every other admin panel hand-rolls server cache state.

Current count in `web/`:

- **50 components under `components/admin/` use `useEffect`**, wrapping **177 `adminFetch` call sites**
- each re-implements some subset of: loading flag, error state, `AbortController`, stale-response guard, manual refetch-after-mutation, polling interval

`lib/poll.ts`'s `pollWhileVisible` is a hand-written version of `refetchInterval` plus the default `refetchIntervalInBackground: false`.

## Clearest candidates

`hooks/useModelDiscovery.ts` (90 lines) and `hooks/useVoiceDiscovery.ts` (104 lines) are near-identical twins — debounce + `AbortController` + stale-response guard + error surface. Both collapse into a `useQuery` with a debounced key. Converting these two first proves the pattern outside the Library page without touching a panel's layout.

After that, the panels with the most fetch surface (`DashPanel`, `SettingsPanel`, `ShowsPanel`) are the ones that pay back most.

## Constraints (from `web/CLAUDE.md`)

- Hoisting the `QueryClient` from `LibraryPanel` to `AdminShell` is called out as a two-line move "when the second panel converts" — that hoist belongs in whichever PR converts the second panel.
- Client defaults stay as-is: `refetchOnWindowFocus: false`, `retry: false` (`adminFetch` turns a 401 into a sign-out, so a retry fires three more unauthenticated calls and races the token wipe), `staleTime: 30_000`.
- v5 removed `useQuery`'s `onError`. Error toasts go through `useQueryErrorToast(error, enabled)`; polls that deliberately swallow failures pass `false`. **Do not** install a global `QueryCache` `onError` — it would toast all of them.
- Never pass an options key you don't mean: an explicit `staleTime: undefined` overwrites the 30s default with `0` and every list refetches on every remount, silently.
- Key-factory discipline from `library/queries.ts` applies to any new family.

## Scope

One panel (or one hook pair) per PR, as established. This issue tracks the campaign, not a single change.

## Verification

`web/` has no test suite — evidence is `web/scripts/verify-query-cache.py` and `web/scripts/verify-library.py` against the isolated verify stack (see the `verify` skill), extended to cover whatever converts.

perminder-klair:
Closes #1368.

## What changed
- hoists the admin QueryClient to the authenticated shell and scopes a credential-keyed client to onboarding
- converts retained admin reads and polls to feature-keyed TanStack queries
- converts cache-affecting writes to mutations with targeted updates or invalidation, including shared schedule, skills, playlists, themes, and restore boundaries
- preserves one-shot commands and streams as imperative actions and adds a strict AST ownership audit
- adds cross-route, credential-boundary, polling, cache-security, dirty-form, and mutation reconciliation coverage

## Verification
- `SUBWAVE_VERIFY_ALLOW_DESTRUCTIVE=1 python3 web/scripts/verify-admin-query.py` — 60/60
- `python3 web/scripts/verify-query-cache.py` — 6/6
- `python3 web/scripts/verify-library.py` — 8/8
- `python3 web/scripts/verify-hooks.py` — 22/22
- `SUBWAVE_VERIFY_ALLOW_DESTRUCTIVE=1 python3 web/scripts/verify-forms.py` — 13/13
- `python3 web/scripts/verify-playlist-dnd.py` — all checks passed
- `python3 web/scripts/verify-schedule-booking.py` — 9/9
- `node web/scripts/audit-admin-query.mjs`
- `node --test web/scripts/audit-admin-query.test.mjs` — 9/9
- `npm --prefix web run lint` — 0 errors; 5 existing warnings
- `npm --prefix web run build` — 33/33 pages
- `git diff --check origin/develop...HEAD`

The browser matrix ran against the isolated verify stack on controller `:7791` and web `:7793` with disposable state and provider credentials neutralized.


perminder-klair:
Review findings — 8 verified correctness regressions, ranked most severe first.

**1. `web/components/admin/dash/queries.ts:27` — dash status poll now fails closed.**
`fetchDashStatus` wraps `/now-playing`, `/state`, `/session` in one `Promise.all` of `adminJson`, so any single non-2xx rejects the whole query. The old tick had no `r.ok` check and `.json().catch(() => null)` per endpoint, so one degraded endpoint (e.g. `/session` mid-roll) still refreshed the rest. Now the entire dash (now-playing, queue, booth log) freezes behind an error every 3s, and on first load renders nothing.

**2. `web/components/admin/settings/queries.ts:97` — a committed POST whose follow-up GET fails is reported as a failed save.**
`useSettingsMutation` gates success on the mandatory authoritative refetch; a transient GET failure after the POST committed makes `mutateAsync` throw. MoodsPanel/FestivalsSection then toast the contradictory `Save failed: settings were saved, but refresh failed: …`, skip `form.reset` (Save stays armed over saved data), SettingsPanel drops the `requiresRestart` receipt, and the errored `settingsKeys.detail()` entry raises a full ErrorState. ThemeSection got dedicated saved-but-refresh-failed copy; these callers did not.

**3. `web/components/admin/schedule/SchedulePanel.tsx:162` (also `WebhooksPanel.tsx:296`) — hydrate-once from stale cache + whole-document save.**
`settingsAppliedRef` applies the *first* delivered `settingsQuery.data` — a cached entry up to gcTime (~5 min) old on remount — and discards the background refetch even when the board is untouched. "Save week" then PUTs the whole week, silently reverting changes made elsewhere (second tab, MCP `subwave_schedule`) since the snapshot. WebhooksPanel baselines the same way and POSTs the entire array. Old code always fetched fresh on mount before hydrating.

**4. `web/components/admin/ShowsPanel.tsx:366` (also `SchedulePanel.tsx:456`) — one failed background refetch hides a loaded, possibly dirty panel behind ErrorState.**
`err` mirrors `settingsQuery.error` unconditionally (line 106) and the `if (err)` early-return precedes the `!data` check. With `retry: false`, a single transient blip during a background refetch replaces the hydrated roster/board — unsaved edits included — with an error screen until a manual retry succeeds. PersonasPanel in this same PR guards exactly this case; these two panels do not.

**5. `web/components/admin/DoctorPanel.tsx:183` — a stream that ends without `done` discards partial sections.**
Old `run()` returned the already-streamed sections as the report when the SSE cut mid-run. New code throws `doctor stream ended before completion` → reruns the full batch; if that also fails, `setInFlightReport(null)` wipes the sections the operator just watched stream in, leaving only an error plus the previous run's stale report. Even the single-failure case now costs a full duplicate doctor pass.

**6. `web/hooks/discovery-queries.ts:85` — ↻ refresh no longer supersedes a hung probe.**
`refresh()` calls `client.fetchQuery` on the same key with no `cancelRefetch`; query-core returns the existing in-flight retryer promise, so against a hung LLM/TTS endpoint — the exact case for clicking ↻ — the click is a silent no-op until the stuck request settles (fetch has no timeout). The old hooks bumped a monotonic reqId and always fired a superseding request.

**7. `web/components/admin/library/useTaggerControls.ts:108` — `/settings` cached under two unreconciled keys.**
The library audio toggles POST `/settings` but invalidate/patch only `libraryKeys.settings()`; the `settingsKeys.detail()` copy that SettingsPanel, MoodsPanel, ShowsPanel, SkillsPanel, SchedulePanel, and BlockRulesCard read is never touched, and no reconciler bridges the families. Flip a toggle in Library, open Settings within the 30s staleTime → pre-toggle values. BlockRulesCard on the *same* library page reads the other copy and never refetches, so the two bodies disagree on one page.

**8. `web/components/admin/DashPanel.tsx:261` — cancel-queued rollback stomps fresher poll data.**
No `cancelQueries` before the snapshot and no error-path invalidation (`onDone` maps to `onSuccess` only). On the documented 409 (track went to air), a 3s poll has usually already delivered the new now-playing; the catch's `setQueryData(dashKeys.status(), previous)` restores the seconds-old full envelope — previous track shown on-air, dispatched queue rows resurrected — until the next poll. Rollback should undo only the optimistic queue filter, or invalidate instead.

Verified and cut for severity/intent: playlist browse modal now serves a ≤30s-stale index with no refresh affordance (`PlaylistBuilderPanel.tsx:253`, old code fetched fresh on every open); cross-tab auth changes remounting the whole shell (`AdminShell.tsx:306`) is documented as intended, but the 401-rotation path silently destroys unsaved work in innocent tabs — worth a note. Minor cleanup: hand-rolled `adminJson`/`AdminResponseError` copies (TakeoverCard, DashPanel, `runDashAction`, `fetchShowCandidates`), the 4x-duplicated settings-hydration effect, and dead `debugKeys.llmCalls()`/`subsonicCalls()` exports.

Issue #1382, opened 2026-08-11
Proposal: Stamp voice events at air time, not handoff (voice.start / voice.end)


mrain1p:
Every signal the station emits about DJ speech — the djLog entry, the dj.say/dj.link webhook, and the player overlay — fires when airVoice() resolves, which is the moment the handoff file is written for Liquidsoap. The audio becomes audible a few seconds later: the ~1s file poll, the queue, the crossfade settle, the ducking ramp. So every consumer of those signals runs early against the stream — including your own player overlay, which shows the line before it's spoken.

Three fixes in ascending effort, each independently useful:

1. Add durationMs to the dj.say/dj.link payloads. You already compute it — speechDurationMs() parses the WAV header for the voice serialiser's own hold. Consumers currently have to guess duration from word counts.

2. Stamp (or emit) at audible start. Liquidsoap knows exactly when voice_queue/intro_queue starts a clip (source.on_track); a ping back to the controller at that moment — you already have file and telnet channels — lets the webhook and overlay fire when the words are actually on the stream, or carry an airedAt alongside the current timestamp.

3. Fullest: dedicated voice.start/voice.end events from the queue boundaries. Exact speech windows for every consumer — Discord pipes get honest timestamps, the overlay syncs itself, and for any apps//connections consuming it stop estimating entirely.

Disclosure. I'm working on a companion app which feature currently estimates the on air ducking based through estimation, however never acurate depending on when the on air DJ is actually slotted into speak over the station.

perminder-klair:
Shipped in #1390. Your diagnosis was right, and slightly understated: the signals fired when the handoff file was *written*, before Liquidsoap had even polled it.

All three proposals, plus the overlay you flagged:

- **`durationMs`** on every voice payload. It's the clip's own measured length, not the padded figure the voice serialiser holds its lock for.
- **`airedAt`.** `radio.liq` now writes a `voice-playing.json` marker when a voice clip starts feeding, echoing an id the controller stamps into the clip's `annotate:` URI. The booth log, session turn and `dj.say`/`dj.link` all fire at that moment instead of at handoff.
- **`voice.start` / `voice.end`**, paired by `voiceId`.

[...]

PR #1466, opened 2026-08-25
feat(beds): front-pad listener request intros (#1465)

### LINKED ISSUE #1465: Front-pad a listener request's intro with a bed instead of talking over the song
From BrentNorrisKYOP in Discord ([#suggestions](https://discord.com/channels/1514647956333002812/1541421789576560640)):

> I love the DJ doing lead ins for requests, but by the nature of it being something someone wants what if it by default pads the front to do the intro, since it is something someone really wanted to hear?

Today a request intro is pushed as `introKind: 'dj-speak'` (all three paths: `routes/request.ts` ×2 and `dj-agent.runRequest`), so it airs on `say.txt` → `voice_queue` → the heavy duck (`p=0.22`), fired from `onTrackStarted`. The words land on top of the first several seconds of the song somebody explicitly asked for.

The mechanism to fix it already exists — `queue.maybePushBed()` queues an instrumental ahead of the item and `onBedStarted()` airs the link over it, giving Song A → bed (DJ talks) → ramp → Song B. Request intros are excluded by one line in `broadcast/queue.ts`:

```
// v1 is links only. Request intros ride the HEAVY duck by design, and a bed
// under a heavy duck is inaudible — bedding them means reworking the duck
// routing, which is its own change.
if (item.introKind !== 'link') return;
```

Two things have to change beyond lifting that gate:

- `bed-policy.bedWanted()` only beds when the DJ would outlast the incoming track's intro. That's damage control for a long script and it's the wrong question here — a request's opening bars belong to whoever asked for it whether the intro runs six seconds or twenty. It needs to outrank the measured budget too, including an instrumental's `Infinity`.
- The duck routing the old comment names: a bedded clip has no song to talk over, so it should take the light duck rather than the heavy one that would push the bed down to a hiss.

The "or favorites" half of the original title has no hook — favourites are only a picker preference signal (`likes.favouritesClause`), not a distinct queue item, so a favourite-influenced pick is an ordinary `introKind: 'link'` pick that can already bed today.

perminder-klair:
Closes #1465.

A listener request's intro now gets a **bed in front of it** instead of being talked over the song's opening — Song A → bed (DJ does the shout-out) → ramp → the requested song, entering clean.

### Why it wasn't already doing that

The mechanism has existed since beds shipped: `maybePushBed()` queues an instrumental ahead of the item and `onBedStarted()` airs the link over it. Request intros were excluded by one line, on the grounds that they ride `say.txt`'s heavy duck (`p=0.22`) where a bed would be inaudible, and that fixing it meant reworking the duck routing.

That routing rework turned out to be four lines in `airIntro` and nothing in `radio.liq`.

### What changed

**`bed-policy.ts`** — `bedWanted()` gains a `BedReason` (`'link' | 'request'`). A link still beds only when the DJ would outlast the incoming track's intro; a request beds regardless of length.

It enters as a *reason* rather than as a `thresholdSec` the caller drops to zero, because a request must also outrank the **measured** budget — including an instrumental's `Infinity`, which short-circuits every other `bedWanted` path and which no threshold could beat. The zero-length guard still applies to both: there's nothing to bed under a clip that doesn't exist.

**`queue.ts`** — `maybePushBed` discriminates on **`requestedBy`, not `introKind`**. All three request paths (`routes/request.ts` ×2, `dj-agent.runRequest`) push `'dj-speak'`, and so does the studio's own bare push, which carries no script and falls out on the next line anyway.

`airIntro` picks its channel from **what the clip plays over, not its kind** — the same split the boundary-deferred ident already relies on (#1382). A bedded clip has no song to talk over: the thing under it is an instrumental put there for exactly this purpose, so it takes `intro.txt`'s light duck (`p=0.30`) rather than `say.txt`'s `p=0.22`, which would push the bed down to a hiss.

Requests never carry `linkPrev` or `linkClockAt`, so neither `shouldDropStaleLink` nor `linkClockDrifted` can drop the line after the bed is committed and leave it airing naked.

**`settings.beds.requestIntros`** — the full three-edit path (`DEFAULTS`, `load()`'s composition, `bedsPatchSchema` — beds is registry-driven, so the schema *is* the patch handler), plus a toggle in the admin Beds tab and the regenerated web mirror.

### The default is a stated behaviour change

`requestIntros` defaults to `true` **within** `beds.enabled`, which is itself `false`. A fresh station is unaffected. A station already running beds gains request-bedding at restart — that is deliberate, not a byte-identical upgrade, and the sub-switch is what makes it reversible. The cold-load test pins the absent-key case in both directions so it can't drift silently.

Worth knowing before merging: with this on, every request adds roughly `entryCross + 2.5s + intro + 2s` of instrumental in front of it. On a busy request night that's real airtime.

### Not in scope

The originating suggestion's title also said "or favorites". Favourites are only a picker preference signal (`likes.favouritesClause`), not a distinct queue item — a favourite-influenced pick is an ordinary `introKind: 'link'` pick that can already bed today. Nothing to build.

### Verification

- `controller` and `web` lint + `tsc`: clean. Schema mirror regenerated and stable.
- `bed-policy.test.ts` extended with the reason's four suppression cases (threshold, measured onset, instrumental `Infinity`, and the default-arg back-compat check); new `beds-request-intros.test.ts` is a cold-load round trip where the absent-key case *is* the upgrade case. 9/9 pass.
- Full suite: **795/803**. The 8 failures (`analyzer-python.test.ts`, 7 in `voice-air-events.test.ts`) are pre-existing — confirmed by running them against a clean detached worktree at `823c4fd6`, where they fail identically. `voice-air-events` doesn't import `queue.ts`.

Not verified on air — the bed→request seam hasn't been heard on a running station yet.


perminder-klair:
End-to-end verification complete on `e47750e8`.

Tested the complete listener-request intro path in an isolated stack using the production `radio.liq`, Liquidsoap 2.4.5, Icecast, the controller, Next.js, and a real headless browser/audio connection:

- **Request intros ON:** admin toggle persisted, the requested track was queued, the bed marker appeared, the intro aired on the `intro` channel, browser metadata updated, and MP3 playback remained connected.
- **Request intros OFF:** the setting persisted without changing `beds.enabled`, no new bed marker appeared, the intro aired on the `say` channel, browser metadata updated, and playback remained connected.

The E2E run found a settings persistence bug: `beds.requestIntros` was accepted by the schema/API but omitted by `settings.update()`. Fixed in `e47750e8` with a regression test covering a `requestIntros`-only patch and cold reload.

Verification results:

- Focused beds/settings tests: **79/79 passed**
- Controller lint: **passed**
- Web lint: **passed**
- MCP lint: **passed**
- Fresh GitHub CI on `e47750e8`: **all 3 jobs passed**
- Full controller suite: **796/804 passed**; the remaining analyzer failure and 7 voice-air-event cancellations are pre-existing baseline failures reproduced on `develop`, not introduced by this PR.

The scoped feature is fully verified end to end. Native mobile clients and the live production station were intentionally not touched.

PR #1478, opened 2026-08-25
fix(dj): station ident names the daypart, never the hour

spaceshipmike:
## What happened

At 15:49 station time the controller logged:

```
19:49:32Z [station-id] It is three in the afternoon on Space and Time with Michael.
```

It aired at 15:50. The station's timezone was correct (`America/New_York`); the model had the right hour and dropped the minutes.

## Why

`generateStationId` shows the model `Local time: 3:49 pm` and, since #864, says: *"If you nod to the clock, keep it loose — the time of day, never the exact minutes."* That bans the minutes but not the hour. gemma4:26b read "loose" as "truncate to the hour", which at :49 is the one number about to become wrong. The hourly time check does not have this problem because `spokenTimePhrase` rounds in code ("coming up on four in the afternoon") and the model must repeat it.

## Fix

- `time.ts`: new pure `spokenDaypartPhrase(hour)` → `in the morning` / `in the afternoon` / `in the evening` / `at night`, with the same thresholds `spokenHourPhrase` already uses. `spokenHourPhrase` now builds on it so the two cannot drift.
- `context.ts`: `getClockContext()` exposes it as `clock.spokenDaypart`.
- `prompts/scripts.ts`: the station-ident nudge becomes *say only "in the afternoon" — never the hour and never the minutes*, with a daypart-only fallback wording when the context carries no clock object. The `speakClockAllowed()` gate is unchanged.
- `scripts/clock-phrase.test.ts`: pins the daypart boundaries (incl. normalisation past the day edge) and asserts every non-midnight/noon `spokenHourPhrase` ends with its `spokenDaypartPhrase`.

Daypart rather than the rounded `spokenTime` phrase because an ident can air several minutes late, and "coming up on four" ages almost as badly as "three".

## Checks

- `tsx scripts/clock-phrase.test.ts` — all pins pass
- `tsc --noEmit` — clean; eslint — 0 errors on touched files
- `npm test` — same three environment failures as on `main` here (`analyzer-python` needs numpy; `aio-log-link` and `state-bootstrap` are macOS temp-path checks); everything else passes

https://claude.ai/code/session_01W8s9fGBF4ahSBKhcWKxSvC

### DIFF
--- a/controller/scripts/clock-phrase.test.ts
+++ b/controller/scripts/clock-phrase.test.ts
@@ -4,7 +4,7 @@
 // Run: `npm test -- clock-phrase` (tsx scripts/clock-phrase.test.ts).
 
 import assert from 'node:assert/strict';
-import { clockDisplay, spokenHourPhrase, spokenTimePhrase } from '../src/time.js';
+import { clockDisplay, spokenHourPhrase, spokenTimePhrase, spokenDaypartPhrase } from '../src/time.js';
 
 let failures = 0;
 function test(name: string, fn: () => void | Promise<void>) {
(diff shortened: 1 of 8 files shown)


perminder-klair:
Fixed the review finding in `e7b2f68e`. Deferred station idents now carry the daypart they were written for and are dropped if that daypart changes before the next track boundary, so an “afternoon” ident can’t air in the evening. Clock-disabled and immediate manual idents are unchanged. Added regression coverage, and all three CI checks pass.

Issue #1500, opened 2026-08-27
Unify the scheduler's talk ticks into a single arbitrated talk-slot scheduler


perminder-klair:
## Problem

`controller/src/broadcast/scheduler.ts` registers five independent node-cron timers that all produce **spoken air time** — `hourlyCheck` (`0 * * * *`), `stationId` (`15,30,45 * * * *`), `skillsTick` (`*/5`), `banterTick` (per-minute across the :20/:50 windows) and `programmeTick` (`*/5`). They compete for one scarce resource (the listener's ear), yet they coordinate only *implicitly*, by hand-partitioned minutes: :00 belongs to the hourly check, :15/:30/:45 to idents, :20/:50 to banter.

That minute-partitioning is fragile and has already produced two classes of bugs:

- **#310** — station ID and hourly check both firing at :00 stacked two voice segments back to back. Fixed by moving idents off :00 — i.e. by tightening the hand-partitioning, not by any real arbitration.
- **#1419** — banter's :20/:50 slots were evaluated at an instant, so a boundary-deferred ident airing at :19:35 (or a zero-floor segment spot) cancelled the whole hour's exchange. Fixed by turning the slot into a ten-minute **window** with a per-minute tick, a quiet-gap check and a skip/wait/fire state machine (`broadcast/banter-policy.ts` `banterTickPlan`).

Only banter got the #1419 fix. Idents and the hourly check still die silently when their minute collides with off-clock talk, and they log nothing when they stand down — the diagnosability work (logged-once stand-down lines with reasons and numbers) also only exists for banter.

## Findings

From reading `scheduler.ts`, `dj-gate.ts`, `banter-policy.ts`, `programme.ts` and `queue.getLastTalkBreakAt()`:

1. **The right architecture already exists in the codebase.** `banterTickPlan` is a pure, I/O-free planner (`Date` + slot state + one eligibility flag in → `skip | wait | fire` out), driven by a per-minute cron, pinned minute-by-minute by `scripts/banter-policy.test.ts`. Generalising it covers every talk tick; nothing new needs inventing.
2. **Two families of crons must not be merged with each other.** Talk ticks (the five above) are the target. Maintenance ticks — `refreshAutoPlaylist`, `cleanup`, `nightlyDoctor`, `overrideJanitor`, and the per-skill crons (operator-authored, station-timezone, re-registered on tz change) — have no arbitration concern and stay on node-cron as they are.
[...]

Issue #1419, opened 2026-08-16
Banter ticks can be systematically starved by boundary-deferred station IDs


MYBLtd:
## Summary

Banter can be systematically prevented from airing because the fixed banter ticks at `:20` / `:50` collide with boundary-deferred station IDs and the 5-minute standalone-talk gap.

This was observed on 1.8.0 with shows that have:

- `banter: true`
- multiple personas / guests present
- chatty/aggressive hosts
- active listeners
- otherwise working DJ segments

The show configuration itself appears correct.

## What happens

`banterTick` runs at `:20` and `:50` and requires at least 5 minutes since the last standalone talk break (`BANTER_MIN_GAP_MS`).

Station IDs in my setup are scheduled at `:15`, `:30`, and `:45`, but they are boundary-deferred until the next track transition.

This means an ID scheduled at `:15` can actually air at e.g. `:19:35`.

The following banter tick at `:20:00` then sees only 25 seconds since the last standalone talk and stands down.

The same thing can happen repeatedly around `:45 -> :50`.

Example from this morning:

```text
09:15:00  station ID cron time
09:19:35  station ID actually airs at track boundary
09:20:00  banter tick
           -> only 25s since last standalone talk
           -> banter suppressed
```

This repeated throughout the morning.

I had one banter cluster shortly after upgrading to 1.8.0, then effectively no banter despite several hours of eligible shows.

## Why this became noticeable

The 1.8.0 change that timestamps DJ speech at the actual on-air moment appears to make the interaction much more deterministic.

That behaviour itself seems correct: the actual airtime is the truthful timestamp.

However, in combination with boundary-deferred IDs, the fixed :20/:50 banter schedule now frequently lands inside BANTER_MIN_GAP_MS.

Additional cron-based programme features also make standalone talk denser, increasing the chance of this collision.

## Diagnostics

Other possible causes were checked:
- banter enabled on the active show
- guests present
- host frequency sufficiently high
- listeners present
- DJ budget not exhausted
- other programme/DJ segments airing normally

The important diagnostic problem is that the banter tick appears to stand down silently when the minimum-gap check fails.

There is therefore nothing obvious in the logs explaining why a fully eligible show produced no banter.

## Confirmation / workaround
As a test I used the existing operator route:

````
POST /api/dj/segment
{
  "type": "banter"
}
````

This immediately produced a normal native multi-persona banter exchange.

I now temporarily trigger this route externally at :10 and :40, deliberately away from the station-ID times.

[...]

PR #1501, opened 2026-08-27
refactor(controller): one arbitrated talk tick instead of four talk crons

perminder-klair:
PR 1 of 3 for #1500. Behaviour-identical by construction and by test — the arbitration this sets up is #1502.

## The problem

Four independent crons all produce spoken air time:

```
cron.schedule('0 * * * *', hourlyCheck);
cron.schedule('15,30,45 * * * *', stationId);
cron.schedule(banterCronExpression(), banterTick);
cron.schedule('*/5 * * * *', programmeTick);
```

They compete for one scarce resource — the listener's ear — and coordinate only *implicitly*, through minutes partitioned by hand across four cron strings and three comments. That partitioning has already produced two bugs, #310 (ident and hourly both at :00, two voice segments stacked back to back) and #1419 (banter's slot evaluated at an instant, so a boundary-deferred ident at :19:35 cancelled the whole hour's exchange), and only banter got a real fix.

So the minutes become a table and the crons become one `* * * * *` tick.

## What's here

**`broadcast/talk-scheduler.ts`** (new) — the slot table and a pure planner over it. `banterTickPlan`'s shape, generalised: banter is simply the row that already had a window, a quiet gap and logged-once stand-downs, so the mechanism moved out of `banter-policy.ts` and the numbers stayed.

```
 programme   opens=external     window=1m   gap=0s    air=immediate   clock=station  stride=5
 hourly      opens=:0           window=1m   gap=0s    air=immediate   clock=process  stride=1
 banter      opens=:20,:50      window=10m  gap=300s  air=immediate   clock=process  stride=1
 station-id  opens=:15,:30,:45  window=1m   gap=0s    air=next-track  clock=process  stride=1
```

Those values *are* today's firing minutes. The programme row keeps a 5-minute stride so it is sampled exactly where its `*/5` cron sampled it.

**The `hourlyCheck` cron's non-talk half is split out.** `rollSessionNow()` ran *before* that handler's gates, so it is not a talk action: `talkTick` awaits it at :00 unconditionally, ahead of the planner. Gating it behind the hourly row would stop a muted, empty or over-budget station rolling its session and planning its episode (finding 3 in the issue, and the one thing a naive merge gets wrong).

**Four fields on each row are load-bearing distinctions a merge must not unify away:** `air: 'next-track'` (the ident's boundary deferral, with its daypart stamp) against everyone else's immediate air; `clock: 'station'` for the programme beat, since station zones sit at :30/:45 offsets and a fixed process minute lands mid-show; `stride`; and `minGapMs: 0` meaning *no gap check at all*, not a check that usually passes.

## Proving it preserves behaviour

`scripts/talk-scheduler.test.ts` replays an **independently written model** of the four old cron expressions and the frequency rungs — minute by minute, for every rung, across two hours so the `quiet` every-other-station-hour check is exercised rather than assumed. The point is for the model and the table to be able to disagree.

The other 17 pin what a merge could quietly lose: disjoint fixed windows (#310), banter's window intact (#1419), per-row air mode and clock, `minGapMs: 0` as no check, and that no policy module is consulted for a row that is not due.

Full suite 900/901 — the one failure is `analyzer-python`'s `numpy is required`, pre-existing and environmental.

## What it deliberately does not do

- **Arbitrate.** No row loses a slot to another one yet. That's #1502.
- **Schedule the segment director.** It has no wall-clock slot at all and keeps its own `*/5` cron, cooldowns and frequency floor. That's PR 3.
- **Touch the manual runners.** `/dj/segment`'s `runHourlyCheck` / `runBanter` / `runStationId` / `runProgramme*` stay outside the scheduler and gate-free — an operator press always fires.
- **Touch the maintenance crons.** Auto-playlist refresh, cleanup, takeover expiry, nightly doctor and the per-skill crons have no arbitration concern.

## One deliberate concurrency change

Dispatch within a tick is now sequential where four crons raced. Only the programme row can come due alongside another (the fixed windows are disjoint), and the voice chain serialises anyway — so this costs nothing and makes the running order a property of the table instead of of cron registration order.

Refs #1500


perminder-klair:
Superseded by #1505, which carries all three stages on one branch against `develop`.

Same commits, same content — rebuilt to drop a `.tmptest/` scratch directory I accidentally committed here (test fixtures only, no real credentials; it came from running the suite with `TMPDIR` pointed inside the repo). Closing rather than force-pushing so the junk never sits on a live branch.

PR #1502, opened 2026-08-27
feat(controller): arbitrate the talk slots — windows, gaps and one talker a minute

perminder-klair:
PR 2 of 3 for #1500. **Stacked on #1501** — review that first; this diff is against it, and the base flips to `develop` once #1501 lands.

This is a stated behaviour change, not a refactor. #1501 shipped the table inert; this turns it on.

## The table

| kind | opens | window | minGap | air | priority |
|---|---|---|---|---|---|
| programme | station clock | — | 0 | immediate | 1 |
| hourly | :00 | 10 min | 3 min | immediate | 2 |
| banter | :20/:50 | 10 min | 5 min | immediate | 3 |
| station-id | :15/:30/:45 | 10 min | 3 min | next-track | 4 |

Idents and the hourly check now retry for ten minutes instead of firing once and vanishing. The short segments hold out for three clear minutes; banter keeps its five, because it is the longest break the station airs.

A real hour through the planner — the segment director lands at :14, the :15 ident is boundary-deferred and airs at :17:40:

```
:00  FIRE hourly
     :14  [segment] weather aired
:15  [station-id] stood down at :15 — last standalone talk 60s ago, minimum gap 180s (retrying until :24)
:17  FIRE station-id (held for the next track boundary)
     :17:40 [station-id] aired at a track boundary
:20  [banter] stood down at :20 — last standalone talk 140s ago, minimum gap 300s (retrying until :29)
:23  FIRE banter
:30  FIRE station-id (held for the next track boundary)
```

Under the old crons that hour produced an ident stacked on the back of the weather at :15, and no banter at all.

## Three rules hold the widened table together

**One talker per minute.** Windows overlap now — the ident's :15 window reaches into banter's :20 — so #310 can no longer be a property of four cron strings and becomes a rule in `talkTickPlan`: the highest-priority *firing* row takes the minute, the others wait inside their own windows. Priority follows a principle rather than a preference, so a future row can be placed without guessing: **a row that cannot retry outranks one that can** (the programme beat leads — `dueBeat` is sampled once, so a held beat is a lost beat, and its log line says exactly that instead of promising a retry), **then fewer remaining chances outranks more** (hourly, then banter, then idents).

**A row yields only to a row that is FIRING**, never to one that is merely open. Yielding to an open-but-blocked row would let a high-priority row sit on its whole window holding everything beneath it — #1419 again, one layer up. This is the subtlest line in the module and has its own test.

**In-flight talk counts.** `queue.pendingVoiceKind()` reports the segment already rendered and queued for the next boundary (`_pendingVoice`). That is talk which has *not aired*, so `getLastTalkBreakAt()` cannot see it — and it is how a :20 tick read a clear gap and still landed ten seconds in front of a :15 ident. This is #1419's actual root cause, and the window alone never fixed it. Rows with a gap honour it; a row with `minGapMs: 0` has opted out of quiet questions entirely. It also closes a quieter bug: the :30 ident used to fire while the :15 one was still pending, and `announceAtNextTrack` keeps one slot, so the older ident was silently replaced.

**Every hold is logged**, once per slot, with its cause and its numbers — gap, pending clip, or which row took the minute — plus a `slot :15 missed` line when a window closes unfired. Before this only banter said anything; an ident that collided with off-clock talk simply did not happen and the booth log recorded an absence.

## Two supporting changes worth a look

- **`dj-gate`'s ident rung asks which SLOT a minute belongs to**, not which minute it is. `[15,30,45].includes(m)` reads a retry at :18 as no chance at all, which would silently cancel every retry the windows exist to allow. Banter's rung already needed this at #1419; both now ask the same table.
- **`programme.suppressHourly()` is consulted unconditionally**, rather than behind `rolled.ctx &&`. Harmless while the hourly fired only at :00 right after its own roll — but the row now retries to :09, and a controller that booted at :03 (or whose `getFullContext` threw) would have talked straight over a programme intro.

## The headline property

**A quiet hour is unchanged.** With room in the hour, every row still fires on the minute its old cron did, for every frequency rung — the windows and gaps cost nothing when nothing is in the way; they only act when something is. That is the first test in the file, and the one to watch: if it drifts, the windows have started moving the schedule instead of rescuing it.

26 tests. The rest walk a disrupted hour, the reporter's own #1419 hour, both arbitration directions, the yield-only-to-a-firing-row rule, the pending-voice hold, and the per-slot rung for every rung. Full suite 906/907 — the one failure is `analyzer-python`'s `numpy is required`, pre-existing and environmental.

## Known limits, left deliberately

- **The segment director still fires from its own `*/5` cron**, so a segment and an ident can be generated in the same minute at :15 and the ident's gap reads state the segment has not written yet. That race is exactly what PR 3 closes, and it is the strongest argument for doing PR 3 rather than stopping here.
- **A deferred clip waits for a track boundary, not a clock**, so on a very long cut it can hold a whole window and cost the slots beneath it. Still the right answer — it airs the moment the track changes, and airing banter one second in front of it is the stacking the gap exists to prevent — and bounded by `PENDING_VOICE_MAX_AGE_MS`. Pinned as an explicit test so the trade is stated rather than discovered.

Refs #1500


perminder-klair:
To be clear about the state this shows: this was auto-marked *merged* because its base was the stacked branch below it, and I fast-forwarded that branch. **Nothing reached `develop`.** The work is open for review at #1505, on one branch against `develop`.

PR #1503, opened 2026-08-27
feat(controller): fold the segment director into the talk table as a fill row

perminder-klair:
PR 3 of 3 for #1500. **Stacked on #1502** (which is stacked on #1501) — the base flips down the stack as each lands.

This is the one that closes the race PR 2 left open.

## The last uncoordinated talker

The segment director was a `*/5` cron with a frequency floor of **zero** on `aggressive`, ticking on :00, :15, :20, :30, :45 and :50 — every opening minute the scheduled rows own. `banter-policy.ts` names it as one of the two things that could air off-clock and starve a slot, so half of #1419 is its doing; the window fixed the other half. After #1502 it could still fire in the same minute as an ident, reading a quiet gap the ident had not written yet.

## Fill rows vs slot rows

It joins the table as a **fill** row rather than a fifth slot. A fill row differs in exactly two ways, and both follow from having no scheduled chance to lose:

- **It stands down whenever a slot row *wants* the minute — including one merely waiting.** A filler that speaks resets the quiet gap and pushes that row's retry further out, possibly past its window. A slot row deliberately does *not* yield that widely (#1502), because a row that yields to a merely-open row can be starved by it. A fill row cannot be: another offer is one stride away.
- **It stands down silently.** A filler that did not fill is not an event, where a scheduled segment quietly not happening is #1419 itself.

## Why "wants the minute" and not "has an open window"

The issue's own wording was *"fires only when no fixed slot is open or pending"*. That was written before the windows widened, and it no longer means what it meant:

```
director ticks:                       [0, 5, 10, 15, 20, 25, 30, 35, 40, 45, 50, 55]
ticks inside ANY open window        : [0, 5,     15, 20, 25, 30, 35,     45, 50, 55]
free if deferring to open WINDOWS   : [   10,                        40            ]   ← 2/hour
free if deferring to WANTED minutes : [    5, 10,         25,     35, 40,        55]   ← 6/hour
```

With ten-minute windows the slot rows cover 50 of the 60 minutes, so the literal reading leaves the director two ticks an hour — not standing down, switched off. A row that has already fired its slot, or is ineligible this minute, produces no plan and wants nothing.

## An hour, traced through the planner

Off-clock talk at :14; the :15 ident is boundary-deferred and airs at :17:40:

```
:00  FIRE hourly
:05  FIRE segment
:10  FIRE segment
     :14  [segment] weather aired (off-clock)
:15  [station-id] stood down at :15 — last standalone talk 60s ago, minimum gap 180s (retrying until :24)
:17  FIRE station-id (held for the next track boundary)
     :17:40 [station-id] aired at a track boundary
:20  [banter] stood down at :20 — last standalone talk 140s ago, minimum gap 300s (retrying until :29)
:23  FIRE banter
:25  FIRE segment
:30  FIRE station-id (held for the next track boundary)
:35  FIRE segment
:40  FIRE segment
```

The director stands down at :15 and :20 — under its own cron it fired at :15, reset the gap, and pushed the ident to :18.

## The cost, stated as a number

In a clean chatty hour the director keeps `:05 :10 :25 :35 :40 :55` and loses the six opening minutes — twelve offers down to six. Those six are exactly the minutes it used to double-talk on, so the halving *is* the fix, and its own floor already limits how often it takes an offer (on `moderate` and below the lost ticks cost almost no airtime). Pinned as an explicit test, because it is the real price of this PR.

## Two smaller things

- The offer is resolved in **two passes** rather than one filtered pass, so on a contested minute the director's four gates are never consulted — the same gate-before-generation reasoning that keeps `speak()` out of the gating path.
- Cooldowns and the frequency floor stay in `skills/_agent.ts`. The table decides only whether the minute is offered, never how often the director should speak.

`startScheduler` is now one talk cron and four maintenance crons.

## Not included

The issue's other PR-3 bullet — *"possibly expose the slot table in settings so operators can shape their own hour clock"* — is left out deliberately. It is a new operator-facing surface with its own schema, validation, mirror and admin UI, and it wants the arbitrated table to soak first: the windows and priorities here are chosen to be *correct*, and letting an operator set them means deciding what happens when they overlap badly. Worth its own issue if wanted.

32 tests in `talk-scheduler.test.ts`. Full suite 912/913 — the one failure is `analyzer-python`'s `numpy is required`, pre-existing and environmental.

Refs #1500


perminder-klair:
To be clear about the state this shows: this was auto-marked *merged* because its base was the stacked branch below it, and I fast-forwarded that branch. **Nothing reached `develop`.** The work is open for review at #1505, on one branch against `develop`.

PR #1559, opened 2026-09-05
feat(scrobble): report plays back to Navidrome so lastPlayed drives rotation

### LINKED ISSUE #1298: Update Play Count and Date Last played in Navidrome when a title was played
## Feature Request: Send scrobble / play status updates to Navidrome

Hello,

Currently, SubWave streams music from Navidrome without sending scrobble or nowPlaying API calls back to the Navidrome backend during track playback.

Because of this, Navidrome does not update the lastPlayed timestamp or playCount metrics for tracks played via SubWave. As a result, Smart Playlists relying on lastPlayed filters (e.g., lastPlayed < 7 days ago to prevent track repetition) do not work as intended when fed into SubWave.
### Proposed Solution

    Now Playing: Send a nowPlaying notification to the Navidrome API when a new track begins playing.

    Scrobble: Send a scrobble API call when a track finishes playing (or crosses a percentage threshold like 50-80% of playback) to update the lastPlayed timestamp and playCount in Navidrome.

## Context & Impact

Without lastPlayed tracking, using pure random sorting on large smart playlists causes the Auto-DJ to frequently repeat the same subset of tracks, despite having thousands of titles available in the library.

Adding this feature would allow users to leverage Navidrome's dynamic Smart Playlists (.nsp) to strictly enforce track rotation and prevent frequent song repeats seamlessly.

Image

perminder-klair:
Closes #1298.

## The problem

SUB/WAVE streams *from* Navidrome but never tells it anything played, so `playCount` and `lastPlayed` stay frozen. That breaks the standard way of stopping the Auto-DJ from circling the same subset of a large library: a `.nsp` smart playlist with a `lastPlayed` filter ("not played in the last 7 days") has nothing to filter on, so pure random sorting over thousands of tracks still repeats.

## What this does

Adds Navidrome as a third backend alongside Last.fm and ListenBrainz in `broadcast/scrobble.ts`, driven from the same `Queue.onTrackStarted` transition — one writer, nothing new in `radio.liq`. Both halves go through the existing Subsonic client in `music/subsonic.ts`, so they inherit salt+token auth, the bounded fetch and the `/debug` call log:

| | call | effect |
| --- | --- | --- |
| incoming track starts | `scrobble?submission=false` | Navidrome shows it under Now Playing, counters untouched |
| outgoing track finishes | `scrobble?submission=true&time=<ms>` | `playCount` + `lastPlayed` move |

`time` is **milliseconds** since epoch (Subsonic's unit here, unlike Last.fm/ListenBrainz which take seconds) and names when the play *started*.

## Two deliberate differences from the other two backends

Both are documented at the decision (`broadcast/scrobble-pure.ts`) rather than at the call site, because both are easy to "unify" away later:

1. **No listener gate.** Last.fm and ListenBrainz fail CLOSED on an unknown listener count — a monitoring blip that pollutes a public profile is worse than a missed entry. Navidrome is the operator's own library and the point is rotation: a track the station aired to an empty room still has to stop the picker reaching for it an hour later. A fail-closed gate would switch the feature off exactly when the station is quietest, which is most of the day. So the Navidrome block runs *before* the `presentListeners()` early return.
2. **The song id is required.** Subsonic `scrobble` addresses a library row by id, so an untracked auto-playlist play that reached the mixer without a `subsonic_id` is skipped rather than guessed at from artist+title.

The shared eligibility rule (>30s long, and >50% played or >4 minutes; with an unknown duration only the >30s floor applies) moved into the new pure `broadcast/scrobble-pure.ts` alongside the Navidrome plan, so it stays one copy and can be driven from a test table instead of a live station.

## Setting

`scrobble.navidrome.enabled`, **default `false`** — an upgrade is byte-identical and sends nothing. Added to `schemas/settings.ts` (`settingsBoolLike`, matching every other settings flag), `settings/defaults.ts`, the `load()` normaliser and the `scrobble` branch of `update()`; `npm run gen:schemas` regenerated the web mirror. No secret sentinel: the credentials are already `config.navidrome`, so there is nothing to paste and nothing to redact.

## Admin UI

A Navidrome card in **Settings → Scrobbling** with the toggle, a note explaining why this one isn't listener-gated, and a Test button; the section header gains an `on/off` metric for it and the settings search index gains the field. `POST /scrobble/test` takes `provider: "navidrome"` and reports the real API error back.

## Failure behaviour

Fire-and-forget with `.catch()` → `console.warn`. No retry, no queue, nothing that can throw into the track-transition path. A Navidrome outage costs a play stamp and nothing else. (The one exception is the admin Test button, which awaits so the operator gets the actual error text.)

## Verification

- `controller/`: `npm run lint` clean, `npm test` — **1010 tests, 0 failures**, including the new `scripts/scrobble-navidrome.test.ts` (15 tests: eligibility table, the plan table, the setting's default + round-trip, and end-to-end wiring against a stubbed Navidrome — asserting the exact query params, that the setting-off case sends *nothing*, that zero listeners still scrobble, and that a failing Navidrome doesn't throw at the caller).
- `web/`: `npm run lint` clean.
- `docs/internals/broadcast.md` gains a bullet covering the reasoning above.


perminder-klair:
Live check on Navidrome 0.63.2: nowPlaying then submit both logged on a skip, and the outgoing track's playCount went 0 to 1 with `played` stamped at its real start time.

One thing to be aware of, not a blocker: auto-playlist tracks carry no duration, so the shared eligibility rule falls back to the 30-second floor. A track skipped after 73s of 207s was submitted. That is pre-existing behaviour shared with Last.fm and ListenBrainz, and the comment in `isEligibleScrobble` saying there is no skip endpoint is stale now that `/dj/skip` exists.

PR #1566, opened 2026-09-05
feat(tts): persona engine can inherit the station default

perminder-klair:
**Stacked on #1564** (`feat/dj-brain-setup`). This branch is based on that one, so
its four commits appear in the diff below — review only the two on top:

- `c15ccbd3` feat(tts): persona engine can inherit the station default
- `69fbc3f4` feat(admin): station-default voice option and the DJ Brain persona warning

Merge #1564 first.

## The bug

`tts.defaultEngine` was dead for any station that had personas. Every seeded
persona pinned engine `piper`, a pinned engine beats the station default, and
the persona schema refuses an empty engine — so an operator who switched the
station to the cloud voice still heard Piper, silently, with nothing in the
logs.

#1564's one-field DJ Brain setup hit this squarely: it configures `tts.cloud`
in one click, and the DJ then never used it. A customer on the £10/mo
Brain + Cloud Voice tier would have paid for a voice that could not air.

Found during the end-to-end test of #1564 against the hosted brain.

## What changed

**An explicit `inherit` engine for personas** — "use `tts.defaultEngine`".

It is deliberately **not** a member of `TTS_ENGINES`. That list is also the
vocabulary for `tts.defaultEngine` (which would then inherit from itself), for
the per-engine `gainDb`/`speed` maps (which need a real engine per key), and
for the `tts.fallback` rescue slot (where it would name the rung *below* it in
the chain). `ttsVoiceSlotSchema` / `repairTtsVoiceSlot` take an `allowInherit`
flag and only the persona call sites pass it; the mirror is regenerated with
`npm run gen:schemas`.

**Resolution is one pure module** — `controller/src/audio/persona-engine.ts` —
applied at the single seam where the slot is read (`djPersonaTts`). Everything
downstream (`requestedEngine`, `resolveEngine`, `ttsTarget`,
`personaCloudProvider`, and every per-engine branch in `speakWith`) compares
`personaTts.engine` against a concrete id, so resolving once keeps the sentinel
out of a dozen comparisons that would each have to learn it. `cloud-speech`'s
three `*ForPersona` entry points resolve too: they key off `engine === 'cloud'`,
so a raw inherit slot would have read as "pinned elsewhere" and dropped the
expression-cue hints on a station whose default *is* cloud.

**The voice rule is the half with teeth.** A voice id on an inherit slot was
chosen without knowing the engine, so it carries only to the local engines that
share the seed roster's id-space; `cloud` and `remote` take their voice from
the station block that names the provider:

| resolved engine | voice |
| --- | --- |
| piper / kokoro / chatterbox / pocket-tts | the persona's |
| cloud | the station's cloud voice + provider |
| remote | the server's own default |

That is what stops a Piper voice id (`bm_george`) reaching a cloud provider as
a voice *name*, where it is either a 400 or a silent substitution. The test
station was found in exactly that state.

**Admin.** The persona editor offers **Station default** as the first engine
card (persona-only — the shared `EngineVoiceFields` takes `allowInherit`, and
the station rescue slot does not pass it), with a note naming the engine the
station is currently on. The DJ Brain section now also saves
`tts.defaultEngine: 'cloud'`, and — because a pinned persona still beats that
default — lists any personas that would not follow, with one click to point
them at the station default.

## Compatibility

Fresh installs are **byte-identical**: `tts.defaultEngine` is still `piper`,
and a piper resolution keeps the persona voice, so the seeded roster is still
three distinct voices. Stored `settings.json` files are untouched — a pinned
engine resolves to itself, and the lenient load path repairs an unreadable
engine to `piper`, never to `inherit`, so a persona written before this existed
keeps the behaviour it had.

## Testing

- `controller/scripts/persona-engine.test.ts` — 14 new cases: inherit→piper
  keeps the voice, inherit→cloud takes the station provider/voice and drops the
  persona's, inherit→remote drops it, a pinned engine is untouched, a legacy
  persona is byte-identical, an unreadable station default falls to the piper
  floor, the fallback slot refuses `inherit`, and the lenient path never
  repairs *to* `inherit`.
- `cd controller && npm test` — **1137 passing**, 0 failing.
- `npm run lint` clean in `controller/` and `web/`; `npm run gen:schemas`
  regenerated and idempotent (CI's regenerate-and-diff passes).
- **End to end against the hosted brain**: with the three personas set to
  `inherit` and the station on the DJ Brain voice, the outbound request now
  leaves as
  `{"model":"dj-brain-voice","input":"…","voice":"alloy","response_format":"mp3","speed":0.92}`
  — **no `bm_george`**. The brain ledgers it as `kind=speech,
  model=dj-brain-voice, status=429, chars=0` (the OpenAI account has no credits,
  the known condition), the controller logs `[tts] cloud failed … — falling back
  to piper`, and the line still airs. No dead air.


perminder-klair:
Pushed five commits: review fixes, plus two bugs that turned up while testing.

Review fixes:

- `TTS_LOCAL_ENGINES` claimed chatterbox and pocket-tts share the seed roster's voice id-space. They don't. Chatterbox reads that field as a reference `.wav` filename and pocket-tts as a built-in id, so an inherit persona resolving to either was handed `bm_george` and failed every synth, quietly pinning the roster to the rescue chain. Narrowed to piper/kokoro and renamed `TTS_INHERITABLE_VOICE_ENGINES`.
- `repairTtsVoiceSlot`'s kokoro floor caught `inherit` too, so a persona stored with an empty voice was stamped `bf_isabella` on the next boot. That is exactly how community installs are created.
- `djSystem()` read the raw slot, so the chatterbox tag hint was dropped on a station whose default is chatterbox.
- The admin surfaces all read the raw engine: preview posted `inherit` and got a 400, the editor note promised a voice field that never rendered, the cloud-key warning stayed silent for inheriting personas, and the roster chip read `inherit`. They now go through the resolver, which moved into `schemas/persona.ts` so the browser and the server answer the same way.
- DJ Brain Save no longer overwrites a deliberate non-piper station engine. It says the engine was left alone and offers the switch as its own button. Its persona warning is provider-aware now, so a persona pinned to cloud/openai gets listed instead of counted as already following.

Bugs found in testing:

- `describeRouting()` read the raw slot, so `/debug` and the nightly doctor's active-routing check reported `requested: inherit`, no voice, and `fellBack: true` on a station where nothing had fallen back.
- Pressing Play sample on an engine the build lacks killed the controller outright. kokoro, chatterbox and pocket-tts never attached a spawn `error` handler, and an unhandled one takes the process down. The preview route skips the availability check on purpose, so the button that exists to report a broken engine was the one causing dead air. It returns 422 now.

Testing: ran the controller and admin UI directly against the real station config and library, and drove the feature through both the API and the browser. Saving a persona onto the station default, routing per station engine, a cold restart to exercise the load path, the preview button, the DJ Brain warnings and the one-click fix. The shared dev stack was busy throughout, so the Icecast stream is not verified and the chatterbox/pocket-tts voice drop is covered by unit tests only. 1152 tests pass, lint clean in controller and web.

The spawn-crash fix is pre-existing and arguably outside this PR's scope. It is only reachable from the persona editor because of this branch. Happy to pull it into its own PR if you would rather keep this one narrow.

PR #1578, opened 2026-09-06
feat(api): expose CLAP sounds-like search over HTTP and MCP (#1575)

### LINKED ISSUE #1575: feat(api): expose sounds-like (CLAP similarity) search via HTTP API and MCP
From Discord: https://discord.com/channels/1514647956333002812/1533731289167826964 (Mrain1, 2026-08-03). Triage parent: #1571.

The CLAP "sounds-like" similarity lookup is reachable only from the admin library panel. An operator building a call-in agent against the HTTP API wants it from outside.

**Shape**
- A station-auth API route (same gate as the other listener-facing library reads — `util/listener-auth.ts`, fails CLOSED) that wraps the existing similarity query used by the admin panel. Input: a track id (or a free-text search resolved to one), limit. Output: the same public track shape the search route returns — never widen it.
- A matching MCP tool in `mcp-subwave/` beside `subwave_search_library` (e.g. `subwave_similar_tracks`).
- Blocklist applies via the existing `hitOf()`/`isBlocked()` chokepoints; do not add a second rule filter.

**Watch out**
- When the heavy analyzer is off or the track has no embedding, return an empty list with a reason field, not an error.
- `mcp-subwave` has no tests; `npm run lint` there is `tsc` only.

perminder-klair:
Closes #1575

The CLAP "sounds like this" lookup was reachable only from the admin Library panel. An operator building a call-in agent against the HTTP API now gets the same neighbours without handing that agent the admin password.

## The route

`GET /similar-tracks?id=<trackId>|q=<terms>&limit=N` — wraps `library.tracksLikeThisAudio`, the same audio KNN the Library tab runs.

```
GET /api/similar-tracks?id=a1b2c3&limit=5
{
  "seed": { "id": "a1b2c3", "title": "Open Eye Signal", "artist": "Jon Hopkins" },
  "results": [ { "id": "d4e5f6", "title": "Cirrus", "artist": "Bonobo", … , "similarity": 0.87 } ],
  "reason": "ok",
  "message": null
}
```

Three things about its shape are deliberate.

**It fails CLOSED behind station auth.** `middleware/station-auth.ts` wraps `stationAuthDecision` for a GET, so the endpoint is open on a public station and shut on a private one — deliberately *not* `requireAdmin`, which is the whole point (a listener-side agent should not carry the admin credential). The password rides as `x-station-auth`, `Authorization: Bearer`, or the same `?auth=` token the stream mount already takes; an `Authorization: Basic` header is never mistaken for it. Only *failed* attempts are charged to the shared station-password bucket that `POST /station-auth` uses — charging correct ones to a 20-per-15-min ceiling would throttle exactly the authorised caller this route exists for, and 429-instead-of-401 on a wrong password is no new oracle.

**An empty result carries a reason, never an error.** A lean analyzer (`no-audio-index`), a library mid-analysis (`seed-not-analysed`), an unknown seed (`seed-not-found`) and a fully blocked neighbourhood (`no-neighbours`) all return zero tracks. `/library/search-sound` answers 503 for the first, which is right for an operator watching a capability flag and useless to an API consumer who cannot tell "ask again after the analysis pass" from "your track id is wrong".

**The row is a fixed public subset.** No tagger provenance (`source`, `originalYearSource`), no era-trust internals, no blocklist annotation — and the year resolves through `resolveEraYear` like every other listener-facing year (#1418), where the admin route publishes the raw year alongside the trust flags because an operator reviewing tags needs both. The test pins the exact key set, since "never widen it" is violated by an added key.

The blocklist rides the existing `rejectBlocked` chokepoint inside `tracksLikeThisAudio` — no second rule filter at a pick path.

## MCP

`subwave_similar_tracks` on both transports. The station password is a different secret from the admin one, so it forwards on its own header: `SUBWAVE_STATION_PASSWORD` over stdio, the caller's own `x-station-auth` over the HTTP mount. A 401 from a station-gated call names *that* secret rather than sending the agent after admin credentials it may already hold.

The Connect catalog and OpenAPI export gain a `station` auth class. Its security requirement is emitted as `[{}, { stationAuth: [] }]` — optional — so a generated client doesn't demand a password most stations don't have.

## Tests

`controller/scripts/similar-tracks.test.ts` (13 cases): the reason precedence, limit clamping, the exact published key set, era-year resolution, the credential carriers (including Basic and a repeated query param, both rejected), and the gate driven through the real middleware against real settings — public station open, private station closed on missing/wrong and open for each of the three carriers, `privatePlayer` off with `listenerAuth` on still closed. Plus an assertion that `requireStationAuth` is actually in the route's handler chain, because a middleware quietly dropped is a private library published.

`connect-catalog.test.ts` gains the optional-security assertion.

`npm test` in `controller/` passes 1136/1136; `npm run lint` clean in `controller/`, `web/` and `mcp-subwave/`. No `src/schemas/` change, so the mirror is untouched (verified by re-running `gen:schemas`).


perminder-klair:
Pushed four commits on top of the original.

Fixed from review, most important first:

- **The seed echo bypassed the blocklist.** Neighbours are filtered inside `tracksLikeThisAudio`, but the seed is resolved separately, and both `library.get()` and `library.filter()` are blocklist-blind. So a never-play track's id, title and artist came back in `seed`, and on a public station `q` turned that into an unauthenticated way to look one up by name. Both paths now resolve through one helper over the existing `isBlocked()` chokepoint, and a blocked seed reads `seed-not-found`.
- **Dropped `audioMoods` from the published row.** It is a `/library/browse` surface behind `requireAdmin`, and this route answers with no credential on a public station. The file's comment claimed a strict subset of the admin sound-search row and wasn't one; it now describes what actually ships.
- **The coverage sentence counts `stats().mirrorTotal`, not `total`.** `total` counts only tagged tracks while the analyzer writes CLAP vectors on its own schedule, so the numerator could exceed the denominator and read "covers 900 of 500 tracks".
- **Station-gated reads charge failures to their own rate-limit counter.** Same ceiling, separate bucket, so an integration left polling with a stale password can't spend the twenty attempts someone needs to unlock the player.

Smaller: `no-neighbours` no longer blames the blocklist alone (the archive filter and self-exclusion drop rows too), `?auth=` is documented as the fallback since query strings land in proxy access logs, `isInstrumental()` moved to `lyric-vocal.ts` (its sixth inlined copy), and `soundKnnWidth()` is shared with `/library/search-sound` rather than restated there.

On testing: I could not get a live station. The dev-stack lock was held by another branch's stack throughout, so nothing here is backed by a running controller, and neither the admin UI nor the logs were checked. What is verified is the suite, 1152 passing, including a new test that drives the real Express handler against a real `library.db`: blocked seed by id, by exact free text and as a would-be neighbour, the mirror-vs-tagged coverage numbers, and the published row's absent admin internals. `subwave_similar_tracks` had no test at all, so it gained seven cases over a stubbed fetch covering the header the station password rides on and the three 401 messages. Lint is clean in `controller`, `web` and `mcp-subwave`. No product bugs turned up in testing; the four above all came from the review.

Two things worth your call. `8b79d87e` is cleanup rather than a review finding (one `AuthPill` component and one auth-class table in place of four copies of the same ternary), so drop it if you would rather keep the diff tight. And `EndpointDoc`/`McpToolDoc` are still hand-mirrored between `connect/catalog.ts` and the web `types.ts`; that predates this branch and I left it alone.

perminder-klair:
Merged `develop` (ten commits, no conflicts, schema mirror already in sync) and ran the live pass I couldn't get to before.

Station up from this worktree against a real library: 1413 tracks, 1405 with CLAP vectors. 73 assertions, all passing.

- The endpoint itself: happy path, exact published key set, era year, similarity ordering, free-text seeds, limit clamping (0, negative, 9999, junk), and all four reasons. A real row comes back as `{"id":"d5wk…","title":"Calleso","artist":"Triplexity","year":2007,"genre":"jazz","bpm":112.3,"musicalKey":"4A","instrumental":true,"similarity":0.878}`, with none of `source`, `originalYearSource`, `eraUntrusted`, `isCompilation`, `yearUntrusted`, `loudnessLufs`, `audioMoods`, `blockedBy` or `lastfmTags`.
- The blocklist fix, driven through the real admin API: blocking a track makes it vanish from `seed` by id, keeps its title out of the `q` path, and keeps it out of another seed's neighbours. Unblocking restores it.
- The gate on a genuinely private station: closed with no credential, closed on a wrong one, open for the `x-station-auth` header, `Authorization: Bearer` and `?auth=`. `Authorization: Basic` with valid admin credentials is refused, and so is a repeated `?auth=`.
- The rate-limit split: 24 failed reads from one address reach 429 while `POST /station-auth` from that same address still answers 200, and 30 correct reads in a row are never throttled.
- MCP over the HTTP mount: 20 tools listed, `subwave_similar_tracks` present and flagged read-only, a call with a forwarded `x-station-auth` returns tracks, and without it the error names the station password rather than sending the caller after admin credentials.
- `/health`, `/now-playing`, `/state`, `/schedule`, `/personas`, `/dj` and `/themes` all still serve correctly, since `routes/public.ts` was edited. `/now-playing` returns a fully enriched track. Zero exceptions in the controller log across the whole run, and the station was restored to public with an empty blocklist afterwards.

One thing turned up, and it is a comment rather than behaviour. `connect/openapi.ts` said "every path is emitted under the `/api` prefix"; it isn't, and shouldn't be. Paths are keyed prefix-free and `servers[0].url` carries `<origin>/api`, which is what the base-path model expects (prefixing both would resolve to `/api/api/...`). The export is correct as it stands, so `8183f8eb` fixes the comment and pins which half carries the prefix. Two failures in my first run were faults in my own harness, `PUT /settings` where the route is `POST`, and the same wrong assumption about the prefix.

Still not checked: the `station` badge on the Connect page. The page compiles and serves, and the catalog data behind it is right, but the badge sits behind the admin sign-in and I don't type passwords into forms. Worth ten seconds of your time. `8b79d87e` also remains optional scope if you would rather drop the cleanup.

Issue #1571, opened 2026-09-05
Discord triage: unfiled broadcast, audio and library suggestions (11 threads)


perminder-klair:
> Sourced from the community #suggestions forum on Discord. Full cross-reference (all 65 open threads against every GitHub issue) lives in [`docs/discord-open-suggestions.md`](../blob/develop/docs/discord-open-suggestions.md).

Discord suggestion threads about broadcast, audio and the music library that have **no GitHub issue at all**.

**Update 2026-09-06:** on a re-check against `develop`, three of the eleven were already delivered before this triage was written. They are ticked below with the PR that landed them; the replies to post are in [`docs/discord-reply-drafts.md`](../blob/develop/docs/discord-reply-drafts.md). The earlier note that *Blocking album doesn't work* was an unfiled bug is withdrawn — it was fixed in #1467 three weeks after the thread was opened.

## Already shipped — reply on Discord and close the thread

- [x] **[Blocking album doesn't work](https://discord.com/channels/1514647956333002812/1534208297459257344)** — Solo_mag, 2026-08-04
  Fixed in #1467 (2026-08-25). The library now stores album/artist ids, so a blocked album matches by id instead of the (album name, track artist) fallback that failed on compilations. Reporter needs a library sync after updating.
- [x] **[Configurable loudness normalisation (louder lift for very quiet tracks)](https://discord.com/channels/1514647956333002812/1521450603094474792)** — Tim Stans, 2026-06-30
  Shipped in #758 (2026-07-02): `loudness.maxBoostDb` (0–12 dB) and `loudness.targetLufs` are operator settings; the boost stays peak-aware.
- [x] **[Tagging optimization](https://discord.com/channels/1514647956333002812/1521908170119839844)** — EveningPill, 2026-07-01
  `embedding.knnNeighbours` has defaulted to 10 since #721, and `moodVoteThreshold` / `confidenceThreshold` are editable. Reply pointing at the knobs; revisit the defaults if the reporter has better values.

## Candidates for the next round (suggested order)

- [x] **[Minimum track length filter](https://discord.com/channels/1514647956333002812/1532717495541174342)** — Solo_mag, 2026-07-31
[...]

PR #1580, opened 2026-09-06
feat(library): consolidate the scene vocabulary after a tagging pass

### LINKED ISSUE #1577: feat(library): consolidate scene vocabulary after a tagging pass
From Discord: https://discord.com/channels/1514647956333002812/1523778045998268537 (Rones, 2026-07-06). Triage parent: #1571.

After a full library tag, the `scene` values are noisy — near-duplicates and one-off spellings that each occur a handful of times. There is no way to see the vocabulary or merge values.

**Shape**
- An admin view (library tagging panel) listing distinct scene values with counts, sortable.
- Merge / rename: pick source value(s) and a target, rewrite in place across the library DB in one transaction, log the change in the booth log.
- Model it on whatever the mood vocabulary tooling already does (`music/tag-library.ts`, `LibraryTaggingPanel.tsx`).

**Watch out**
- Do not touch the tagger prompt contract; this is post-hoc curation, so `TAGGER_CONTRACT_VERSION` stays put.
- Rewrites must go through the library DB module's own write path, not raw SQL from a route.

perminder-klair:
Closes #1577. From Discord (Rones, 2026-07-06); triage parent #1571.

After a full library tag the genre tags — **scenes**, as the Observatory calls them — carry near-duplicates and one-off spellings that each hold a handful of tracks. There was no way to see the vocabulary, let alone merge it.

## What lands

**Admin — Library → Tagging → Scene vocabulary.** A collapsible section beside the mood/energy coverage it comes out of: every distinct value with its track count, sortable by count or A–Z, filterable, one-offs (≤3 tracks) flagged. Tick the spellings that mean the same thing, pick or type the survivor, merge behind a confirmation. Fetched on expand rather than polled, like the analysis-failures list. The folds in force are listed below it, each removable.

**A merge is two halves, and both are required.**

- **The rewrite** — `db.mergeScenes` (`library-db/scenes.ts`), one transaction over just the rows carrying a retired value. Row-at-a-time in JS so the dedupe and the order-preserving rewrite state the same rules `songGenres` applies at ingest, once, rather than twice in two languages. The scalar `genre` column is GENERATED, so it follows; a rewritten row that already holds a text vector is marked `text_vector_dirty`, because the genre line is part of the embed text.
- **The durable rule** — `state/scene-aliases.json` (`music/scene-vocab.ts`), applied by `subsonic.songGenres`, the single ingest normaliser. **Without this half the feature is a no-op within one cycle**: the mirror is rebuilt from Navidrome on every walk and `upsertTrackMeta` writes the walked genres over the stored ones, so a merge with no memory is undone by the next tag pass — which is the pass that produced the noise. This is the one thing in here the issue did not ask for, and it is why the rest is worth having.

`library.consolidateScenes` pairs the two so a route cannot do one and forget the other, and records the rule **first** — a rolled-back rewrite still leaves the fold in place, where the reverse order can leave rewritten rows with no rule.

The alias map is kept **flat**: a target that is itself retired resolves through, and existing rules pointing at a source are repointed. So a lookup is one hop, and a cycle cannot be stored. The fold is case- and whitespace-insensitive but **never punctuation-insensitive** — "Hip Hop" and "Hip-Hop" are different labels, which one survives is the operator's call, and folding them in the listing would hide the thing they came to fix. Sources match the exact stored value that was ticked. Loading is lazy and synchronous because `songGenres` runs in two processes (controller + tagger child), and a boot hook one of them forgets is a silently inert feature.

Forgetting a rule stops future folds and leaves the already-rewritten rows alone. There is nothing to restore them to — several spellings became one, and which row had which is exactly what the merge discarded. The UI says so on the button.

## Routes

| | |
| --- | --- |
| `GET /library/scenes` | distinct values + counts, plus the rules in force |
| `POST /library/scenes/merge` | `{ from: string[], to }` — rewrite + record, logged in the booth log |
| `DELETE /library/scenes/aliases/:from` | stop applying one rule |

## Watch-outs from the issue

- `TAGGER_CONTRACT_VERSION` is **untouched** — this is post-hoc curation and changes nothing the tagger prompt asks for.
- The rewrite goes through the library-db module's own write path, in one transaction. No raw SQL in a route.

## Verification

- `controller/scripts/scene-vocab.test.ts` — 16 tests against a real SQLite library: the fold's boundaries, the flatness of the rule set under repeated merges, the listing, the rewrite (dedupe within a row, tag order, the generated `genre` column, dirty-vector marking, untouched rows), and the walk round trip that proves a merge survives a rescan. Real DB deliberately: the generated column, the dedupe and the dirty marking would all pass a pure test while broken.
- `npm test` in `controller/` — 1139 pass.
- `npm run lint` clean in `controller/` and `web/` (0 errors), `npm run gen:schemas` re-run, and `node scripts/audit-admin-query.mjs` + `node --test scripts/audit-admin-query.test.mjs` pass for the new admin reads.


perminder-klair:
Pushed three commits from the review.

The blocker was that reversing a merge did nothing. `planAliases` resolved the target through the alias map before filtering self-merges, so undoing an earlier merge sent the target back onto a source of the same call: every source dropped out, no rule was recorded, no row was rewritten, and the route still answered 200 naming the spelling you were replacing. The resolve-through is now skipped when it lands on a value the merge is retiring.

Two more surfaced while fixing it. A case-only merge (`rock` ticked onto `Rock`) counted as a self-merge in both halves, so no rule was recorded and the rewrite lasted only until the next walk put the file's tag back. And the alias cap sliced the head of an oldest-first list, so at the limit it dropped the rule just recorded while that merge's row rewrite still committed.

Smaller things: `dedupeScenes` now states the alias-then-dedupe rule once for both ingest and the rewrite (`applyAliases` previously had no caller outside the test); the merge route separates "nothing to rewrite, rule recorded" from "nothing to do at all"; and the web panel invalidates `['library','rows']`, since a merge rewrites the generated `genre` scalar and Browse/Tracks kept showing retired spellings until remount.

On testing, the dev stack was busy on another branch, so I ran the controller directly against a copy of the 1,413-track library and drove the real routes: 59 checks over the listing, merges, the generated column, dirty vectors, validation, the booth log and the delete route, plus values containing slashes, ampersands and accents. The walk round trip was checked from a second process reading the alias file, which is the tagger-child case. Full suite is 1145 green.

That run turned up one real bug, in a test rather than the code. The identity-rule reload test built its map with `aliasMapOf`, which never touches `coerce`, where the fix lives, so it passed with the bug reinstated. It now goes through the file and fails if the old guard returns.

Three things for you:

- The case-only fix changes behaviour the original test asserted (a value merged onto itself records no rule). Say the word and I'll split it into its own commit.
- The admin UI hasn't been clicked through, only the API. The verbatim source filter is the bit worth checking by hand, since it's what lets you tick `rock` and `Rock` together.
- A merge can still retire a genre a show or blocklist rule names. `normGenre` strips punctuation, so case and punctuation merges stay matched; only a semantic rename can orphan a filter. Probably its own issue.

PR #1581, opened 2026-09-06
feat(broadcast): configurable show handover timing and ordering (#1576)

### LINKED ISSUE #1576: feat(broadcast): configurable show handover timing and ordering
From Discord: https://discord.com/channels/1514647956333002812/1544364265060704298 (Marcel, 2026-09-01). Triage parent: #1571.

The outgoing host's sign-off is cron-fired at :55 of the show's final hour (`broadcast/programme.ts` `outroTick`) and the incoming host's intro follows with no offset, so listeners get two voices back to back and the outgoing host's last five minutes are dead weight.

**Shape**
- A configurable handover offset (minutes before the boundary the sign-off airs; default 5 so an upgrade is byte-identical).
- An ordering rule: sign-off → one closing track → incoming intro at the boundary, rather than intro immediately after outro.
- The outro row lives in the talk table (`broadcast/talk-scheduler.ts`) — adjust its window there, do **not** add a second cron. One talker per minute is a planner rule.
- The hard-roll handoff rules in the root `CLAUDE.md` (prompt memory, sign-off filtered out of the new session) must keep holding.

**Coordinate first**
Jaz666 said in #general (2026-09-03) they are splitting handover-timing work into PRs. Check for an open PR or branch before starting so this does not duplicate their work.

perminder-klair:
Closes #1576.

A show changeover is two spoken moments — the outgoing host's **sign-off** (the programme outro beat) and the incoming host's **first words** (the persona mic-pass, or the standalone programme intro when the persona did not change). Both were placed by accident rather than by design: the outro at a hardcoded station-minute `:55`, the greeting at whatever pick cycle came next. On a station with short tracks the two landed seconds apart — two voices back to back, with five minutes in which the outgoing host had already said goodbye and had nothing left to do.

## The dial: `handover.offsetMinutes`

How many station-clock minutes before a show boundary the sign-off airs. Default **5**, which puts the outro window exactly where the hardcoded `:55` put it, so an upgrade is byte-identical. Bounds `[5, 20]`, in steps of 5.

The step is not decoration. The outro is a **station**-clock window that the talk table's programme row samples **once** from a fixed **process** minute, so a window narrower than that stride — or opening off a multiple of it — is one the row never lands inside, and nothing logs because nothing was due: the show simply stops signing off. So the bound's step and the row's `stride` are now the **same constant** (`HANDOVER_OFFSET_STEP_MINUTES`, declared in `schemas/settings.ts` and imported at both ends) rather than two literals that can drift. The upper bound is the feature beat: at 20 the outro opens at `:40`, one minute clear of `:35–:39`.

Enforced at the save path, repaired at load (`normalizeHandoverOffsetMinutes` — same rule, not a copy) and re-normalised on read, since a station-profile switch and a backup restore both reach `get()` without passing `load()`.

No second cron: the outro's window moved inside `programme-pure.beatWindow`, driven by the existing programme row in `talk-scheduler.ts`.

## The rule: one closing track

Whatever the offset, one closing track plays between the sign-off and the incoming host. This is deliberately **not** configurable — the only alternative on offer is the arrangement the issue was raised about.

The wait is counted, never timed, and it needs **both** counters:

| | asks at | boundary count alone | opportunity count alone |
| --- | --- | --- | --- |
| eager drain | the track **start** | releases as the closing track *begins* ✗ | ✓ |
| pair-aware drain | ~120s before the on-air track ends — right after a sign-off, still the track it ducked | ✓ | releases *inside* that same track ✗ |

With both, a pair-drained station hears the incoming host over the closing track's outro into the transition (the mic-pass's existing spot) and an eagerly-drained one hears them at the boundary that ends it.

State lives on the queue: `noteHandoverSpeech()` hangs off `onSpoken` — the one post-air bookkeeping site — so the trigger is what the **listener heard** rather than what the scheduler fired (an operator pressing the outro pad counts; a beat fired into a muted station does not). A held mic-pass is left **pending**, never marked aired, and nothing new times it out: the mic-pass already expires at `HANDOFF_MAX_AGE_MS` (20 min) and a pending intro already survives its own gates.

The hard-roll handoff rules still hold — this only defers *when* the mic-pass airs; which memory each side is handed, and the sign-off being filtered out of the new session, are untouched.

## Elsewhere

- `broadcast/handover-policy.ts` is a new entry in the CLAUDE.md policy-module list; the decision is reached from two call sites (the queue's boundary path and `programme.maybeRunIntro`) and would have drifted inlined at both.
- Admin control sits on the TTS section's *Station voice* card, beside **Talk placement** — the same "when does the DJ talk" question one step out. Four fixed steps, so it can never carry a free-text error.
- `handover` is registered in the `/settings` patch registry from the start rather than converted into it (like `ducking` before it), so it has `fieldErrors` on day one.
- Surfaced on `/debug` beside `voice` / `clock` / `talkAir`, so "the new DJ hasn't said hello yet" reads as waiting rather than missing.

## Testing

`controller/scripts/handover-timing.test.ts` (13 tests) pins the byte-identical default, the coercion of an absent or hand-edited value, the save path refusing an unsamplable offset, **every permitted offset against every real IANA zone offset** (exactly one sample per outro window, never zero), both drain-mode orderings including a repeated ask inside the sign-off's own track, and that the trigger is the aired kind. `programme.test.ts`'s window cases were extended to the moved window.

`npm run lint` clean in `controller/` and `web/`; `npm test` 1136/1136 in `controller/`; `npm run gen:schemas` re-run.


perminder-klair:
Pushed three commits on top of the original.

The main one is a correctness fix. `closingTrackHolds()` banked a declined opportunity on every call, and one caller isn't a handover moment: `rollSessionNow` reaches `maybeRunIntro` from the wall-clock :00 tick. At the default offset the sign-off airs at :55 over a track still playing at :00, so that ask spent the one required opportunity inside the track the sign-off ducked, and the next boundary then released the incoming host as the closing track began. That is the eager-drain case the second counter exists to prevent, on the default settings. Asking is now pure and only a drain/boundary cycle records a decline, with the question moved below the voice and budget gates so a cycle that could never speak doesn't spend the spacer.

Two smaller bugs. `_handover` was never cleared, so a sign-off nobody answered deferred an unrelated later mic-pass by a cycle; it now ages out on the second changeover, counted in changeovers rather than on a clock. And the takeover-expiry roll aired the mic-pass with no ordering check, while the operator-driven takeover start and cancel keep their exemption. Also dropped the duplicated stride default in `beatWindow`, pinned the `max: 20` bound against the feature window with a test, added the live wait to `/debug`, and normalised `handover` on the way out of `GET /settings` so a restored off-step value can't leave the admin control with no matching option.

The shared dev stack was tied up, so I ran a controller against a copy of the state dir with real Navidrome and real LLM calls and drove the feature over HTTP. Verified: bounds and step rejections with `fieldErrors`, cold-load repair of a hand-edited `7`, a real sign-off arming the wait, the wall-clock roll holding without banking, release exactly one closing track later, and a stale stamp surviving the first changeover and dropping on the second. Track boundaries came from writing `now-playing.json`, the same file Liquidsoap writes. `npm test` is 1145/1145 and lint is clean in `controller` and `web`.

Audio isn't covered. With no Icecast or Liquidsoap the audible gap between the two voices is unverified, and I didn't click through the admin control. The last commit is a test-only guard that reading the wait on `/debug` never spends it, since the row is polled every couple of seconds; drop it if you want the branch minimal. The PR description still says the offset is re-normalised on read and that `/debug` distinguishes waiting from missing, both of which only became true with these commits.

Issue #447, opened 2026-06-19
Limit the length of track that can be played


R0ckGam3r:
I have a couple of album mixes (but a single track) which are about an hour long. The DJ keeps selecting one of these at least once a day, which is rather annoying, so it would be nice if we had the option to select the max length of track that can be played so I could limit to say 10 mins.

rones:
Additional proposal: I also have some "ready to go" DJ sets in my library which sometimes have a length of more than two hours. It would be nice to not have them played during autonomous blocks, but have the possibility to create a multi-hour show which allows them to be played as long as they fit in the rest of the show time.

perminder-klair:
Added an option for this in #601. There's a max track length setting in admin Settings now (in minutes, 0 means no limit), so the DJ stops auto-picking anything longer than that. Individual shows can override it too, and a show set to 0 will play any length, so you can keep a slot for the longer DJ sets. Requests still play whatever you ask for, whatever the length.

@rones the part about fitting a long set into the time left in a show is a bigger change, so I've left that out of this one for now.

PR #1582, opened 2026-09-06
feat(picker): minimum track length filter, per show and station-wide (#1573)

### LINKED ISSUE #1573: feat(picker): minimum track length filter (per show, beside maxTrackSeconds)
From Discord: https://discord.com/channels/1514647956333002812/1532717495541174342 (Solo_mag, 2026-07-31). Triage parent: #1571.

Only the maximum track length is settable today (#447, `maxTrackSeconds` per show + Danger Zone cut-on-air). Operators with libraries full of 40-second interludes, skits and intros want a **floor** so those never get picked.

**Shape**
- Per-show `minTrackSeconds` beside `maxTrackSeconds` in the show schema (`controller/src/schemas/` + `npm run gen:schemas` mirror), plus a global default under picker settings.
- `music/playlist-gen.ts` already accepts a `minTrackSeconds` knob (line ~56/328) that nothing feeds — wire it from the show.
- The picker's SQL candidate filter and the JS filter must agree (same rule as the era filter — see `docs/internals/music.md`).
- Show editor field in the web admin.

**Watch out**
- `settings.minTrackSeconds()` already exists and means the crossfade-derived floor (`settings/store.ts:141`, exposed on `/settings.values.minTrackSeconds`). Either give the new key a distinct name or make that floor the new key's lower bound. Do not overload the existing name.
- Absent setting must coerce to today's behaviour (no floor) so an upgrade is byte-identical.
- Listener requests are exempt (a request is an explicit action).

perminder-klair:
Closes #1573

Only the *maximum* track length was settable (#447). Operators whose libraries are full of 40-second skits, interludes and album intros had no way to keep them off air. This adds the floor.

## The shape

- **`settings.picker.minTrackLengthSeconds`** — station-wide, seconds, **default 0 = off** so an upgrade picks byte-identically.
- **Per-show `minTrackLengthSeconds`**, beside `maxTrackSeconds` in the show schema. `null` = inherit the station default, `0` = an explicit opt-**out** of it, `>0` = this show's own floor.
- Fields in the **show editor** and on the **LLM provider** settings card (beside the album cooldown, which is the other `picker` key).

## Why it is not just "another maxTrackSeconds"

The cap is an on-air `liq_cue_out` cut, which is why `filterPickerCandidates` says in as many words that track length is *not* a selection criterion — an over-long track stays in the pool and crossfades out at the limit. A short track has no equivalent: it cannot be stretched, so the only place to act is the point of choice.

- **`music/track-floor.ts`** owns the policy (a module, not a branch — the same decision is reached from four places) and takes the same `{ starve }` fork `show-filter.applyStrictLocks` does:
  - **hard** in the agent's discovery tools (`picker/scope.ts` `collect`, before the playlist lock so a pinned playlist's own interlude drops too),
  - **never-starve** in the pool picker's final selection and in the auto.m3u coast — those two *are* the dead-air scope the hard one falls back into, and "the station must keep making sound" outranks the floor.
- **`settings.effectiveMinTrackSec`** is the one resolver (twin of `effectiveMaxTrackSec`), so both pick paths, the coast and the show-editor diagnostic cannot disagree.
- **Unknown length passes.** A `0`/negative/non-finite duration reads as unmeasured, never as short — otherwise a 60s floor becomes "play only what we happen to have walked".
- Length is read from **both** field names (`duration` on a Subsonic child, `durationSec` on a library row).
- **Listener requests are exempt** — the request scope simply never sets the field.
- `autoPlaylistShowKey` carries the field, since editing it on the live show changes which tracks the fallback may contain.

## The name (per the issue's "watch out")

Both halves, as asked: a **distinct key name**, *and* the crossfade-derived floor as its **lower bound**. `settings.minTrackSeconds()` keeps its existing meaning (the crossfade floor on the cap) and is what a positive `minTrackLengthSeconds` must clear — enforced in `update()`, where the crossfade is known, on the same reasoning `maxTrackSeconds` already uses: below 2× the crossfade a track gets no solo airtime at all, so a floor under it expresses nothing the mixer does not already impose. `0` always passes that check, which is what keeps an untouched station identical.

## SQL / JS agreement

`library-db.candidateFilterTracks` gained `duration_sec`, so the show-editor candidate funnel applies the identical rule the pickers will. It lands **before** the strict split (the floor is not gated on `filtersStrict`, and nor is the cap) and the funnel names it in a warning, so a smaller "indexed" figure doesn't read as a broken library.

## Drive-by

`GET /settings` now returns `values.picker`, and `picker` is in the LLM section's `formKeys`. The admin form has read `values.picker` since #1563 but the route never sent it, so the **album cooldown** never round-tripped — it showed 0 on every load and the next save on that card wrote the operator's value away.

## Left out

The issue's note that `music/playlist-gen.ts`'s `minTrackSeconds` knob "nothing feeds" is stale: the Playlist Builder has fed it since the recipe knobs landed (`PlaylistBuilderPanel.tsx`). That engine is an operator tool with its own knobs and no show scope, so there is nothing to wire from a show there and it is untouched.

## Verification

`npm run lint` clean in `controller/` and `web/`, `npm run gen:schemas` re-run (mirror committed), `npm test` in `controller/` — **1146 pass, 0 fail**.

New/extended tests: `scripts/track-floor.test.ts` (pure policy incl. both `starve` postures and the unknown-length tolerance, resolver precedence, a **cold-load** round trip for the new settings key, the crossfade lower bound, the diagnostic funnel), plus cases in `show-schema.test.ts`, `picker-lock-forwarding.test.ts` and `auto-playlist-show.test.ts`.


perminder-klair:
Pushed three commits on top of the feature work.

**Review fixes.** The blocker: `applyTrackFloor` returned its input array on the never-starve branch, and the auto.m3u coast rebuilds its pool in place (`pool.length = 0; pool.push(...kept)`), so the rescue cleared the array it was about to spread back in and wrote an empty playlist. `applyStrictLocks` had the same shape and was already reachable, via a strict show whose only genre resolves to no library tag. Both now return a fresh array, and the four in-place replacements in `refreshAutoPlaylistInner` go through one identity-guarded helper. I also collapsed the two copies of "how long is this track": `trackLengthSeconds` takes the first *usable* value rather than the first present one, since `duration ?? durationSec` let a Subsonic `duration: 0` shadow a real `durationSec` beside it, and `recency.durationSeconds` now delegates to it. `library.indexed` and `playlist.total` in the candidate funnel are pre-floor counts again, so the show editor's "Playlist anchor: N tracks" still means the playlist. Six new tests cover the aliasing contract, the coast idiom, and the two 3600 ceilings that have to move together.

**Testing.** Ran a dev stack against the real 1413-track library, crossfade at 18s so the derived bound was 36s rather than the 30s floor. Covered settings validation, per-show precedence (inherit, opt-out and override each matched the library arithmetic exactly), the funnel, the coast and both pick paths. A/B on the blocker: with the old code the log read `Auto-playlist refreshed: 0 tracks` and auto.m3u was a bare `#EXTM3U`; with the fix, 40 tracks. At a 420s floor, where 30 of 1413 tracks qualify, three consecutive picks came back 465s, 461s and 435s. At 3600s every discovery call went empty, the pool picker took over and never-starved, and Icecast kept serving. A 200s request aired under a 600s floor.

**One thing testing turned up that is not this PR.** A 15-second track never reaches `dj_queue` at all, with the floor off as well as on, while a 31-second one plays fine. That is the 18s crossfade. It is the practical case for making `settings.minTrackSeconds()` the lower bound rather than relying on the distinct name alone, so I wrote it into `docs/internals/music.md`. Drop that commit if you would rather keep the branch to code.

**Open.** I did not drive the admin UI, so the settings-card field, the show editor field and the dirty-dot from `formKeys: ['llm','picker']` are code-reviewed only; their server halves are covered above. `buildCandidates` is up to 16 positional arguments and wants a scope object, which is a behaviour-preserving refactor and belongs in its own PR. The `values.picker` and `formKeys` change is a real fix for the album cooldown round trip but has nothing to do with #1573, so say the word and I will split it out.

Issue #1111, opened 2026-07-20
Scheduled show & Takover transition leaves stale auto.m3u, bypassing Playlist Only (Strict) until manual refresh


CreakyLid:
Active-show changes do not immediately rebuild auto.m3u, allowing Playlist Only (Strict) to use the previous show’s fallback.

I have reproduced what appears to be a stale auto-playlist issue when a scheduled show becomes active.

Setup:
- subwave v0.44.0
- Scheduled show: "FAULTLINE"
- One pinned Navidrome playlist containing 298 tracks
- Playlist Only (Strict): enabled
- Live queue empty, so playback showed “auto-playlist fallback”

The saved show configuration was correct:

playlistIds: ["q90[...]hsP"]
playlistStrict: true

The UI showed "FAULTLINE" as the active scheduled show, but the fallback continued playing completely unrelated tracks.

I compared the contents of auto.m3u against the pinned playlist while the problem was occurring:

Before manually refreshing:
Fallback tracks:  28
Inside playlist:   2
OUTSIDE playlist: 26

The most recent automatic refresh in the logs had been generated for the previous general show:

[scheduler] Auto-playlist refreshed: 28 tracks (... show=My Radio)

There was no immediate auto-playlist refresh when the schedule moved to "FAULTLINE".

I then pressed Broadcast > Auto-playlist > Refresh while "FAULTLINE" was active.

After refreshing:
Fallback tracks:  16
Inside playlist:  16
OUTSIDE playlist: 0

PASS: every fallback entry belongs to the strict playlist.

The next track played normally and was inside the pinned playlist. I skipped it, and the following track was also inside the pinned playlist.

This appears to show that Playlist Only (Strict) is correctly enforced when auto.m3u is rebuilt, but a show transition does not immediately rebuild/reload the fallback. When the live queue is empty, the new show can therefore inherit the previous show’s auto.m3u and play off-playlist music until the next periodic or manual refresh.

The above occurred during a normal scheduled-show transition. I have also observed the same likely refresh gap when starting a manual takeover as well. 
The scheduled-transition case is confirmed by the 28/2/26 before-refresh and 16/16/0 after-refresh comparison.

The auto.m3u should therefore be rebuilding/reloading whenever the resolved active show changes through any path:

- scheduled show start or end;
- takeover start;
- takeover cancellation;
- takeover expiry.

Otherwise the Strict playlist feature can fail to function until the manual refresh is completed.

perminder-klair:
[...]

Issue #930, opened 2026-07-08
Dashboard show picker to pin a selected show for x hours


mrain1p:
Feature Request: Temporary Schedule Override (Pin a Show)
Problem

Currently, if I want to temporarily listen to a different scheduled show, I have to manually edit my weekly schedule and then remember to change it back afterward.

Proposed Solution

Add a temporary schedule override that allows a user to pin a selected show for a specified duration (e.g., 1 hour, 2 hours, etc). Once the timer expires, playback would automatically return to the normal weekly schedule. Could be called Station Takeover, Pinned Show, or Schedule Override

Example Use Case

I'm heading to the gym and want to listen to my Hard Rock show for the next hour. Instead of modifying my weekly schedule, I could simply select the show, choose "Play for 1 hour", and after the hour is up, the system would automatically resume the regular schedule.

Suggested Functionality
Select any available show.
Choose a duration (e.g., 1 hr, 2 hrs, Custom).
The selected show overrides the normal schedule for that duration.
When the timer expires, playback automatically returns to the scheduled programming.
Option to cancel the override early and immediately resume the scheduled playlist.

This would make it easy to accommodate temporary listening preferences without disrupting or editing the existing weekly schedule.

perminder-klair:
Update on the pin-a-show request, built it. There's a new takeover control at the top of the admin shows page. Pick any show, give it an hour or two (anything from 15 minutes to 12 hours), and it takes over the station. When the timer runs out the weekly schedule picks back up on its own, and you can cancel early any time. The song that's playing never gets cut, the switch rides in on the next track like a proper handover. It's in review now, should land in the next release.

Issue #353, opened 2026-06-12
Clock is off


maykilzor:
Even though my coordinates are on point the clock seems to be off by 2 hours which does effect the auto radio list (ie; afternoon when it's 6 pm etc) maybe you can add a custom one?
(Sorry to bother think of me as a play tester)

perminder-klair:
@maykilzor added new option in settings for timezone. next release will have this update.

PR #1583, opened 2026-09-06
feat(broadcast): fade a long track out at a show boundary (#1574)

### LINKED ISSUE #1574: feat(broadcast): fade out a long track at a show boundary
From Discord: https://discord.com/channels/1514647956333002812/1544691963695988777 (shinjuku70, 2026-09-02). Triage parent: #1571.

Shows built on genres with 20–30 minute tracks (ambient, classical, prog) spill well past the show boundary, so the next show's intro airs mid-way through the previous show's music.

**Shape**
- At push time (`queue.drainToLiquidsoap`, the only writer of `next.txt`), if the track's remaining runtime would cross the next show change by more than a tolerance, stamp `liq_cue_out` at the boundary. The #447 cut-on-air path already stamps `liq_cue_out` (`queue.ts:~834`, `radio.liq:~451`) — reuse it rather than adding a second cue writer.
- The cut must be a **fade**, not a hard stop: `radio.liq` needs a fade-out on a cue-out'd track. Keep fade duration equal to the cross buffer (see `liquidsoap/CLAUDE.md`).
- Per-show switch (`fadeAtShowEnd` or similar) with a global default, absent = off so an upgrade is byte-identical.
- Route any listener-facing remaining-time through `music/silence-trim.ts`'s onset shift rather than local subtraction.

**Watch out**
- Only autonomous picks; a listener request is an explicit action and is exempt.
- The clock the drain uses is a forecast (see `broadcast/queue/pure.ts` paired guards) — the pick is made before it starts, so compute the boundary from the expected start, not now.

perminder-klair:
Closes #1574.

A schedule built on long-form material (ambient, classical, prog) picks its last track minutes before the slot ends and is still playing it deep into the next show, so the incoming presenter's opening link airs over the outgoing show's music.

### What it does

With `fadeAtShowEnd` on, the drain cues such a pick out at the show change and Liquidsoap fades it there. **Off by default at both levels** — a station-wide toggle plus a per-show tri-state where `null` means inherit — so an upgrade sounds byte-identical. The show field is tri-state rather than a plain boolean on purpose: a boolean would read every existing show as an explicit "no" the moment an operator turned the station default on.

The cut is **not a new mechanism**. It folds into the same `liq_cue_out` the #447 length cap and the silence trim already write, arbitrated "earliest wins" by `subsonic.getAnnotatedUri`, because those three all cut the same tail and only stay consistent while one call site folds them together.

### The policy (`controller/src/broadcast/show-boundary.ts`, pure)

- The boundary is computed from the pick's **expected air time**, never from `now` — the drain's clock is a forecast, and a track held in `dj_queue` would otherwise be cut short by however long it waits.
- The scan walks the **station** clock minute by minute rather than adding an hour to an instant: slots fire at the hours the operator painted them in (#353), zones sit at :30/:45, and a timed takeover's start/expiry (#930) is not hour-aligned, so it rides in as an extra candidate.
- The span measured is the **playable** one — after the cap and the trimmed tail, through `drain-policy.playableDurationSec` and `music/silence-trim.ts`, never a local subtraction — and the returned cue is an **absolute** offset, so a head-trimmed track's cut moves by exactly the skipped seconds.
- Two floors pull in opposite directions and both matter: a **60s tolerance** leaves a small overrun alone (the incoming show's first pick has its own latency), and a **90s minimum-play floor** refuses to cut a track down to a stub.
- **Listener requests are exempt**, mirroring the #447 cap. An armed cut vetoes a rendered stem blend through the existing `outCapped` flag, for the reason the cap does: the clip is mixed from a tail that will not air.

### The fade (`liquidsoap/radio.liq`)

`liq_show_fade` on the outgoing track disarms `liq_washout` and `liq_loop` — both are gestures built around the track's *own* ending, and the loop branch hands the transition `a.source` with **no fader at all** (it rides its own gain after the comb cascade). Measured on a reduced harness mirroring the arming and fade-shape selection: a loop-armed track cued out mid-record held full level right through the cross and stopped dead, the seam summing to **-33.6 dB** against a -36.1 dB track; the same cut carrying `liq_show_fade` fell back to the default `fade.out(duration=d)` and restored the equal-amplitude curve, **-39.1 dB** at the midpoint. Fade duration stays equal to the cross buffer. The incoming pick's own gestures (sweep, dissolve, chop) are deliberately untouched.

### Also

- Admin: a **Show boundaries** toggle under Broadcast, and a three-way control (Station default / Fade at the boundary / Let it run over) in the show editor, plus a list facet.
- Docs: a rule in the root `CLAUDE.md`, one in `liquidsoap/CLAUDE.md` with the measured numbers, and the full paragraph in `docs/internals/broadcast.md`.

### Verification

- `controller/scripts/show-boundary.test.ts` — 9 cases covering the station-clock scan, takeover candidates, the absolute offset with a head trim, both floors, the unknowable-input paths, the tri-state precedence, and a save/load round trip against a real settings store.
- `npm test` in `controller/`: **1132 pass, 0 fail**.
- `npm run lint` clean in `controller/` and `web/`; `npm run gen:schemas` re-run.
- `liquidsoap --check radio.liq` against the real broadcast image (2.4.5): exit 0.

### Left out

- The `auto.m3u` fallback is not boundary-cut. Its entries have no expected air time, so there is nothing to measure a boundary from; it is already rebuilt per show at every boundary (#1111).
- The tolerance and minimum-play floors are module constants, not operator dials — the switch is the operator's choice, the numbers are what make it mean "don't spill" rather than "cut on the dot".


perminder-klair:
Pushed two commits on top of the original change.

The first fixes two things from review. A bed queued ahead of a pick was invisible to the air-time forecast, because `maybePushBed` writes straight to `next.txt` and never enters `upcoming`, so the cut landed a whole link late and the track still spilled 20 to 35 seconds into the next show. Separately, the exit gestures were only stripped in the mixer, so a controller rebuilt ahead of its broadcast image could hand an armed loop to a mixer that ignores `liq_show_fade`, and that branch applies no fader at all, which is the hard stop this exists to prevent. The drain now strips them upstream too. One judgement call went the safer way: the four gestures armed off the incoming track stand down as well, since each of them reshapes the outgoing branch. A boundary cut is now a plain crossfade at both ends of the seam.

The second commit fixes a bug found while testing. `liq_show_fade` rides `item.track`, which persists in `queue.json`, but only the arming branch ever wrote it, so a crash-recovery re-drain could carry a stale flag and disarm the gestures on a seam that is not a boundary cut.

Tested on a live dev stack with the real picker, TTS and Icecast, driven through the API. With a show ending at a timed takeover's expiry, an autonomous pick tagged 293s was stamped `liq_cue_out=181.79` and played 163.5s on air before the next track began. Adding the 18s crossfade, that is a cut at about 181.5s, landing 4 seconds short of the boundary instead of spilling 111 seconds past it. Captured stream audio across that seam shows no silence and no discontinuity, a 3.4 dB dip through the crossfade. Bed accounting checked out on two real bedded links, `bedDelaySec` 11.1 and 13.98, each matching its own log line. The per-show tri-state and its validation round-trip through the API. 1139 controller tests pass, lint is clean in controller and web, and `liquidsoap --check` passes against the 2.4.5 broadcast image.

Two things remain open. With `maxTrackSeconds` set below roughly 150s, the 60s tolerance and the 90s minimum-play floor are contradictory and the cut can never arm. That is the right outcome, since the cap already stops the spill, but the feature is silently inert on those stations and the admin hint probably should say so. And two paths are covered by tests rather than live: the upstream gesture strip, because washout and loop only arm on analysed tracks and the test library has no analysis, and the listener-request exemption.

PR #1584, opened 2026-09-06
feat(tts-heavy): release idle TTS models instead of pinning ~4GB forever (#1579)

### LINKED ISSUE #1579: tts-heavy holds Chatterbox resident forever: ~4GB pinned on an idle station, and the idle pause does not reach the sidecar
Reported in #support on Discord (2026-08-08, JermaNoiD): [Chatterbox doesn't release memory when idle. Permanent 4GB in use.](https://discord.com/channels/1514647956333002812/1535605655674327050)

> I'm running Subwave using docker-compose. the tts-heavy container on cuda. I've got the "Pause when the room is empty" feature enabled. Is there any way to have chatterbox release the memory when idle?

## What happens

`tts-heavy` loads Chatterbox at startup and keeps it resident for the life of the container. On a CUDA host that is a permanent ~4GB, held whether or not the station has spoken in the last twelve hours.

## Why the existing levers don't cover it

- **`DISABLE_CHATTERBOX`** (#710) skips *loading* an engine. That is the answer for an operator who never wants Chatterbox, not for one who wants it at 8pm and not at 4am.
- **The idle pause** (`stream.idleAfterMinutes`, #1256) stands the *programme* down when nobody is listening. It has no reach into the sidecar, so the engines stay warm through the whole quiet period. This is the specific surprise in the report: the operator had the pause enabled and reasonably expected it to help.

So an operator who wants Chatterbox at all pays for it continuously, and the one setting that sounds like it should help does not.

## The ask

Unload idle TTS models in `tts-heavy` after a configurable idle period, and load them again on the next `speak()`.

## Things worth deciding before coding

- **Reload cost is the whole trade-off.** Amortising model load is why the sidecar exists at all, so an unload window set too short converts a memory problem into a latency problem on the first spoken line. The setting needs a default that is clearly on the safe side, and the reload has to be silent rather than a stall.
- **This must not become a way for the station to go quiet.** A reload that fails, hangs, or races a queued segment has to degrade the way the rest of the voice path does: fall through the rescue chain, never block the music. Music never stops for a TTS problem.
- **Per-engine, not per-sidecar.** `tts-heavy` carries Chatterbox and PocketTTS. An operator using one and not the other should not have the other's idle behaviour decided for them.
- **Prior art on the same class of problem:** #1204, where analyzer RSS grew unbounded because glibc was not returning freed buffers. Worth confirming the memory here is actually the model tensors and not an allocator artefact before building an unload path for it.

## Related

- #710 — `DISABLE_CHATTERBOX` to skip loading unused engines
- #1256 — idle pause hardening (programme only, not the sidecar)
- #1204 — analyzer RSS growth over long uptime

perminder-klair:
Closes #1579.

`tts-heavy` loaded Chatterbox at startup and kept it resident for the life of the container — on a CUDA host a permanent ~4GB, held whether or not the station had spoken in twelve hours. The programme's idle pause reaches the music chain only, so the one setting that sounds like it should help does nothing here; that's the surprise the report actually hit.

Each engine now keeps its own idle clock. After `TTS_HEAVY_IDLE_UNLOAD_S` without a render its worker is stopped and the memory goes back to the host; it loads again on the next `/speak`, or ahead of one on the new `POST /warm`, which the controller fires the moment the idle pause releases — so the reload lands while the music returns rather than on the first spoken link.

## The reclaim is a process exit

Not an in-process `del model`. #1204 settled this on the analyzer: dropping the model singletons leaves torch's imported modules — and on CUDA its context — resident until the process ends, so an in-process release hands back only part of the memory and the operator still sees gigabytes pinned. It is also nearly free here, because `run()` was already a restart supervisor: "unload" is one `terminate()` away and the worker scripts need no idle protocol of their own.

## Three things the design turns on

**A cold engine stays inside `/health`'s `engines`.** This is the one-way door and the least obvious part. `ttsHeavyClient.probeOnce` keys availability on `engines.includes(engine)`, caches it, and `tts.ts` reads that cached boolean to decide whether the engine is a usable voice slot at all. An engine dropped from the list while unloaded would never be asked to speak, and a sidecar only ever wakes an engine because someone called `/speak` on it — memory freed, voice gone for good. So `cold` and `not ready` are separate flags: a still-booting or crash-looping engine stays out exactly as before, while a cold one stays in and pays a load on its next line.

**The window sits well above the station's natural talk cadence.** The talk scheduler can offer the segment director a minute every five, so a short window would thrash a station that is merely between links rather than idle — converting a memory problem into a latency problem, which is the trade-off the issue flagged. Defaults are 1800s on cuda (VRAM contention is the urgent case, and a GPU reload is the fast one) and 3600s on cpu; both are unreachable by a station that is on air and talking. `0` restores the always-resident behaviour, and per-engine overrides mean an operator using one engine doesn't have the other's idle behaviour decided for them.

**Degrading stays immediate.** Only a cold worker — or one already loading for another caller — is waited for. A worker that is merely down fails its caller at once, which is the pre-existing behaviour and must not change: the rescue chain is what keeps the station talking, and 90s of silence before reaching it is worse than the cold start the wait exists to cover. A load that never arrives inside `TTS_HEAVY_LOAD_TIMEOUT_S` is a `503` rather than an unhandled `500`. Music never stops for any of it.

## Operator surface

```ini
# root .env — empty = 1800 on cuda / 3600 on cpu; 0 = always resident
TTS_HEAVY_IDLE_UNLOAD_S=1800
# CHATTERBOX_IDLE_UNLOAD_S=
# POCKET_TTS_IDLE_UNLOAD_S=0
# TTS_HEAVY_LOAD_TIMEOUT_S=90
```

`/health` carries the state so the reclaim is confirmable rather than inferred from `docker stats`: `cold: [...]`, `unloads: {engine: n}`, `idle_unload_s: {engine: s}`, and the existing `*_loaded` booleans now meaning residency. `pocket_voice_cloning` deliberately survives a cold stop — cloning is a property of the image, not of the process, and flickering it to unknown every idle window would make the admin warning worse about a fact that didn't change. `availableEngines()` gains `heavyCold` alongside `heavyEnabled`; nothing routes on it.

## Testing

`controller/scripts/tts_heavy_idle_test.py` — 17 dependency-free cases (fastapi/pydantic stubbed, worker subprocess faked), registered in the existing `analyzer-python.test.ts` shim. It pins the cold-engine health contract, the crash-vs-cold distinction, the in-flight and claimed-worker guards, the fail-fast path, a second caller joining an in-flight load, and the wake-racing-the-unload latch — that last one verified to fail against the pre-fix logic.

Also smoke-tested end to end against a real uvicorn server with a stand-in worker: boot resident at 165MB RSS → idle release (process gone, `cold: ["chatterbox"]`, `engines` unchanged) → `/speak` loads on demand and renders a WAV → release again → `/warm` returns in 7ms while the load runs behind it. Full `npm test` 1123/1123, controller and web lint clean, schema mirror unchanged, CLI assets regenerated.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -264,6 +264,29 @@ SITE_URL=
 # --profile tts-heavy.
 # TTS_HEAVY_ENGINES=pocket-tts     # or: chatterbox  |  chatterbox,pocket-tts
 #
+# How long an idle heavy-TTS engine stays loaded. Chatterbox is ~4GB resident
+# (weights, plus torch's CUDA context on a GPU host) and used to be held for
+# the life of the container whether or not the station had spoken all day —
+# the programme's idle pause stands the music down but has no reach into the
+# sidecar. After this many seconds without a spoken line the sidecar stops
+# that engine's worker and hands the memory back, reloading it on the next
+# line (or a little ahead of one: the controller warms the sidecar when the
+# idle pause releases, so the reload lands while the music returns rather than
+# on the first link). Empty = 1800 on cuda, 3600 on cpu; both sit far above
+# any gap a talking station produces, so this only fires when the station
+# really has gone quiet. 0 keeps the old always-resident behaviour — worth it
+# if you have RAM to spare and never want a cold reload. Per-engine overrides
+# win over the shared value. Only matters with --profile tts-heavy.
+# TTS_HEAVY_IDLE_UNLOAD_S=1800
+# CHATTERBOX_IDLE_UNLOAD_S=      # just Chatterbox (the expensive one)
+# POCKET_TTS_IDLE_UNLOAD_S=0     # e.g. keep the small, fast engine resident
+#
+# Seconds /speak waits for a cold engine to load before giving up and letting
+# the DJ fall through to its rescue voice. Raise it on a slow disk or a first
+# load that still has weights to fetch; the ceiling has to leave room inside
+# TTS_HEAVY_TIMEOUT_MS (180000) for the render that follows.
+# TTS_HEAVY_LOAD_TIMEOUT_S=90
+#
 # Own TTS server? There's nothing to set here — the *Remote* engine points the
 # DJ at any HTTP server that answers GET /health and returns rendered audio from
 # POST /speak (a GPU box, a Tailscale peer, a bridge in front of a vendor API
--- a/.env.example
+++ b/.env.example
@@ -270,13 +270,17 @@ SITE_URL=
 # the programme's idle pause stands the music down but has no reach into the
 # sidecar. After this many seconds without a spoken line the sidecar stops
 # that engine's worker and hands the memory back, reloading it on the next
-# line (or a little ahead of one: the controller warms the sidecar when the
-# idle pause releases, so the reload lands while the music returns rather than
-# on the first link). Empty = 1800 on cuda, 3600 on cpu; both sit far above
-# any gap a talking station produces, so this only fires when the station
-# really has gone quiet. 0 keeps the old always-resident behaviour — worth it
-# if you have RAM to spare and never want a cold reload. Per-engine overrides
-# win over the shared value. Only matters with --profile tts-heavy.
+# line — or ahead of one, since the controller warms the sidecar both when the
+# programme's idle pause releases and when the DJ decides to talk. The first
+# of those hides the reload completely but needs the idle pause switched ON
+# (Settings -> Stream, off by default); with it off you get only the second,
+# which overlaps the load with writing the script, so part of a cold reload is
+# still audible as a longer gap before the first line. Empty = 1800 on cuda,
+# 3600 on cpu; both sit far above any gap a talking station produces, so this
+# only fires when the station really has gone quiet. 0 keeps the old
+# always-resident behaviour — worth it if you have RAM to spare and never want
+# a cold reload. Per-engine overrides win over the shared value. Only matters
+# with --profile tts-heavy.
 # TTS_HEAVY_IDLE_UNLOAD_S=1800
 # CHATTERBOX_IDLE_UNLOAD_S=      # just Chatterbox (the expensive one)
 # POCKET_TTS_IDLE_UNLOAD_S=0     # e.g. keep the small, fast engine resident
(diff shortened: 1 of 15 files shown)


perminder-klair:
Pushed three commits for the review.

Both blockers were on the reload side. `idle_for()` read the pre-unload `last_spoke`, so a `/warm` reload was released again on the next 30s tick and `/warm` bought a full model load for nothing. It also had no trigger on a stock station: its only caller was the idle pause releasing, and `stream.idleWhenEmpty` defaults off while the unload is armed by default. It now fires when a talk slot fires too, where the load overlaps script generation. Also fixed: a cold engine whose load was abandoned held its caller for the full 90s ceiling, and `TTS_HEAVY_LOAD_TIMEOUT_S` was documented but forwarded by no compose file and parsed with a bare `float()`.

Testing found two more. A worker mid-reload fell out of `/health`'s `engines`, so the controller stopped routing to it for the 30-60s of a reload and the DJ took its rescue voice, meaning the new warm caused the fallback it was added to prevent. And an empty `TTS_HEAVY_LOAD_TIMEOUT_S` crashed the sidecar at import, which is what `${TTS_HEAVY_LOAD_TIMEOUT_S:-}` supplies, so forwarding the knob would have broken every stock install.

Tested on the dev stack with the real `server.py` under uvicorn and a stand-in worker in place of the torch model. Each fix also ran against the pre-fix code through the same harness: the warm now survives its next tick, an abandoned load answers in 258ms instead of 30s, and a junk ceiling boots instead of never binding the port. On the live stack the release surfaced in `/settings` as `heavyCold`, a cold `/speak` via `/dj/say` loaded the engine on demand and aired, and the hourly check logged `warming idle-unloaded engine(s): chatterbox` at the top of the hour with the load overlapping script generation. Music kept playing throughout, and `npm test` is 1128/1128.

Open: the admin badge was checked against the live `/settings` payload rather than in a browser, and real Chatterbox weights were not exercised. `CHATTERBOX_REFERENCE_WAV` has the same unforwarded-knob problem but predates this, so it is excluded from the new drift check rather than fixed. The admin badge commit (520a82e) is optional scope if you would rather keep this to the sidecar.

Issue #1570, opened 2026-09-05
Discord triage: unfiled admin UI and operations suggestions (10 threads)


perminder-klair:
> Sourced from the community #suggestions forum on Discord. Full cross-reference (all 65 open threads against every GitHub issue) lives in [`docs/discord-open-suggestions.md`](../blob/develop/docs/discord-open-suggestions.md).

Discord suggestion threads about the admin UI and day-to-day operation that have **no GitHub issue at all**. Several are quality-of-life asks from operators running large rosters. The last three are grab-bag threads that need reading and splitting before anything can be filed.

- [x] **[Shows and personas sorting](https://discord.com/channels/1514647956333002812/1535671834522030080)** — shinjuku70, 2026-08-08
  This project is so awesome that I now have 11 DJs and 44 distinct shows And of course I’m continuously tweaking them… so it would be helpful to be…
  Nothing on GitHub for sorting/filtering large show and persona rosters.
- [ ] **[Unique Listener Profiles](https://discord.com/channels/1514647956333002812/1535427441055502399)** — Gurthyy, 2026-08-07
  This would be a long term thinga nd might not even be something anyone is interested in - but I think it'd be pretty neat if as the next…
  No issue. Listener identity beyond the beacon has never been scoped.
- [x] **[Mood->Speech->Corrections added Feature](https://discord.com/channels/1514647956333002812/1535374779459313664)** — Drachenhort, 2026-08-07
  Wouldnt it be great to be able to test the correctional Speech Output without having to wait for the Persona to speak. I think that would be nice…
  Shipped after this thread was filed: `ffd5d945` "Test corrections button in Moods → Speech tab" (#1350, 2026-08-12). `MoodsPanel.tsx` renders the preview beside the corrections list and `POST /settings/tts` accepts an **unsaved** `corrections` override, so a rule can be heard before it is saved.
- [x] **[Be able to assign skill durring its creation](https://discord.com/channels/1514647956333002812/1530780746162372768)** — Mrain1, 2026-07-26
  When you create a new skill you are not able to assign it to dj's from the creation screen. it could streamline the process if you are able to…
  Nothing on GitHub.
- [x] **[Settings Search](https://discord.com/channels/1514647956333002812/1529175110735495248)** — Gurthyy, 2026-07-21
  We have a decent amount of settings across different pages and sometimes I can't remember where they actually live (ADHD is a crazy thing) - was…
  Nothing on GitHub.
[...]

PR #1585, opened 2026-09-06
feat(backup): scheduled, rotating backups (#1570)

perminder-klair:
Scheduled, rotating backups. Closes the "Backup DB/Settings on Schedule" item in #1570 — the admin UI could take a backup by hand, but there was no way to say "daily, keep the last N".

## Shape

- `backup/zip.ts` — the zip builder, extracted verbatim from `GET /backup/export`. The export route is now four lines and byte-identical: the manifest's `contents` is still computed before `manifest.json` is added, the tmp dir still dies in a `finally`.
- `backup/pure.ts` — name grammar, `backupDue`, `lastScheduledBackupAt`, `backupsToPrune`. No clock, no fs.
- `backup/scheduled.ts` — the run. Returns `{written, pruned, errors}`; the scheduler logs. Atomic write via `writeFileAtomic`, and the `.tmp` name isn't a `.zip`, so a half-written file can't be listed or restored.
- `settings.backups = {cadence, keep}`, default `{off, 7}` — an operator who upgrades and changes nothing gets byte-identical behaviour.

## Three decisions worth review

**Scheduled backups get their own name.** `GET /backup/restorable` accepts any top-level `*.zip`, so no new restore path was needed — but a `subwave-backup-*` scheduled name would be indistinguishable from an operator's hand-copied restore point. The writer uses `subwave-auto-backup-YYYY-MM-DD-HHMMSS.zip` against an anchored regex, and `backupsToPrune` filters through it before counting anything. Retention can only ever delete a file it could have written itself.

**Hourly tick, not a nightly cron.** A daily cron only fires on a station that happens to be up at that minute, and plenty of these run on a box that's off overnight — a backup feature that never fires on exactly the installs most likely to lose their state dir isn't one. The tick is cheap; `backupDue` does elapsed-time arithmetic against the stamps on disk.

**The newest scheduled file's name is the record of the last run** — no marker file, so a failed run leaves nothing claiming success. A stamp in the future is ignored rather than trusted, so a clock that was once wrong can't wedge backups shut for years.

`GET /backup/restorable` now returns `auto` per file from the same grammar the sweep uses, so the badge can't disagree with the prune.

## Verification

`npm run lint` clean in `controller/` and `web/`; `npm test` 1146 pass / 0 fail; mirror regenerated by `gen:schemas` and stable.

One thing checked deliberately: adm-zip's `addLocalFile` reads with `readFileSync` at add time, so cleaning the tmp dir inside the builder before the caller's `toBuffer()` is safe — that was the one way this refactor could have silently dropped the tag DB from every backup.

## Not done

No entry in `web/components/admin/settings/registry.ts` — the UI lives in the Backup panel, so "the registry" was read as the controller patch registry. Say if you want a settings-search entry pointing at the Backup panel instead.


perminder-klair:
Pushed five commits on top of the original.

Review fixes:

- A backup killed mid-write left a partial file that nothing ever cleaned up. The temp name is invisible to both `isSafeBackupName` and the retention grammar, and the run retries hourly, so a restart during a write quietly cost disk every hour. `writeFileAtomic` now removes its own temp when the write or rename fails, and every run sweeps the ones a killed run left behind, under its own anchored name so another writer's in-flight `settings.json.<hex>.tmp` is never touched.
- Added a free-space pre-flight so a backup cannot take the volume's last byte. It fails open on a filesystem `statfs` cannot measure.
- `backupsToPrune` and `normalizeBackups` disagreed about an unreadable `keep`: the prune fell to 1, the loader to 7. There is one clamp now, and it reads absent or empty as no answer rather than as `Number(null) === 0`.
- Backup panel: "Save schedule" could stay lit with no save that would clear it, because the patch dropped `keep` while the cadence was off, so the form and the stored value could never converge. The card also never re-seeded, so restoring from that same panel left it showing the old schedule, and the on/off pill read unsaved state.

Found while testing: the free-space message said "67 MB headroom" because the constant was 64 MiB while the message divides by 1e6. It is 64,000,000 now, so the constant and the message agree.

Testing: ran the dev stack and waited for the real `23 * * * *` tick rather than calling the function directly. With fixtures planted first, one tick wrote the backup, pruned to `keep`, swept a stale temp, and left both a hand-copied `subwave-backup-*.zip` and another writer's `.tmp` alone. The archive carried a real WAL-safe `library.db`, downloaded byte-identically, passed `unzip -t`, and restored through `/backup/import-file`. Icecast's `stream_start` did not move across either the backup or the restore. The decline path ran on a 64 MB tmpfs: no file written, no partial temp, no stamp recorded. Also settings validation and cold-load across a container restart. Added route tests over real HTTP, including the traversal guard on the download route. Suite is 1159 passing, lint clean.

Open:

- `09ba3da0` adds `GET /backup/file/:name` plus a Download button, so a scheduled snapshot can leave the box instead of only ever living on the disk it protects. That is additive rather than a review fix, so happy to drop it if you would rather it went in on its own.
- The panel was not driven in a browser, so the two client-side fixes above are verified server-side only.
- An `off` station never sweeps stale temps, since it returns before the readdir. Deliberate, but worth knowing.

PR #1586, opened 2026-09-06
feat(analyzer): ANALYZER_REPLICAS=0 turns the local analyzer container off (#1570)

perminder-klair:
`ANALYZER_REPLICAS=0` in the root `.env` turns the local analyzer container off. Closes the "Allow disabling the Analyzer package via .env" item in #1570 — the operator runs acoustic analysis on another machine via `ANALYZE_URL`, but the stack still started a redundant local analyzer.

`deploy.replicas: ${ANALYZER_REPLICAS:-1}` on the analyzer service in all three composes. The next `up -d` stops and removes the container, and it stays gone.

## Why not profiles

The obvious objection to a plain profile gate is that it needs `COMPOSE_PROFILES` set, breaking every existing install. But the clever inversion — `profiles: ["${ANALYZER_DISABLED:+off}"]`, empty by default — is broken in a way that isn't obvious:

```
unset, COMPOSE_PROFILES=tts-heavy  ->  analyzer GONE
```

A non-empty `COMPOSE_PROFILES` switches Compose from the default set to the named set, and an empty-string profile is in neither. `docs/tts-heavy.md` tells every Unraid/Portainer operator to set exactly that line, so that whole population would silently lose analysis on upgrade.

## Verification

- Matrix of unset / empty / `0` / `1` across all three files → `replicas: 1,1,0,1`.
- Default-on is genuinely unchanged: rather than only diffing the rendered config, the real container was brought up from `develop`'s compose and from the new one and `Config`/`HostConfig`/`Mounts` compared — identical. The rendered config gains one `replicas: 1` line, which is Compose's own default and a runtime no-op.
- Lifecycle: `up -d` with `0` removed a running container; back to `1` recreated it.
- GPU overlay: `deploy` merges key-wise, so `analyzer-gpu.yml`'s reservation survives.

A throwaway container name was used throughout — the live `sub-wave-analyzer` was never touched.

## Controller degradation — confirmed, one fix

Ran the controller against an unresolvable analyzer host: `resolveBackend` → null, `renderTransition`/`embedTexts` → null without throwing, `runAnalysisPass` → `{available:false}` after one informational line. Zero stderr, and 20 repeat probes cost 128ms total (a DNS miss fails fast, no 5s stall). Text tagging never consults the analyzer. No behavioural fix needed.

But that log line still named the tts-heavy sidecar, which hasn't been an analysis backend since the split — and it's the exact line this change makes operators read. It now names the analyzer sidecar, `ANALYZE_URL` and the local venv.

## Sharp edge worth a look

A non-integer value (`ANALYZER_REPLICAS=false`) is a hard Compose interpolation error on every command, not a silent fallback. The var is named for a count rather than a boolean specifically so nobody types `false` and can't boot — documented in `infrastructure.md`, but it is a real edge.

## Scope beyond the ask

`docs/unraid.md` (the most affected population) and the `docs/discord-open-suggestions.md` row for this thread, which still said "no issue, `ANALYZE_URL` is the workaround" — moved to shipped with the #1570 ref and the two counts adjusted. Revert either if that snapshot doc should be regenerated wholesale instead.

The AIO needs nothing — it runs the analyzer in-process via `ANALYZE_PYTHON`, not as a service. Stated explicitly in `.env.example`, `tts-heavy.md`, `unraid.md` and `infrastructure.md` rather than left silent.

CLI assets regenerated with `npm --prefix cli run embed-assets` (re-running is a no-op, so the staleness check passes). `controller`: lint clean, 1123/1123 tests passing. `web/` untouched.


perminder-klair:
Pushed three commits on top of the original (`1095cc86..e81fc4cc`).

**Review fixes**

- `docs/unraid.md` contradicted itself. Three lines above the new paragraph it still said an "off" acoustic engine means the container stopped and `Pull & Up` brings it back, which is precisely what keeps it gone after `ANALYZER_REPLICAS=0`.
- `infrastructure.md` had the new sentence welded onto the end of the previous paragraph with no separator, rendering as "throughput.The off switch".
- Only `0` and `1` are valid, and that is now documented. `ANALYZER_REPLICAS=2` fails *every* compose command with `can't set container_name and analyzer as container name must be unique`, not only `up`. The docs previously warned about the non-integer case alone.
- The no-backend log line named three sources for two channels. The sidecar *is* `ANALYZE_URL`, and `config.analyzer.urls` has been a single-entry list since the split, so it now reads `ANALYZE_URL sidecar / ANALYZE_PYTHON venv`.
- `README.md`'s env block listed `ANALYZER_HEAVY` but not the new switch.

**One behaviour change**

`ANALYZER_REPLICAS` is inert on the AIO, and setting it to `0` there fails worse than `ANALYZER_HEAVY` did: analysis keeps running in-process and keeps its RAM while the operator believes it is off, with the Library panel still reading on. That is the shape of #1300 bug 9, so it gets a warning rather than a docs line. `warn_if_analyzer_replicas_ignored` follows the existing `warn_if_analyzer_heavy_ignored` and names the knob that does work there, a blank `ANALYZE_PYTHON`. Two tests were missing and are added: one pinning that warning, one driving `deploy.replicas` across all three composes from a single table, verified to fail on a seeded `:-` to `-` mutation.

**Bug found while testing**

The docs promised the Library panel reads "acoustic engine off". No such string exists in the UI. The row is labelled `Acoustic analysis · bpm / key` and its value becomes `engine off`. Anyone searching the UI for the quoted phrase would come up empty. The behaviour behind the claim is fine; only the quote was wrong.

**How it was tested**

The compose render matrix (unset / empty / `0` / `1` → `1,1,0,1`), the GPU overlay merge at `replicas: 0`, and both hard-error cases ran against real docker. Controller degradation was driven through the real modules with `ANALYZE_URL` unresolvable and `ANALYZE_PYTHON` blank: `resolveBackend` returns null in 27ms, twenty repeat probes cost 64ms total, and `runAnalysisPass` returns `{available:false}` without throwing. Lint clean, 1129/1129 tests, CLI assets in sync.

The container lifecycle itself was **not** re-verified here (stop-and-remove, staying gone across `up` and `restart`, recreating at `1`, stream unaffected), because the dev machine had another stack running for the whole window. Your original verification covers those four.

**Open**

`resolveBackend` caches only non-null, so a *silently dropping* remote analyzer costs 5s per probe rather than the fast DNS-miss path measured above. Pre-existing and untouched here, probably worth its own issue given these docs point the remote-analyzer population at this config.

**Optional scope**

I moved the `docs/discord-open-suggestions.md` row from ✅ Shipped to 🔨 Tracked, since #1570 is open with the item still unticked and the file's own legend reserves Shipped for delivered-and-closed. Counts corrected to 6/13/20/26. Drop that hunk if you would rather regenerate the snapshot wholesale.

PR #1587, opened 2026-09-06
fix(library): count the library on request, not on every page load (#1570)

perminder-klair:
Opening the admin Library page no longer walks the whole Navidrome catalogue. Closes the library-check item in #1570 (Wylie B: *"it probably should not check the library unless you want to make that check. Maybe a button to trigger it instead"*).

## What was actually happening

`GET /library/coverage` looked cheap from the web side. `library-coverage.get()` had this (`controller/src/music/library-coverage.ts:115`):

```js
if (isStale() && !cache.scanning) refresh();
```

`refresh()` walks `subsonic.iterateAllSongs()` — one `getAlbum` HTTP call per album — roughly 2,000–3,000 sequential Navidrome requests on a 29k-track library. The cache is in-memory with a 6h TTL, so it was stale on every controller restart and again every 6 hours. A page view fired the whole walk, silently, in the background. That's the thing being watched happen.

The rest of the payload (`countTagged`, `analysedCount`, `vectorCount`, `analysisFailedCount`) is unindexed `COUNT(*)` scans over `library.db`, and `labelOnlyVectorCount` is a full-table scan memoised for 5 min. Real but small — those are what the 60s poll was actually paying for.

## The fix

**Controller** — `get()` is now a pure read: DB counts plus the last-known total stamped `scannedAt`. The walk belongs to `refresh()`, whose only callers are the operator's new `POST /library/coverage/refresh`, the existing `GET …?refresh=1` (kept for scripts; nothing in-tree uses it), and the two runs that have already walked the catalogue — a finished tagger run and a library reset. A POST rather than reusing `?refresh=1` precisely because a GET is the shape something eventually polls by accident.

**Web** — the coverage query still runs on mount (the read is cheap now) but nothing puts it on a timer except two live states: `tagger.running` (the 3s climbing-percentage behaviour, preserved exactly) and `scanning` on an in-flight count. `staleTime: 0` is gone so the client's 30s default applies per `web/CLAUDE.md` rule 4. One edge: `running` going false is what stops the poll, but that's also when the controller kicks its post-run recount — so a run ending refetches once and comes back with `scanning: true` to resume the 3s loop until the total lands.

**UI** — a Count library / Re-count button beside the track total (the `<Btn sm><RefreshCw size={11}/>` pattern from BlockedTab/HistoryTab) and the age via `relTime()`. An uncounted library reads "Library not counted yet" instead of "Library size unknown".

## One consequence worth a call at review

A fresh install shows no track total until the operator presses the button. That was chosen over keeping a one-per-process auto-scan, because a scan on first page view after every restart is still the reported bug. The first tagger run fills it for free, so the normal onboarding path never sits empty.

Left alone deliberately: the default Tracks tab's `/dj/recent?limit=50` (~50 bounded calls, constant regardless of library size — that's the content the operator asked for by opening the page), and `library.stats()` behind the separate 30s `/settings` poll (heavier, ~7 full scans on a 5s TTL, but it isn't "checking the library").

Documented in `docs/internals/music.md` and one root `CLAUDE.md` bullet — never reintroduce a scan on the read path.

## Green

- `web/`: lint clean (5 pre-existing warnings, none in touched files); `audit-admin-query.mjs` + its suite pass 9/9.
- `controller/`: lint clean; `npm test` 1123/1123.
- **Not run**: `verify-library.py` / `verify-query-cache.py` — they need an isolated running stack. Worth pointing at this before merge.

## Rest of the thread, not filed

The other tweaks in Wylie B's grab-bag couldn't be listed from here: both the issue body and `docs/discord-open-suggestions.md` truncate the post at exactly the quoted sentence. Someone with thread access needs to read it and split the rest out. Only the library-check item is implemented here.


perminder-klair:
Pushed five commits: the review fixes, plus two bugs that turned up in testing.

**Review fixes.** The count now persists to `state/library-count.json`. Without it an in-memory cache blanked the total, and every percentage derived from it, on each controller restart, and `scannedAt` could never read older than process uptime. `POST /library/reset` no longer refreshes coverage, since it wipes `library.db` rather than Navidrome and the total it recomputed cannot have changed. Dropped `?refresh=1` from the GET. Nothing in-tree used it, and a read that can start a walk is what eventually gets polled by accident. A failed count now records why (`scanError` on the payload, rendered beside the total) instead of only reaching the log. A tagger run kicks a count at start when nothing has ever been counted, so the first run's meter isn't stuck at "—". Both doc sites corrected; they described a library reset as a run that had already walked the catalogue.

**Bugs found while testing.** Coverage percentages were uncapped: a controller with a 1413-track `library.db` pointed at a 48-song music server rendered "2943%". The progress bars already clamped `aria-valuenow`, but the printed figure didn't, and the stale denominator this PR introduces makes the overrun ordinary rather than rare. Capped at the source. Separately, a tagger run that cannot spawn killed the controller: a ChildProcess `error` event with no listener is thrown, and only `exit` was handled, so `spawn npx ENOENT` took the broadcast down with it. Now reported as a failed run instead.

**Testing.** No docker dev stack, as the shared slot was held by another branch throughout. Used the isolated verify stack instead (controller on 7791, web on 7793) with a stub music server that counts every Subsonic call, so the costs are measured rather than assumed: 15 coverage reads cost 0 requests, a count costs one `getAlbum` per album, three simultaneous presses walk once, a reset costs 0, and the count survives a real restart carrying its original timestamp. Seven Playwright checks against the real Library page cover the button labels, the failure banner, the disabled state during a run, and the 3s poll taking the page from counting to done with no reload. `verify-library.py` 8/8, `verify-query-cache.py` 6/6, lint clean in both packages, `npm test` 1152/1152.

**Open, and optional scope.** The tagger spawn fix (4ca9806b) is pre-existing and unrelated to #1570, so it is a separate commit if you would rather drop it. Not covered: a count against a real Navidrome at full scale, and a completed tagger child run, since `tagger.ts` pins `cwd: '/app'` and the start-count guard was verified by its effect rather than through a finished run.

PR #1604, opened 2026-09-06
feat(web): heart and never-play on play-history rows (#1600)

### LINKED ISSUE #1600: Play history rows: add the heart and block controls
From the Discord #suggestions triage in #1568 — [thread](https://discord.com/channels/1514647956333002812/1534332709961338890) (BrentNorrisKY, 2026-08-04):

> Can we put a block/heart option on the play history? Would make it really easy to go back and mark stuff.

## What already works

Both halves exist, just not on this tab.

- Hearts shipped (#991): `POST /likes/song/:id`, `DELETE /likes/song/:id/operator` in `controller/src/routes/likes.ts`.
- The durable History tab shipped (#1065): `web/components/admin/library/HistoryTab.tsx`.
- The row-action cluster — heart toggle plus the three-way block menu (track / album / artist) — is already built in `web/components/admin/library/TrackTable.tsx` (`BlockMenu` at `:408`, heart at `:210` and `:351`).

## The gap

`HistoryTab.tsx` renders time, thumb, title, artist, source and a single queue button. Recognising a track you just heard and marking it means leaving the tab, switching to Browse, and searching for it — which is exactly the round trip the request is about.

## What to build

Lift the heart + block-menu cluster out of `TrackTable.tsx` into a shared piece and render it on history rows too. This should be a reuse job, not a second implementation — a duplicated block action is the kind of thing that drifts from the canonical one.

The one real decision: a `PlayEntry` can carry a null or empty `trackId` (a track that has since left the library). Those rows have nothing to act on and should render without the controls rather than with dead buttons.

Worth checking while in there: whether the heart state for history rows can be read from the same cache the Browse tab uses, so a heart set on one tab shows on the other without a refetch.

## Acceptance

- Heart and block (track / album / artist) are available directly on a play-history row.
- The action is the same code path as the Browse tab's — one implementation, not two.
- A history row whose track is no longer in the library degrades cleanly.
- Optimistic state matches the Browse tab's existing behaviour.

perminder-klair:
Closes #1600.

## What changed

The heart button and the three-way never-play menu are lifted out of `TrackTable.tsx` into a new `web/components/admin/library/row-actions.tsx`, and the Play history rows render the same two components.

- `row-actions.tsx` — `HeartButton`, `BlockMenu` and `likeStateFor`. `BlockMenu` and `likeStateFor` move verbatim; `HeartButton` is the same markup with its busy/disabled state taken as a prop instead of read off `TrackTableProps`. Both components take an explicit `Track` and a `className`, so `TrackTable` keeps its `hidden sm:inline-flex` collapse-to-overflow-menu layout while history rows show them at every width.
- `HistoryTab.tsx` — a `PlayEntry` is an air-time snapshot, so `historyTrack()` composes the `Track` the thumb and the two actions take. The Queue button drops to its icon below `sm:` now that there are three buttons where there was one.
- `HistoryTabContainer.tsx` — `likeIndex`, `liking`, `toggleLike`, `blocking` and `blockTrack` come straight off `LibraryContext`, the same handlers the Browse rows use.

## Why it's one code path

Both actions go through the provider, so a history row gets the Browse rows' optimistic like patch, the block toast's Undo, and the shared `['library','likeIndex']` entry — a heart set on either tab shows on the other with no refetch, and every history row for the same track lights up at once. The phone overflow menu in `TrackTable` still owns its own copy of the menu-item markup; that is a different control (it also carries queue/edit/retag/clear-likes), not a second block implementation.

## Rows with no track id

A `PlayEntry` can carry a null `trackId`, and it means one specific thing: the annotated URI that aired carried no `subsonic_id` — untracked auto-playlist plays, mainly, per the `sourceTrackId` note in `broadcast/queue.ts`. Nothing ever nulls the column after the fact, so it says nothing about whether the track is still in the library.

Those rows render a muted dash titled *"no track id recorded for this play"* instead of the action cluster: hiding the permanently-disabled Queue button is right, but the sentence it carried was doing work — without it an operator cannot tell a play with no id from buttons that failed to render.

A track that has since **left** the library is the other case, and it is not handled here: the play keeps its id, so the row shows all three actions and they fail at the server (block 404s; the heart records against a dead id). Detecting that needs a resolve check the row does not have.

## Two things deliberately left alone

- History rows carry no `blockedBy` stamp, so a track you have already blocked still shows the block menu rather than an Unblock button. Stamping them would mean filing the history query under the `['library','rows']` key family, which `HistoryTabContainer` documents as wrong — a `PlayEntry` is not a `Track`. The cost is in the follow-ups below.
- The block/heart handlers are untouched; this is a render-side change only.

## Follow-ups, not in this PR

Each of these is real, none is caused by the extraction, and all three want their own change.

- **The heart from History stores a thinner snapshot than the same heart from Browse.** `resolveSnapshot` (`routes/likes.ts`) short-circuits on `if (b.title || b.artist)` and returns the body's `genre`/`year`/`duration`, on the stated premise that "the admin library always posts the row it already has". A history row only has `{id, title, artist, album}`, so the body still trips that branch and the `db.getTrack` fallback that would have filled the other three never runs. Nothing breaks — `/library/liked` re-enriches from the index, and the picker's `listener-liked` pool treats unknown length the way `belowTrackFloor` treats every unmeasured row — but the stored snapshot is poorer than it needs to be. The fix belongs in `resolveSnapshot` (merge rather than short-circuit), not at this call site.
- **An already-blocked history row offers Block and gets a 409, with no route back off the list.** `TrackTable` swaps the menu for a one-click Undo whenever `blockedBy` is set; a history row can never take that branch, so re-blocking returns `already blocked` as an error toast. (Checked: the 409 throws before `notify.undo`, so there is no hazard of the undo toast lifting a block the operator did not create.) Worth solving alongside whatever gives history rows a block state.
- **No `verify-library.py` check covers the new surface.** `like_toggle` and `blocked_marks` were each added when these controls landed on Browse, and the cross-tab property this PR claims is exactly what they exist to pin. Deliberately not written here rather than written unverified. Note for whoever adds it: history rows have no `.lib-row`, and both existing checks scope their selectors under it, so it needs its own scope rather than a reuse.

## Verification

`npm run lint` in `web/` is clean (only the five pre-existing warnings in `playlist-builder/generate.ts` and `ai-elements/file-tree.tsx`). `node --test web/scripts/audit-admin-query.test.mjs` passes 9/9. `node web/scripts/audit-admin-query.mjs` reports one violation — `components/admin/settings/BrainSection.tsx:162`, pre-existing on `develop` and untouched here. No controller changes, so no `npm test` run.

### DIFF
--- a/web/components/admin/library/HistoryTab.tsx
+++ b/web/components/admin/library/HistoryTab.tsx
@@ -22,10 +22,13 @@ function playDayLabel(iso: string): string {
 }
 
 // A play is an air-time snapshot, not a library row, so the thumb and the shared
-// row actions get a Track composed from it. `trackId` is null for a track that has
-// since left the library: the thumb still gets the words for its letter tile, but
-// there is nothing left to heart, block or queue, so those rows render no action
-// cluster at all rather than three dead buttons.
+// row actions get a Track composed from it. `trackId` is null when the annotated
+// URI carried no `subsonic_id` — untracked auto-playlist plays, mainly (see the
+// `sourceTrackId` note in broadcast/queue.ts). Nothing is ever nulled after the
+// fact, so this says nothing about whether the track is still in the library: a
+// removed track keeps its id here and its actions fail at the server. The thumb
+// still gets the words for its letter tile, but an id-less play has nothing to
+// heart, block or queue, so it renders the dash below instead of dead buttons.
 function historyTrack(p: PlayEntry): Track {
   return {
     id: p.trackId || '',
@@ -113,7 +116,13 @@ export function HistoryTab({
                     <span className="hidden w-24 shrink-0 text-right text-[11px] text-muted sm:block" title="how it was picked">
                       {playSourceLabel(p)}
                     </span>
-                    {p.trackId && (
+                    {!p.trackId ? (
+                      /* Says WHY the row is action-less. Without it an operator
+                         cannot tell a play with no id from buttons that failed to
+                         render — which is the work the old disabled Queue button's
+                         tooltip was doing. */
+                      <span className="shrink-0 text-[11px] text-muted" title="no track id recorded for this play">—</span>
+                    ) : (
                       /* Three buttons where there was one, so Queue drops to its icon
                          below sm: — the row still has to leave the title readable on a
                          phone. */
--- a/web/components/admin/library/HistoryTab.tsx
+++ b/web/components/admin/library/HistoryTab.tsx
@@ -8,8 +8,9 @@ import { num } from '../LibraryTaggingPanel';
 import { SkeletonRows } from '@/components/ui/skeleton';
 import { EmptyState } from '@/components/ui/empty-state';
 import { PAGE_SIZE } from './types';
-import type { PlayEntry, Track } from './types';
+import type { BlockType, LikeIndex, PlayEntry, Track } from './types';
 import { Thumb } from './bits';
+import { BlockMenu, HeartButton, likeStateFor } from './row-actions';
 
 function playDayLabel(iso: string): string {
   const d = new Date(iso);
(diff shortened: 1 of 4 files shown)


perminder-klair:
Reviewed the diff against `develop`. The extraction is clean — `HeartButton` and `BlockMenu` moved verbatim, both callers go through `LibraryContext`, so history rows genuinely share the Browse rows' optimistic patch and the one `['library','likeIndex']` entry rather than getting a parallel implementation. That was the main risk in this issue and it's avoided.

**One thing to settle before merge — it's cross-PR**

`BlockMenu` carries this line under "Never play this artist":

> primary credit only — collabs filed under other artists still play

#1608 (open, lint-green, same base) makes that false: after it, an artist block reaches tracks the artist only guests on. Neither PR updates the copy, so whichever merges second ships a user-visible statement that contradicts the behaviour.

It's in two places on `develop` — `TrackTable.tsx:394` (the phone overflow menu) and `TrackTable.tsx:447` (the block menu). This PR moves the second one into `row-actions.tsx` and leaves the first where it is, so after both land the stale sentence lives in two different files. Worth deciding now which PR owns the fix; my suggestion is #1608, since it's the one changing the behaviour, but it needs to know this PR relocated one of the two copies.

**Minor: rows with no `trackId` lost their explanation**

Before, such a row showed a disabled Queue button titled *"no track id recorded for this play"*. Now the whole cluster is omitted, so the row is silently action-less and an operator can't distinguish "this play has no track id" from "the buttons failed to render". Hiding the dead buttons is the right call, but the sentence was doing work — a muted dash or a `title` on the row would keep it.

**Checked and fine:**

- `historyTrack()` composes `id: p.trackId || ''`, and the action cluster is guarded by `p.trackId &&`, so no action ever runs with an empty id. The `Thumb` still gets the words for its letter tile.
- Leaving history rows without a `blockedBy` stamp is the right trade, and the reasoning in the PR body is the same one `HistoryTabContainer` already documents — a `PlayEntry` is not a `Track` and shouldn't be filed under the `rows` key family.
- `disabled={!!blocking}` on the history `BlockMenu` matches the Browse behaviour, so a block in flight locks both tabs the same way.

Approving once the copy question has an owner.

PR #1605, opened 2026-09-06
feat(web): add a custom era range to the show editor (#1599)

### LINKED ISSUE #1599: Show era picker: allow a custom from/to year range, not just decade chips
From the Discord #suggestions triage in #1568 — [thread](https://discord.com/channels/1514647956333002812/1536356905742631003) (PwrBlackout, 2026-08-10):

> i have a "new music monday" show that i want to only play 2026 music. the API already allows me to manually pass in "fromYear" and "toYear", the UI doesn't.

## What already works

The data model is complete. A show's `eras` is an array of `{ fromYear, toYear }` windows, validated in `controller/src/schemas/show.ts:174` (with the `fromYear <= toYear` check and the legacy scalar migration at `:218`). The API accepts arbitrary windows today.

## The gap

`web/components/admin/shows/ShowEditor.tsx` only offers the `DECADES` chip row (`:484`). A window that matches no preset is rendered read-only — the block at `:499` exists purely so an API-set custom window stays *visible and removable*, with the comment "Custom windows (set via the API — no preset matches)". There is no way to **add** one from the admin UI, so "2026 only" is unreachable without hand-writing a PATCH.

## What to build

An add-a-range control beside the decade chips: two year inputs (either side optional — an open-ended `fromYear: 2026, toYear: null` is a legal window) and an Add button that pushes onto the same `eras` array the chips write to. The existing removable custom chip already renders the result via `eraLabelOf`, so the display half is done.

Notes for whoever picks it up:

- Reuse the schema's bounds rather than restating them — `showYear` in `controller/src/schemas/show.ts` owns the valid range, and it is mirrored into `web/lib/schemas.generated.ts`.
- A range that happens to equal a decade should light the matching chip rather than adding a duplicate custom window; `sameEra` already answers that question.
- `FILTER_VALUES_MAX`-style caps: check whether `eras` has its own limit in the schema before adding a second one in the UI.

## Acceptance

- An operator can set a show to 2026-only from the admin UI, save, and see it survive a reload.
- The decade chips and custom ranges coexist in one `eras` array; toggling a chip does not drop a custom window.
- No schema change and no controller change — this is a frontend-only gap.

perminder-klair:
Closes #1599.

## What changed

`ShowEditor`'s eras field now has an **add-a-range** control beside the decade chips: two year inputs and an *Add range* button, writing onto the same `eras` array the chips write to. That makes "new music Monday, 2026 only" reachable from the admin UI — previously it needed a hand-written PATCH, and an API-set window rendered read-only (visible and removable, but not addable).

Either side may be left blank, since an open-ended `{fromYear: 2026, toYear: null}` is a legal window. Adding a range that spells out a decade is refused rather than appended, so it lights that chip instead of stacking a duplicate beside it — `sameEra` already answered that question.

## Why it's shaped this way

- **The year test is the schema's own, imported — not restated.** `validEraYear` is now `export`ed from `controller/src/schemas/show.ts` and rides the generated mirror into `resolveEraDraft`, replacing the local `Number.isInteger(n) && n >= YEAR_MIN && n <= YEAR_MAX`. `types.ts` still re-exports `SHOW_YEAR_MIN`/`SHOW_YEAR_MAX`, but only for the `<input min/max>` and the error copy — the *rule* has one definition. The editor keeps its own `trim()`, deliberately: `eraYearOf` on the wire does **not** trim (`' '` is a malformed post, not an open end) while a draft box legitimately holds whitespace mid-keystroke, so the two were never quite the same predicate and the restatement was live drift, not just duplication. `web/lib/schemas.generated.ts` was regenerated with `cd controller && npm run gen:schemas` and committed; it is never hand-edited.

- **The cap prop is required BY this feature, not polish alongside it.** `eras` is already capped at `SHOW_FILTER_VALUES_MAX` by the schema, so no second limit was added. But the decade `ChipRow` caps on what *it* has selected, which cannot see custom windows. `SHOW_FILTER_VALUES_MAX` is **15** and `DECADES` has **8** entries, so that row alone can never reach the cap — an earlier version of this description said the chips would offer "a ninth window", which is not the boundary. The window at risk is a **sixteenth**. Before the add-a-range control, filling `eras` took 15 API-set windows; now the operator can fill it entirely from this UI, at which point an uncapped chip row offers that sixteenth, the schema refuses the save, and nothing on screen says why. Hence `cap={FILTER_VALUES_MAX - customEras.length}`. For any show with no custom windows, `cap` is 15 — the `ChipRow` default — so the behaviour is unchanged.

- **The draft resolver is pure** (`resolveEraDraft` in `shows/types.ts`, next to `sameEra`/`eraLabelOf`) and returns a reason rather than throwing; the error only appears on Add, so a half-typed `20` doesn't scold anyone mid-keystroke.

- **Every write to `eras` clears the draft error.** The message is a verdict on one Add press against the array *as it stood*: with the 90s chip lit, typing `1990`/`1999` is refused as a duplicate — but untoggling that chip used to leave "That range is already selected." standing, inside a live region, over a range that would now add cleanly. All three writers (Add, the chip `onToggle`, the custom-chip remove) now go through one `setEras` helper that clears it, rather than each handler remembering to.

- The two inputs carry their own `aria-label` — the eras group title names the whole field, not either box.

## Verification

**Run here, passing:**

- `npm run lint` in `web/` — 0 errors (5 pre-existing warnings, none in changed files).
- `npm run lint` in `controller/` — 0 errors (pre-existing `no-explicit-any` warnings only).
- `cd controller && npm run gen:schemas` produces a file **byte-identical** to the committed `web/lib/schemas.generated.ts`, so CI's mirror-drift job is satisfied.
- `cd controller && npm test -- schema` — 352/352.

**Written and NOT run: `web/scripts/verify-forms.py shows`.**

The check now drives the whole flow — toggle the 90s chip; confirm re-adding `1990–1999` is refused as a duplicate; add the single-year window `2026–2026`; add the **open-ended** window `2030+` (one bound blank — the only shape `eraLabelOf` renders through its no-`toYear` branch, and the only one that exercises `parse()` returning `null` on one side); untoggle the 90s chip and assert both custom windows survive it (#1599's second acceptance criterion, which had no assertion behind it); re-light it; save; and read the show back through the controller as `[{2026,2026},{2030,null},{1990,1999}]`.

None of that has been executed against this branch. `verify-forms.py` needs the isolated verify stack (controller `:7791`, web `:7793`) plus `SUBWAVE_VERIFY_ALLOW_DESTRUCTIVE=1`, and it makes destructive writes — it cannot run in CI or from a bare checkout. **Before merge**, someone with the verify stack up should run:

```
SUBWAVE_VERIFY_ALLOW_DESTRUCTIVE=1 python3 web/scripts/verify-forms.py shows
```

`web/CLAUDE.md` is explicit that this run is the evidence for a forms PR. Treat the PR as unverified until its output is posted here.

## Docs

`docs/discord-open-suggestions.md` still carried this thread as *"🟡 Partly covered — … Exposing the API's fromYear/toYear as a show-level UI control is not tracked."* Moved to **🔨 Tracked** against #1599, with the at-a-glance counts adjusted (13→14 Tracked, 20→19 Partly covered).

### DIFF
--- a/controller/src/schemas/show.ts
+++ b/controller/src/schemas/show.ts
@@ -160,8 +160,15 @@ function showStringList(opts: {
 // the load path's repairEraWindow (below) so the two can never disagree about
 // what a valid year is. null / '' means "open end". A numeric string is
 // accepted because that is what an <input type="number"> posts.
+//
+// `validEraYear` is EXPORTED so it rides the mirror into the admin show
+// editor's add-a-range control (#1599), which has to refuse a year the save
+// would then reject. It owns only the integer-and-range test; the editor keeps
+// its own trim, because eraYearOf deliberately does not trim (' ' reaching the
+// wire is a malformed post, not an open end) and a draft box legitimately holds
+// whitespace mid-keystroke.
 const eraYearOf = (v: unknown): number | null => (v == null || v === '' ? null : Number(v));
-const validEraYear = (n: number | null): boolean =>
+export const validEraYear = (n: number | null): boolean =>
   n == null || (Number.isInteger(n) && n >= SHOW_YEAR_MIN && n <= SHOW_YEAR_MAX);
 
 const showYear = z
--- a/docs/discord-open-suggestions.md
+++ b/docs/discord-open-suggestions.md
@@ -9,8 +9,8 @@ Captured 2026-09-06 — 65 of 125 threads, checked against all 334 issues (28 op
 | Status | Count | Meaning |
 | --- | ---: | --- |
 | ✅ Shipped | 6 | Delivered on `develop`; the thread can be closed out. |
-| 🔨 Tracked | 13 | An **open** GitHub issue covers it. |
-| 🟡 Partly covered | 20 | Related work has landed, but the specific ask has not. |
+| 🔨 Tracked | 14 | An **open** GitHub issue covers it. |
+| 🟡 Partly covered | 19 | Related work has landed, but the specific ask has not. |
 | ⬜ No issue | 26 | Nothing on GitHub — needs filing or an explicit reply. |
 
 Open GitHub issues doing the most work here: [#1485](https://github.com/perminder-klair/subwave/issues/1485) (the follow-up tracker, 5 threads), [#1486](https://github.com/perminder-klair/subwave/issues/1486) (CarPlay + Android Auto, 2 threads) and [#1400](https://github.com/perminder-klair/subwave/issues/1400) (skill-aimed banter, 2 threads).
@@ -501,9 +501,9 @@ PwrBlackout · 2026-08-10
 
 > i have a "new music monday" show that i want to only play 2026 music. the API already allows me to manually pass in "fromYear" and "toYear", the UI even shows that…
 
-**🟡 Partly covered** — Era resolution itself was fixed (#842, #1418). Exposing the API's fromYear/toYear as a show-level UI control is not tracked.
+**🔨 Tracked** — #1599 is the open issue, and it is exactly this ask: an add-a-range control beside the show editor's decade chips, so a single year ("2026 only") or an open-ended window is reachable from the admin UI rather than only through a hand-written PATCH. Implemented and in review, not yet on `develop`. Era resolution itself was fixed earlier (#842, #1418). Moves to ✅ Shipped when #1599 closes.
 
-Refs: [#842](https://github.com/perminder-klair/subwave/issues/842) · [#1418](https://github.com/perminder-klair/subwave/issues/1418)
+Refs: [#1599](https://github.com/perminder-klair/subwave/issues/1599) *(open)* · [#842](https://github.com/perminder-klair/subwave/issues/842) · [#1418](https://github.com/perminder-klair/subwave/issues/1418)
 
 ### [Skill-blocks, triggered at a specific time](https://discord.com/channels/1514647956333002812/1536054559045062866)
 
--- a/web/components/admin/shows/ShowEditor.tsx
+++ b/web/components/admin/shows/ShowEditor.tsx
@@ -44,7 +44,10 @@ import {
   TAG_RE,
   TOPIC_MAX,
   VOCAL_OPTIONS,
+  YEAR_MAX,
+  YEAR_MIN,
   eraLabelOf,
+  resolveEraDraft,
   sameEra,
 } from './types';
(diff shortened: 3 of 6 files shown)


perminder-klair:
Reviewed the diff against `develop`. This does what #1599 asked and stops where the issue said to stop — no schema change, no controller change.

**Verified rather than taken on trust:**

- The cap claim holds: `eras` really is `.max(SHOW_FILTER_VALUES_MAX)` at `controller/src/schemas/show.ts:375`, so passing `cap` to the chip row instead of adding a second limit is right.
- `ChipRow` already had `cap?: number` and computes `atCap = !on && selected.length >= cap`, so a lit chip is never disabled — an operator at the cap can still deselect their way out. Worth stating because a naive cap here would be a trap.
- The decade-equivalent case behaves as the issue asked in *both* directions: typing 1990–1999 with the chip unlit appends `{1990,1999}`, which `DECADES.some(sameEra)` then classifies as a preset, so it lights the chip rather than adding a custom entry. With the chip already lit it's refused. Good.

**One thing worth changing**

`resolveEraDraft` restates the schema's year predicate — `Number.isInteger(n) && n >= YEAR_MIN && n <= YEAR_MAX` is exactly `validEraYear`. Importing the bounds and then re-deriving the test around them is half the rule.

The reason it's written this way is real: `eraYearOf` and `validEraYear` are in the mirror (`web/lib/schemas.generated.ts:3870-3872`) but declared as bare `const`, so they aren't importable. Exporting `validEraYear` from `controller/src/schemas/show.ts` and re-running `npm run gen:schemas` would close it in two lines.

I checked the two for divergence and there is none today — both sides do `Number()` over a trimmed string, both treat empty as `null`, both demand an integer in range, and both accept the same odd inputs (`Number('0x7CE')` is 1998 on either side). So this is drift-prevention, not a bug, and it does not need to block the merge.

**Minor**

The `cap={FILTER_VALUES_MAX - customEras.length}` line is a latent bug fix, not part of the feature — the chip row would have offered a ninth window on a full array. It's only reachable because custom windows became addable, so shipping it here is right, but it deserves a line in the commit body so it isn't invisible later.

**Not a complaint:** no unit test for `resolveEraDraft`, because `web/` has no test runner by design — the `verify-forms.py` extension is the right home for it, and driving a real POST/read-back round trip is a stronger assertion than a unit test would have been.

Approving as-is.

Issue #1282, opened 2026-08-03
[bug] - clock links ~2min ahead of air time; hourly check ignores minutes-elapsed


tinkermesomething:
## Related prior issues

- **#864** (closed) is the original bug this traces back to: links used to speak a *stale* clock (behind actual time). The fix added a look-ahead (`showAt`), but it reuses a constant meant for something else, overcorrecting in the opposite direction. Looks like a regression from #864's own fix, not a new failure mode.
- **#1258** - unrelated (idents/links stacking audibly, not clock accuracy).
- **#418, #205, #353** - older, whole-hour/timezone-flavored offsets, different cause.
- No hits searching issues for `PICK_SHOW_LOOKAHEAD_SEC` or `"just gone"`.

---

## Bug 1: pick-attached links are ~2min ahead

Links (generated while the current track is still playing) state a clock time consistently **~2 minutes ahead** of when the line actually airs.

Live examples:
- Logged `08:48:01`, says "Local time eight fifty" (08:50).
- Reported separately: stated 10:54 at actual 10:52.

**Cause** — `broadcast/queue/pure.ts`:
```ts
// Padding for SHOW-BOUNDARY attribution (deciding which show owns a track
// near an hour changeover), calibrated to ~half a typical track length.
export const PICK_SHOW_LOOKAHEAD_SEC = 120;
```

Reused in `broadcast/dj-agent.ts` as generic clock look-ahead padding, unrelated to its purpose:
```ts
showAt = new Date(Date.now() + (leadSec + PICK_SHOW_LOOKAHEAD_SEC) * 1000);
const ctx = await getFullContext(showAt ?? undefined);
link = await dj.generateLink({ ..., context: ctx, clockIsAirTime: !!showAt });
```

`llm/internal/prompts/scripts.ts`'s `generateLink` then explicitly permits the model to speak `ctx.clock` as the real air time whenever `clockIsAirTime` is true:
```ts
const clockClause = clockIsAirTime
  ? ` If you mention the clock, "Local time" below is the moment this link airs — use that, never an earlier time.`
  : ` Never state the clock time — this line airs when the next track starts, and you can't know exactly when that is.`;
```

Since `leadSec` (the real remaining track time) already gets the full 120s show-boundary constant added on top before reaching this gate, every link's stated time overshoots by exactly that paddin, matching both examples almost to the second.

[...]

PR #1607, opened 2026-09-06
feat(dj): vary the hourly time check's wording, not its reading (#1602)

### LINKED ISSUE #1602: The hourly time check sounds identical every hour — let the DJ vary the phrasing
From the Discord #suggestions triage in #1568 — [thread](https://discord.com/channels/1514647956333002812/1534429009717235816) (Gurthyy, 2026-08-05):

> So over in #ask-the-bot, @PwrBlackout was asking the bot about the 'Just gone 6pm' type callout that always seems to sound exactly the same

## Why it happens

This is not a bug in the clock — the clock is right, and repeatedly fixed (#1282, #1314, #1286, #1380). It is a direct consequence of how accuracy was bought.

`spokenTimePhrase()` (`controller/src/time.ts:169`) rounds the wall clock into **one of six fixed phrases**:

    m <= 4   → "just gone <hour>"
    m <= 14  → "just after <hour>"
    m <= 24  → "quarter past <hour>"
    m <= 39  → "half past <hour>"
    m <= 49  → "quarter to <hour+1>"
    else     → "coming up on <hour+1>"

and the hourly prompt then hands that exact string to the model with an instruction not to deviate (`controller/src/llm/internal/prompts/scripts.ts:387`):

> The time to announce is "<spokenTime>" — say exactly that time, in natural spoken words — never digits or 24-hour form, never a different time.

Since the check normally rides the `:00` cron, the minute band is nearly always the first one — so every hour of every day opens with the identical five words. The model is not being repetitive; it is being obedient.

## The constraint that must survive

The rounding lives in code for a measured reason: small models get the 24-hour conversion wrong at the edges ("00:03" announced as "one in the morning"), and a manual `/dj/segment` trigger at 18:31 used to say "just gone six" (#1282). **The code must keep owning which time is announced.** This issue is about the wording, not the reading.

## What to build

Keep the rounding exactly where it is, but stop pinning the wording to a single string. Roughly: each minute band gets a small set of equivalent phrasings for the same rounded time, and the prompt offers the band rather than dictating one member of it — so the DJ can vary and stay in character while the announced time remains the one the code chose.

Things to get right:

- Every phrasing in a band must mean the same rounded time. "just gone six" and "a few minutes past six" are interchangeable; "six" and "around six" are not equally safe at `m = 4`.
- The hour word still comes from `spokenHourPhrase`, including its day-edge normalisation (23:50 reads "coming up on midnight") — do not let a phrasing variant reintroduce raw hour arithmetic.
- The persona should be able to colour the delivery; the *time* is not the persona's to change.
- Avoid the obvious near-miss of letting the model pick freely with the phrase as a "hint" — that is what #1282 fixed.
- `spokenTimePhrase` is used from more than the hourly check; check every caller before changing its return type rather than assuming it is a private helper.

## Acceptance

- Consecutive hourly checks do not open with the same words.
- The announced time is still the one the code rounded to, at every minute band and at the day edge.
- Existing time tests still pass, and the new variation is covered by a test that pins each band's phrasings to the same rounded time.

perminder-klair:
Closes #1602.

## What changed

`spokenTimePhrase()` rounds the wall clock into six fixed minute bands and the hourly prompt hands the model that one string with "say exactly that time … never a different time" (#1282). The check rides the `:00` cron, so it lands in the first band nearly every time — every hour of every day opened with the identical five words.

- **`time.ts`** — the six bands become a table (`TIME_BANDS`), each carrying several equivalent wordings of the one rounded time. `spokenTimePhrases(h, m)` returns the band, canonical form first; `spokenTimePhrase(h, m)` returns that first form, so it is byte-identical for every existing caller (`context.ts` was the only one, plus the tests). The hour word still comes from `spokenHourPhrase` for `h` or `h+1`, so the day-edge normalisation is untouched.
- **`context.ts`** — `clock.spokenTimeOptions` alongside the unchanged `clock.spokenTime`.
- **`prompts/context.ts`** — the "never the same entry twice running" rule behind `pickAngle` is now one `pickVaried` helper, with `pickTimePhrase` as the second caller. `pickAngle` behaves identically.
- **`prompts/scripts.ts`** — the time clause is extracted as `nextHourlyTimeClause(clock)` (pure enough to test, like `linkPrompt`): it picks one wording and dictates it in the **unchanged** sentence. Both old fallbacks (a context predating the options, then one predating `spokenTime`, then a bare context) are preserved verbatim.

## One deviation from the issue

The issue suggests the prompt offer the band and let the model pick. This picks in code and still dictates a single string, for two reasons: a time clause that lists options is exactly the latitude #1282 removed (and small models echo lists), and a model handed a list leans on its first entry — which is the repetition being fixed. The band is real either way; only the point of selection differs, and the acceptance criterion ("consecutive checks do not open with the same words") is then guaranteed rather than hoped for.

## Constraints kept

The code still owns **which** time is announced: every wording in a band is interchangeable at every minute in that band — **including the minute the band opens on** — and no variant re-derives the hour.

### The two forms that overclaimed their band's first minute

Two wordings asserted more precision than their band's **first** minute, which the table's own rule forbids: `a minute or so past X` in the `:00`–`:04` band (and `:00` is the minute this row's cron fires on almost every time), and `gone quarter past X` in the `:15`–`:24` band — the latter being exactly the construction the `:25`–`:39` band already refuses as "gone half past". The inconsistency was the defect, not either form on its own.

Both are **dropped rather than annotated**. The tempting defence is the broadcast buffer: a listener hears the stream `stream.bufferSeconds` late (22s default) plus voice-queue latency, so on air both forms are usually true. That defence is rejected deliberately — `stream.bufferSeconds` is an operator dial defined in another module, so a wording that is only honest because of it stops being honest the moment the dial moves. A form has to stand up against its own band.

The `:15`–`:24` band gains **`around quarter past`** so that no band falls below **three** forms: with the no-repeat picker a two-form band alternates deterministically, which is a different fixed pattern rather than variation. "Around" is also the safe direction — it widens the claim rather than sharpening it, and reads true across the whole `:15`–`:24` span.

The rule comment now states "including the minute it opens on", and every band with an obvious near-miss carries a note naming the form it **refuses** and why — the refusals are the part of the table a new form gets checked against.

### `spokenTimeOptions` no longer rides out on `/now-playing`

`GET /now-playing` returns the context object whole, so the band the prompt is deliberately never shown was going out to every listener on a 5s poll. `clock.spokenTimeOptions` is the picker's raw material, not a fact about the moment, and the root rule is that a public read never widens to carry a behaviour internal.

It is now stripped in `routes/public.ts`, at that one public boundary. The **context type is untouched**, so every in-process prompt caller still reads the band from the same `getFullContext` it already holds — no threading, no second shape. `clock.spokenTime` stays public: it is a reading, and skins have always seen it.

While in there: `getClockContext` was building the band **twice** per call — `spokenTime` re-walked the table for a value already in hand — on that same per-listener path. It now builds it once.

**Honest caveat:** the strip is guarded by comments on both sides (`context.ts` marks the field controller-internal and points at the strip; `routes/public.ts` says why), **not by a test**. Pinning it needs a route-level harness for `/now-playing`, which is more machinery than the field is worth; a future field added to the clock context could regress it.

### `hourlyTimeClause` → `nextHourlyTimeClause`

The function is not the pure formatter its old name implied: picking advances the no-repeat rotation in `prompts/context.ts`, so calling it to preview or log a clause **spends a wording**. The name now warns. (Advancing on a call whose generation then fails is harmless in the only direction that matters — it can skip a wording, never repeat one — but there is exactly one production caller and it should stay that way.)

## Tests

- `scripts/clock-phrase.test.ts` — existing pins unchanged, plus: the canonical wording is still `forms[0]` at all 1440 minutes; every wording in a band names the same hour as the canonical one and no other; each band's forms pinned literally; the day edge normalises in every wording; the per-band form floor is pinned at three.
- `scripts/hourly-time-phrasing.test.ts` (new) — the clause quotes exactly one wording, always from the current band, never the same one twice running, and never borrows from a band it has just left.

Two follow-up fixes to the tests themselves:

- **All three fallback clauses are now `assert.equal` against the full string.** Two of them were asserted with `startsWith`, and the tail a prefix check skips is the #1282 instruction ("never digits or 24-hour form, never a different hour") that this whole feature exists to leave intact — a change that dropped it would have passed.
- **The per-band snapshot test was retitled to stop claiming a check it does not perform.** It was called "a new one must be equivalent at every minute in it", but its body is six `deepEqual` snapshots of literal arrays: it pins the lists so a change reads as a deliberate diff, and verifies nothing about equivalence. That gap is how both near-miss forms got in. The only half that *is* machine-checked is the hour word, by the test above; equivalence stays a review-time human check, and the title and comment now say so.

## Verification

`npm test` 1368/1368 pass, `npm run lint` exit 0 (`eslint . && tsc --noEmit`; warnings are the repo's existing `no-explicit-any` baseline) in `controller/`. Reasoning documented in `docs/internals/broadcast.md`.


perminder-klair:
Reviewed the diff against `develop`. The shape is right: the rounding stays in `time.ts`, the prompt still dictates exactly one string, and `spokenTimePhrase` returning `spokenTimePhrases(...)[0]` keeps every existing caller byte-identical — I checked, `context.ts` was the only one outside the tests.

**On the deviation from the issue: you're right and I was wrong.**

I wrote "offer the band and let the model pick". Picking in code is better for the reason you give — a clause listing options is the latitude #1282 removed — and there's a second reason worth having in the record: the acceptance criterion becomes a property of the code rather than a hope about the model. The test that asserts consecutive clauses never repeat could not exist under the version I asked for.

**Two things, neither blocking**

**1. `hourlyTimeClause` looks pure and isn't.** It advances `lastPickIdx` through `pickTimePhrase` on every call, so calling it to preview or log a clause would consume a rotation step. With one production caller that's harmless, but the name gives no warning. Either say so on the export or name it for the side effect.

**2. `a minute or so past {h}` in the `m <= 4` band.** The rule you wrote — every form interchangeable at *every* minute in its band — is the right rule, and the tests pin it. This form arguably breaks it at `m = 0`, which is the minute the check almost always fires on: at exactly six o'clock, nothing is yet "a minute or so past" it.

The saving argument is that a listener hears it `stream.bufferSeconds` late (22s default) plus whatever the voice queue adds, so on air it's usually true. But that's an argument from the buffer, not from the band, and it's the one form in the set whose honesty depends on a number defined somewhere else. Either drop it or write that reasoning next to it, so the next person adding a form doesn't take it as licence to add "five past".

**Nit, no action:** `pickVaried` stores an index per key, and the time-phrase key is shared across bands whose lists differ in length (4/3/3/3/4). After a band change the stored index refers to a different list, so you get an occasional spurious avoidance. It can't produce a wrong wording — `pickVaried` only ever indexes into the list it was handed — so it's cosmetic.

**Checked and fine:** the `:25`–`:39` band correctly refuses "gone half past"; the day-edge test covers every form and not just the canonical one; `pickTimePhrase(...) ?? clock?.spokenTime` degrades correctly for an empty options array; both legacy fallbacks are asserted byte-for-byte.

Approving.

PR #1608, opened 2026-09-06
fix(blocklist): an artist block reaches the tracks they only guest on (#1603)

### LINKED ISSUE #1603: Blocking an artist doesn't block tracks where they are only featured
From the Discord #suggestions triage in #1568 — [thread](https://discord.com/channels/1514647956333002812/1534144640520028300) (Solo_mag, 2026-08-04):

> I'm not sure what the best way to handle this is, but blocking each track individually doesn't seem like the best option.

Blocking an artist blocks the tracks they are credited as leading. A track credited to someone else "feat. <blocked artist>" still airs, so the operator has to block those one by one.

## What already exists

Collaboration-aware artist keys landed for the **repeat guard** in #1251 / #1406, in `controller/src/music/recency.ts`:

- `artistKey()` (`:113`) — the identity key.
- `FEATURE_SPLIT` (`:120`) — a "featuring" matcher, noting that `feat.` / `ft` / `featuring` is "always a credit on someone else's track, never part of an artist's own name".
- `leadArtistKey()` (`:149`) — collapses a credit onto its **primary act**, so "Kanye West (feat. Jay-Z)" keys as "kanye west".

Note the direction: that work exists so a collaboration counts against the artist who **leads** it. #1425 is the open follow-up there, and it is about the repeat guard, not the blocklist.

## The gap

`controller/src/music/blocklist.ts` matches id-first with a normalised-name fallback (`:12`), and the artist fallback compares a single normalised artist string. It therefore never sees a featured credit. What the blocklist needs is the opposite of `leadArtistKey`: **every participant** in a credit, checked against the blocked set.

## The sharp edge

Splitting a credit into participants is not a solved problem, and getting it wrong over-blocks — which is worse here than under-blocking, because the blocklist is absolute (no never-starve anywhere, requests included).

- `feat.` / `ft.` / `featuring` are safe: `FEATURE_SPLIT` already encodes that they are never part of a name.
- `&`, `,` and `x` are **not** safe. Splitting on `&` blocks Simon & Garfunkel, Hall & Oates and Florence + the Machine as collateral. Whether to touch them at all is the design decision this issue is really about — the defensible answer may be "no".
- Per-track artist lists from Subsonic, where available, are a better source than string-splitting a display credit. Check what the library DB actually stores before writing a parser.

## Where the change belongs

The blocklist is deliberately funnelled: rules ride the **existing** chokepoints via `hitOf()` / `isBlocked()`, and adding a second rule-filter at a pick path is called out as the bug. So this is a change to how the artist fallback resolves keys inside `blocklist.ts`, plus whatever participant-key helper is added beside `leadArtistKey` in `recency.ts` — not new filtering anywhere else.

## Acceptance

- Blocking artist X also blocks a track credited "Y feat. X".
- Bands whose names contain a separator are not blocked as collateral; this is covered by tests naming real cases.
- Matching stays synchronous — it sits inside `matchOf()`'s hot path.
- The blocked-tracks admin view attributes such a block to the artist entry that caused it, so an operator can see why a track disappeared.

perminder-klair:
Closes #1603.

## The gap

Blocking an artist blocked the tracks they **lead** and nothing else. A Subsonic song carries one `artistId` — its lead act — and `tracks.artist_id` (migration 23) stores that same single id, so a track credited "Y feat. X" missed the id tier. `matchOf()`'s name fallback compared the whole credit string as one value, so it missed too, in both directions: after blocking X, "Y feat. X" aired; after blocking Y, "Y feat. Z" aired. The operator had to block those track by track.

## What changed

- `music/recency.ts` gains **`artistParticipantKeys`** — every act credited on a track, in credit order — beside `artistRootKey`, which answers the opposite question (who *leads* a credit, for the repeat guard).
- `artistNameKey` is extracted from `artistRootKey`'s own base fold (case, curly-vs-straight apostrophes, whitespace) and exported, because both sides of the blocklist's name comparison now use it. A stored entry name and an incoming credit have to be folded by the same function or a stored value stops matching the value that stored it.
- `blocklist.matchOf()`'s artist tier probes the **whole credit first, then the participants**; `blocklist-rules`' `field: 'artist'` case does the same. Both, because that module's comment already declares its artist case to *be* the fallback's semantics — changing one and not the other would break a stated contract.
- The credit is walked **in order**, so a row several artist entries could claim always resolves to the most specific. That is `matchOf`'s ordering contract, which the admin `blockedBy` badge and its one-click unblock depend on. Attribution therefore names the artist entry that caused the block, as the issue asks.

### Why the whole credit is still probed

An entry's stored `name` is a display **credit**, not an artist's name. `POST /library/blocklist` resolves an artist block from a track row as `{ id: song.artistId, name: song.artist }`, so pressing "Never play this artist" on a row credited `"Host feat. Guest"` persists the composite `"Host feat. Guest"` as the entry name — and blocks the **lead** by id, not the guest the operator may have been aiming at. That wart predates this PR, but it is what makes the whole-credit probe load-bearing: no participant key can ever equal that composite, so matching the acts alone strands every such entry. Rows this station had already blocked would start airing again wherever there is no `artistId` to fall back on — library rows whose `artist_id` predates migration 23 (the walk `COALESCE`s, so a row never re-walked keeps NULL), `ruleMatchRows`, queue items. The rule half carries the same probe because a rule value can be a pasted credit, and without it that rule's `matchCount` quietly reads 0.

Whole credit ∪ participants is exactly old behaviour ∪ new behaviour, so nothing that was playable becomes blocked.

Enforcement is unchanged: no new filter anywhere, the change is entirely inside the keys `hitOf()`/`isBlocked()` already resolve.

## Only `feat.` / `ft.` / `featuring` are split, and that is the answer

The issue names this as the real design decision, and the honest answer is that nothing else can be split safely.

`&`, `+`, `,` and `x` sit **inside** act names far more often than they join two credits: Simon & Garfunkel, Hall & Oates, Florence + the Machine, Earth, Wind & Fire, Tyler, the Creator, Chase & Status. The blocklist is the one list with **no never-starve behind it** — requests included — so a wrong key silently removes music the operator never blocked, with no starvation fallback to make the loss visible. `FEATURE_SPLIT` already carries the reason it is the exception: a featuring marker is always a credit on someone else's track, never part of a name.

`artistRootKey`'s *other* folding (leading-article strip, root aliases, the `&`/`and` join split) is deliberately **not** reused for the same reason — it widens a matching key, which is right for a preference the repeat guard reads as "pick someone else" and wrong for a hard drop.

The cost is under-matching, which is the direction that is recoverable:

- `"Y feat. A & B"` keys as `["y", "a & b"]`, so a block on A alone does not fire. Splitting that tail needs exactly the name-vs-credit judgement the join is refused for, and the tail is where the ambiguous punctuation actually lives.
- A co-credit `"A & B"` stays reachable only through the id tier migration 23 stored it for.

A marker at index 0 is not treated as one ("Ft. Lauderdale …"), the same guard `artistRootKey` applies. The bracket cleanup after a split drops **one orphaned closer**, and only from the segments after the split — a greedy run over every segment ate real names ("Sunn O))) feat. Someone" keyed its lead `"sunn o"` and missed a block the operator did make, and a balanced suffix on the tail lost the bracket it came with).

### It only reaches guests named in the ARTIST field

Navidrome/Subsonic frequently carry the guest in the **title** instead — `"Otis (feat. Otis Redding)"` with artist `"JAY-Z & Kanye West"` — and this change deliberately cannot see that. Reading credits out of titles is a different and much noisier parse ("(Live)", "(Remastered 2011)", "(feat. …)" and a hundred other parentheticals share one field), and it would be guessing in exactly the direction this list cannot afford. The participant ingest noted above is the route to those, not a title parser.

## Behaviour change stated separately: apostrophes

`rebuildIndex` now keys `artistNames` with `artistNameKey` instead of the module-local `norm`, and those differ — `artistNameKey` folds curly apostrophes onto straight ones, `norm` does not. So an entry stored `Guns N' Roses` now matches a credit tagged `Guns N’ Roses`, which it did not before.

That is almost certainly what an operator wants and is the natural consequence of keying both sides with one normaliser, but it is a **widening of an absolute list** with nothing to do with featured artists, so it is called out here rather than arriving silently inside the refactor.

The **album tier is deliberately not folded to match**: `albumKey` still keys with `norm`, so the two name tiers now disagree about apostrophes. Folding it would be a second silent widening in a PR about something else. It carries a comment saying so and is tracked in #1611.

## Operator-facing copy this change falsified

Fixed in the same commit, because the sentence under "Never play this artist" was the only description of this behaviour outside the docs:

- `web/components/admin/library/TrackTable.tsx:394` and `:447` — was *"primary credit only — collabs filed under other artists still play"*, now *"also blocks tracks they're only featured on — acts joined by & or , stay separate"*. Identical wording in both copies so they cannot drift.
- `web/components/admin/library/BlockRulesCard.tsx:56` — the `field: 'artist'` rule hint, adapted for its context.
- `controller/src/schemas/blocklist.ts` — the `normText` doc comment claimed to describe the artist fallback's normalisation, which is no longer true. That file is mirrored verbatim, so the false sentence was already shipping in the web bundle; `npm run gen:schemas` re-run and the mirror committed.

**Expect a conflict with #1604, and it is the wanted kind.** #1604 relocates the `:447` block verbatim into a new `web/components/admin/library/row-actions.tsx` and leaves `:394` in place. Because this PR edits the same hunk #1604 deletes, whichever lands second conflicts instead of auto-merging — which is what forces the new wording into `row-actions.tsx` rather than the old sentence being silently reinstated on the History tab. **Resolution: keep this PR's wording, in whichever file(s) the merged layout has.**

## On the issue's suggestion to use per-track artist lists

Checked, as asked: there is no such list to use. `tracks` carries free-text `artist` plus the single `artist_id`, and OpenSubsonic's per-track `artists`/`participants` are not ingested anywhere in `music/subsonic.ts`. The display credit is the only source that exists today. Ingesting participants (a schema migration plus a full library walk to backfill) is the way to reach the shapes this change refuses to guess at, and is a bigger change than this one; it is noted in `docs/internals/music.md`.

The issue also refers to `leadArtistKey()` at `recency.ts:149` — the function is actually `artistRootKey`, and the line numbers have moved. No behaviour difference, just naming.

## Tests

New `controller/scripts/blocklist-featured-artists.test.ts`: the key helper, the entry path (including attribution and the ordering contract), and the rule path. The collateral cases are named real bands — Simon & Garfunkel, Hall & Oates, Florence + the Machine, Earth, Wind & Fire, Tyler, the Creator, Chase & Status — each asserted to stay playable with its first word blocked.

Two of those tests pin the whole-credit probe specifically — an entry whose stored name is a feature credit still blocks that row, and the same for a rule value — and both fail without it. The protected-name list includes `Sunn O)))` for the bracket cleanup.

One existing assertion in `blocklist-album-id.test.ts` pinned the old behaviour ("the exact-name fallback cannot match a featured credit"). It is now a **co-credit**, which is the shape the name tier genuinely still cannot reach, so the test keeps its original point — the id tier is load-bearing — against a case that is still true.

The new test file's `src` imports are all dynamic, so the `STATE_DIR`-before-`config.js` ordering its comment depends on is defended rather than accidental: a static import that later reaches `config.js` would point the test's `blocklist.json` write at the operator's real state dir. It removes its temp dir on the way out, like its sibling.

`npm run lint` clean in `controller/` and `web/`; full `npm test` passes (1369 tests).


perminder-klair:
Reviewed the diff against `develop`. The split-only-on-`feat.` decision is the right answer and the reasoning holds up — this is the one list with no never-starve behind it, so under-matching is recoverable and over-matching isn't. The named-band test cases are the right way to pin that.

Also confirming two mechanics I checked rather than assumed: `String.split` with a non-global regex still splits on every occurrence, so the `A feat. B featuring C` case genuinely yields three keys; and `base.search(FEATURE_SPLIT) <= 0` is the correct guard for "Ft. Lauderdale" — `search` returns 0 there, not -1.

**One change that should be stated, because it isn't in scope**

`rebuildIndex` changes the `artistNames` key from `norm(e.name)` to `artistNameKey(e.name)`, and those differ: `artistNameKey` folds curly apostrophes to straight, `norm` doesn't. So an entry stored as `Guns N' Roses` now matches a credit tagged `Guns N’ Roses`, which it did not before.

That is almost certainly what an operator wants, and it's a natural consequence of using one normaliser on both sides. But it's a **widening of an absolute blocklist** that has nothing to do with featured artists, and this repo's own rule is that a behaviour change rides its own statement rather than arriving inside a refactor. It needs a line in the PR body and the commit — not a separate PR, but not silent either.

**A second-order effect of the same change**

`albumKeys` still builds its artist half with `norm()`. So after this, the artist tier folds apostrophes and the album tier doesn't — an album entry stored with a curly apostrophe in the artist name still misses a straight-quoted credit. The two name tiers now disagree about what the same string means. Either fold both or say why album is deliberately left alone.

**The admin UI now contradicts the code**

`web/components/admin/library/TrackTable.tsx:394` and `:447` both say, under "Never play this artist":

> primary credit only — collabs filed under other artists still play

This PR makes that false and doesn't touch it. It should — the operator-facing sentence is the only place this behaviour is described outside the docs.

Coordination note: #1604 relocates the `:447` copy into a new `web/components/admin/library/row-actions.tsx` and leaves `:394` in place, so depending on merge order the fix lands in one file or two. Worth syncing with that PR rather than discovering it in a conflict.

**Nit:** the `artistNameHit` comment says "Allocation-free on the ordinary credit". It isn't — `artistParticipantKeys` returns `[base]`, an array allocated per call, and this sits in `matchOf`'s per-track path. The cost is irrelevant in practice; the claim just isn't true, and a false performance note in a hot path is the kind of thing someone later optimises against.

**Good calls worth naming:** walking the credit in order so `matchOf`'s ordering contract survives (the admin badge depends on it); refusing `artistRootKey`'s article-strip and root-aliases because widening a *matching* key is right for a preference and wrong for a hard drop; and rewriting the `blocklist-album-id` assertion onto a co-credit so the test keeps its original point against a case that's still true, instead of deleting it.

Approving once the UI copy is handled and the apostrophe widening is stated.

PR #1610, opened 2026-09-06
feat(shows): takeover option that ends when the schedule would have changed (#1601)

### LINKED ISSUE #1601: Takeover option: end the pin when the schedule would have changed anyway
From the Discord #suggestions triage in #1568 — [thread](https://discord.com/channels/1514647956333002812/1533571088741699667) (Gurthyy, 2026-08-02):

> We have an option to force a takeover for a new show and set how long it will override - can we have an option w/ that that will just set the [takeover to end when the schedule would have moved on anyway]

## What already works

The timed takeover shipped in #930 and was tightened by #1507 and #1543. `ScheduleOverride` is `{ showId, startedAt, expiresAt }` (`controller/src/schemas/schedule.ts:232`), where `expiresAt` is already an **absolute epoch-ms instant**, not a duration. `showId: null` is the explicit Default-programming target; the three-way read is settled by `takeoverShowId()` / `isDefaultTakeover()` and every caller goes through them.

## The gap

The admin takeover dialog only offers fixed durations, so an operator pinning a show mid-slot has to guess how long is left and re-pin (or over-pin and shadow the next scheduled show). What they actually want is "hold this until the grid would have moved on".

## What to build

An option in the takeover dialog — alongside the fixed durations — that resolves `expiresAt` from the **next weekly-grid boundary** instead of `now + N`.

This should need no schema change: `expiresAt` stays an absolute instant, so the resolver, the janitor sweep, the programme span and the roster sweep all keep working untouched. The work is finding the next boundary and feeding it to the existing field.

Things to get right:

- The boundary is on the **station clock**, and grid slots are not necessarily hour-aligned — resolve it through whatever already answers "what show is on at time T", never by rounding to the next hour.
- Decide what happens when the grid has no next change (an empty schedule, or the same show in the following slot). Falling back to a bounded duration is probably right; silently pinning forever is not — `schedule.ts:40` notes the cap exists so "a forgotten pin can't shadow the grid for days".
- The resolved end time should be shown to the operator before they confirm, so the option is not a black box.
- Whether the same option is meaningful for a Default-programming takeover (`showId: null`) is a call to make explicitly rather than by accident.

## Acceptance

- The takeover dialog offers an "until the schedule changes" option and shows the concrete end time it resolved to.
- The stored override is an ordinary `ScheduleOverride` — no new field, no new state, no change to how expiry is judged.
- A schedule with no upcoming change falls back to a bounded window rather than an unbounded pin.

perminder-klair:
Closes #1601.

## What changed

The timed takeover only offered fixed durations, so an operator pinning a show mid-slot had to guess how much of the slot was left and then either re-pin or over-pin and shadow the next scheduled show. The dialog now offers a fourth option — **til change** — that ends the pin at the next weekly-grid boundary.

**Nothing new is stored.** `ScheduleOverride.expiresAt` has always been an absolute instant rather than a duration, so the choice is spent at the route and never persisted: `resolveActiveShow`, the janitor sweep, the programme span and the roster sweep all keep reading an ordinary pin and cannot tell which way the window was picked.

- **Request** (`schemas/schedule.ts`): a new `until` field, default `'fixed'`, so a body written before this is byte-identical. `minutes` is **required** under `'fixed'` and **refused** under `'schedule-change'` — both halves of one rule, that the two fields must not be able to disagree about what was asked for. Demanding a duration the server then ignores is one way to let them; silently discarding one the caller did send is the worse way, since nothing tells the caller its number went nowhere. A fixed window without minutes still 400s with the message it always did.
- **Resolution** (`broadcast/takeover-window.ts`): the boundary is the **grid's**, not the resolver's. `resolveActiveShow` honours a live takeover, so scanning through it would answer "when does the pin I am replacing run out" — the scan hands that same resolver a snapshot with the override taken out rather than restating the day/hour lookup. The scan itself is `show-boundary.ts`'s existing `nextShowChangeMs` with no `extra` candidates, so it walks the station clock minute by minute (slots fire at the hours the operator painted them in, and zones sit at :30/:45).
- **Not a black box.** `GET /schedule/next-change` (admin) serves the resolved end plus the `source` that decided it — `schedule`, `maximum`, or `ceiling` — and the dialog shows the concrete time and the reason before the operator commits; the booth log prints the same reason. It is advisory: the pin resolves the window again at its own `startedAt`. Deliberately not folded into the public `GET /schedule` the dialog already polls — the scan is a minute at a time across twelve hours (≈1.9 ms of `Intl` formatting per call, measured), and that is not a cost to hand every listener-facing schedule read.
- **Default programming** (`showId: null`) gets the option on the same terms as a show pin — "drop the scheduled show and coast until the grid moves on" is the same gesture, and the boundary belongs to the grid rather than to whatever is pinned over it. Called out explicitly rather than falling out by accident.

### 1. There is a ceiling and no floor — and that asymmetry is the rule

The first cut applied `OVERRIDE_MIN_MINUTES` to the resolved instant, as the ceiling's mirror image. That inverted the feature: a boundary five minutes out was held to fifteen, so a pin asked to end when the schedule moves on ran **ten minutes past the change** and shadowed the incoming show — the one harm the option exists to remove, produced by the option itself.

The two bounds are not a matched pair, for the same reason the track-length cap and the track-length floor are held apart in `CLAUDE.md`. **A ceiling trims a window that is otherwise valid**, so applying it costs nothing the operator asked for. **A floor cannot lengthen a window that is genuinely short** — it can only replace the request with a different one.

So: the ceiling stays, unchanged (`OVERRIDE_MAX_MINUTES`; a forgotten pin must not shadow the grid for days). The floor no longer applies to a resolved instant, and still bounds `until: 'fixed'`, where it limits what an operator may type. `scheduleOverrideSchema` imposes no minimum of its own — only `expiresAt > startedAt` and the cap — so a five-minute pin is an ordinary stored override and always was.

**The trade this buys, stated rather than discovered:** the switch lands at the next track boundary, so a window shorter than the track now on air may air the takeover **late**, or **lapse having aired nothing at all**, costing a session roll each way. That is accepted. A pin asked to end at the change ending early is a disappointment; one that runs past the change is the bug.

### 2. A grid with no change in reach still holds at the 12-hour ceiling

`source: 'maximum'` on a uniform or empty grid stays as it is — a bounded pin at `OVERRIDE_MAX_MINUTES` rather than a shorter default. The operator is told before committing ("no schedule change in reach"), and #1601's acceptance criterion is a bounded window, which this is. The Take-over button is not gated on the preview having resolved.

`'maximum'` previously carried two opposite meanings behind one reason string — "the grid never moves on" and "it moves on, later than a takeover can run". The second is now its own `'ceiling'` source. It is unreachable from the live scan (whose horizon *is* the ceiling) and exists as a guard on a caller supplying its own instant.

## Also fixed

- **The dialog checks its failure states before its data.** TanStack Query keeps `data` across a failed refetch *and* a disabled query, so reading it first made the outage copy unreachable after the first success — a dead controller was answered with the last end time the card happened to know, and re-entering the option could repaint a cached absolute instant already in the past. The card now requires the data to be newer than the current selection.
- **`fetchTakeoverWindow` registered in `audit-admin-query.mjs`'s ownership registry.** The new admin read was being reported as outside query ownership; `web/CLAUDE.md` makes that audit a pre-merge gate and it is not wired into CI.
- **`verify-forms.py` compares the previewed minute count with a tolerance of 1.** It is `round((expiresAt - now) / 60_000)` taken per request, so it crosses a rounding half-point once a minute and the browser's fetch and the script's curl straddle it about one run in sixty. An exact match there was a flake, not a stronger assertion.

## Verification

- `controller`: `npm run lint` (eslint + tsc) clean; `npm test` **1376/1376** pass, including `scripts/takeover-window.test.ts`. That file's storability test would have passed with the floor still applied, so it now also asserts `expiresAt <= nextChangeAt`, and there are dedicated cases for a near boundary resolving verbatim (1, 2, 5 and `MIN-1` minutes out, live and pure), for the floor still governing `until: 'fixed'`, and for the `'ceiling'` source.
- `web`: `npm run lint` (eslint + tsc) clean; `npm run gen:schemas` reproduces `web/lib/schemas.generated.ts` byte-for-byte.
- `node web/scripts/audit-admin-query.mjs` reports only the pre-existing `BrainSection.tsx:162` violation; `node --test web/scripts/audit-admin-query.test.mjs` 9/9.
- Driven end to end against an isolated controller + worktree dev server (never the live station): the option resolves a concrete end time against a seeded grid, the stored `expiresAt` matches the preview, the minute box is hidden while it is selected and comes back on a preset, and the pre-#1601 body shape still pins for its 60 minutes. Re-probed after the fix: a boundary five minutes out now resolves to that boundary, `source: 'schedule'`, overshoot 0.


perminder-klair:
Reviewed the diff against `develop`. The core decision is right: `expiresAt` was always an absolute instant, so resolving it from the grid instead of from `now + N` stores an ordinary `ScheduleOverride` and nothing downstream can tell the difference. Excluding the live override from the scan — measuring the grid rather than the pin being replaced — is the subtle part and it's handled correctly.

Confirmed rather than assumed: `nextShowChangeMs` already took `keyAt`/`minuteAt` (`show-boundary.ts:83`), so this is genuine reuse with no change to that module; there is no `/schedule/:param` route for `next-change` to collide with; and a pre-#1601 body still pins for its 60 minutes via the `until` default.

**The one thing worth a decision before merge: the `minimum` clamp inverts the request**

If the next grid change is 5 minutes out, `OVERRIDE_MIN_MINUTES` holds the pin at 15 — so an operator who asked for *"end when the schedule changes"* gets a takeover that runs **ten minutes past** the change, shadowing the incoming show. That is the exact harm the option exists to prevent, produced by the option itself.

The reason string is honest about it, which is much better than silence. But honesty about a surprising outcome isn't the same as the right outcome, and there's a third option neither clamp covers: refuse. When the resolved source would be `minimum`, disabling the choice (with the reason shown) puts the operator in front of the real decision — take a fixed window, or don't pin at all — rather than handing them a window that does the opposite of what the label says.

I'd rather see that than a clamp, but it's your call and the current behaviour is defensible and documented. It should be a deliberate choice rather than a fallout of reusing the request bounds.

**Smaller: `maximum` gives a 12-hour pin from a control labelled "til change"**

A uniform or empty grid resolves to `OVERRIDE_MAX_MINUTES`. The preview does say "no schedule change in reach" before they commit, which is most of the mitigation, so I'm not pushing on this — but falling back to the 60-minute default would match the operator's mental model more closely than the longest window the station allows.

**Comment accuracy, since it's load-bearing**

Two places justify the separate endpoint with "the scan walks a minute at a time across a twelve-hour horizon". Half right, and the half that's wrong is the expensive half: `stationHourBoundaries` does iterate 720 minutes, but only to call `minuteAt`; `keyAt` — the `resolveActiveShow` call — runs on the ~12 instants where the station clock reads `:00`.

The endpoint still earns its place, just for a different reason: 720 `zonedParts` calls means 720 `Intl.DateTimeFormat` reads per request, which is a real cost to keep off a listener-facing route. Worth saying precisely, because that sentence is the justification for a new API surface and someone will check it.

**No action:** the preview can drift from what's stored if the dialog sits across an hour boundary. Already documented as advisory, and the 30s refetch keeps the gap small.

**Good calls worth naming:** making `minutes` optional *only* under `schedule-change` rather than demanding a value the server ignores; resolving after `settings.load()` so the window is measured against the roster the pin is about to sit over; and deciding the Default-programming case explicitly instead of letting it fall out.

This is much larger than the issue implied and the extra surface is justified. Approving on the `minimum` clamp being a stated decision either way.

PR #1614, opened 2026-09-06
fix(broadcast): record what the icecast render trusted, and say so in the dash (#1613)

### LINKED ISSUE #1613: Real listener IPs behind a BYO reverse proxy: no trusted-proxy default, and the miss is invisible in the dash
Carved out of #1568 — [Discord thread](https://discord.com/channels/1514647956333002812/1528123266181632160) (Gurthyy, 2026-07-18):

> I noticed that the listener section of the dash shows the loopback address (127.0.0.*) instead of their real, external IP (123.45.67.*)

## Where this already works

The mechanism is built and correct. `docker/broadcast-entrypoint.sh` resolves a trusted-proxy list — `ICECAST_TRUSTED_PROXY_IPS` (explicit) beating DNS for `ICECAST_TRUSTED_PROXY_HOSTS` (default `caddy`) — and renders it into `<!--@TRUSTED_PROXIES@-->` in `docker/icecast.xml.template`, with the AIO's `render_icecast()` keeping step. On the default `docker-compose.yml` the bundled Caddy resolves and admin → Listeners shows real client addresses.

## Where it doesn't

`docker-compose.byo.yml`. There is no `caddy` service to resolve, so the DNS path always misses, `ICECAST_TRUSTED_PROXY_IPS` is empty unless the operator already knew to set it, and every listener row in `DashPanel` renders the proxy's container address. That is the reported symptom, and every BYO deployment has it by default.

## The actual problem is silence, not the setting

The operator's only signal is one line on stderr in the broadcast container:

```
broadcast: no trusted proxy resolved — listener IPs will show the connecting peer
```

Nothing in the admin UI says why the Listeners table is showing one repeated private address, so the setting that fixes it is undiscoverable from where the symptom is seen. `ICECAST_TRUSTED_PROXY_IPS` is documented in `docs/deployment.md` and `.env.example`, but you have to already suspect the cause to go looking.

## Shape

1. **Have the render record what it decided.** The entrypoint already knows the resolved list and the count; write it as a marker under `STATE_DIR` (shape of the existing `music-starved.json` / `liquidsoap_*.txt` handoffs) — trusted proxy count, the addresses, and which source won. Bootstrap must stay non-fatal per the state-bootstrap rule: a failed write warns, it never stops icecast.
2. **Keep the AIO in lockstep.** `docker/aio/supervisor.sh` `render_icecast()` duplicates this resolution and must write the same marker. This is the fifth thing those two files share; `scripts/state-bootstrap.test.ts` drives both copies from one table and should cover this one the same way.
3. **Surface it where the symptom is.** When the marker says no proxy was trusted, admin → Listeners shows a one-line hint next to the IP column pointing at `ICECAST_TRUSTED_PROXY_IPS` and the deployment doc. Advisory only — no behaviour changes, no gating.
4. **Document the BYO case.** `docs/reverse-proxy.md` covers the header on the proxy side (`proxy_set_header X-Forwarded-For $remote_addr`) and links the deployment section, but does not say that a BYO stack must *also* name its own edge in `ICECAST_TRUSTED_PROXY_IPS` because there is no `caddy` name to resolve. Say it in the nginx and Traefik walkthroughs, where an operator is already standing.

## Constraints worth restating

- icecast-KH matches an **exact IP**. A CIDR is accepted and then silently never matches, so subnet shorthand must keep being rejected loudly rather than interpolated.
- A malformed entry is dropped, never written into the XML — invalid XML here turns a cosmetic setting into a station that will not boot.
- **Listener counting must not start keying on IP.** `groupConnections` deliberately does not: the header can be untrusted, and one NAT is many listeners. Nothing here changes that.
- A miss must keep degrading to the current behaviour (peer address shown), never to a wrong or guessed address.

## Acceptance

- A BYO stack with `ICECAST_TRUSTED_PROXY_IPS` set shows real client IPs, unchanged from today.
- A BYO stack without it shows the proxy address **and** an admin-visible reason, rather than only a container log line.
- The bundled-Caddy default path is byte-identical.
- Entrypoint and AIO write the same marker; the shared test table covers both.

perminder-klair:
Closes #1613.

## The problem

`docker-compose.byo.yml` has no `caddy` service, so the trusted-proxy DNS path misses on **every** boot of **every** BYO install. Icecast's only peer is the operator's own edge, and every row in admin → Listeners renders that container's private address — the reported symptom.

The mechanism was already right: `ICECAST_TRUSTED_PROXY_IPS` fixes it. What was wrong was the **silence**. The operator's only signal was one line on the broadcast container's stderr, three layers from the table showing the symptom, so the setting that fixes it was undiscoverable from where it is seen.

## The change

Both icecast renderers now write `state/trusted-proxies.json` alongside the XML fragment — `{count, source, proxies, dropped, at}` — and `GET /listeners/connections` reads it, so the Listeners card explains itself when the render reported a miss.

```json
{"count":0,"source":"ICECAST_TRUSTED_PROXY_HOSTS","proxies":[],"dropped":["10.0.0.0/8"],"at":1757160000}
```

- **One function per file, not two hand-copied loops.** `render_trusted_proxies()` replaces the inline block in `docker/broadcast-entrypoint.sh` and the AIO's `render_icecast()`; only the caller differs (IPS→DNS vs. loopback). `scripts/trusted-proxies.test.ts` drives both copies from one table, and pins that each file has exactly one call site so the duplicate cannot come back.
- **The marker names the source that was TRIED, not just the count.** "Nothing resolved from `ICECAST_TRUSTED_PROXY_HOSTS`" is a different instruction to the operator than the same miss from `ICECAST_TRUSTED_PROXY_IPS`, and the count alone cannot tell them apart.
- **A CIDR is still dropped rather than interpolated**, but now recorded as well as logged. icecast-KH matches an exact IP — invalid XML here would turn a cosmetic setting into a station that will not boot — and a silently-ignored subnet is the second way an operator sets the var and still sees the proxy address.
- **Writing the marker is never fatal**, per the state-bootstrap rule: an unwritable state dir costs a warning and the trusted proxies are still rendered. It is rewritten on **every** render, so it can never describe a config icecast is not running, and it leaves no temp behind.
- **An absent or self-contradicting marker is UNKNOWN and renders nothing.** A controller ahead of its broadcast image finds no file; degrading to today's behaviour (the peer address, unexplained) is strictly better than a hint that guesses.

## Constraints held

- Bundled-Caddy and AIO paths render byte-identical XML (the AIO's `127.0.0.1 ::1` default and the split stack's `caddy` DNS default are both pinned by tests).
- A miss degrades to the connecting peer, never to a guessed address.
- Malformed entries keep being dropped, never interpolated.
- **Listener counting does not start keying on IP** — `groupConnections` is untouched. The hint is advisory and gates nothing.
- Dropped entries are operator input on their way into JSON, so they are reduced to a safe token first; a quote can't produce a marker the controller cannot parse.

## Docs

`docs/reverse-proxy.md` now says a BYO stack must also name its own edge in `ICECAST_TRUSTED_PROXY_IPS` — in the nginx, Nginx Proxy Manager and Traefik walkthroughs, where an operator is already standing. `docs/internals/infrastructure.md` and the root `CLAUDE.md` record the marker as part of the entrypoint/AIO lockstep.

## Verification

```
cd controller && npm test      # 1449 pass, 0 fail (24 new)
cd controller && npm run lint  # 0 errors
cd web && npm run lint         # 0 errors
node web/scripts/audit-admin-query.mjs  # only the pre-existing BrainSection finding
```


perminder-klair:
Both follow-ups from the smoke boot are in `06f7f2f6`.

**Address validation now checks SHAPE, not characters.** Reproduced first — with `ICECAST_TRUSTED_PROXY_IPS=172.20.0.100,10.0.0.0/8,not a host,fd00::1,cafe` the render wrote `<x-forwarded-for>a</x-forwarded-for>` and counted four trusted proxies. `trusted_proxy_valid()` now requires a dotted quad with four 0-255 octets, or hex-and-colons carrying at least one hex digit for IPv6 (dots allowed for `::ffff:1.2.3.4`). Same input after:

```
<x-forwarded-for>172.20.0.100</x-forwarded-for>
<x-forwarded-for>fd00::1</x-forwarded-for>
{"count":2,...,"proxies":["172.20.0.100","fd00::1"],"dropped":["10.0.0.0/8","not","a","host","cafe"]}
```

It lives in the one shared `render_trusted_proxies`, so both scripts get it by construction — the two copies are byte-identical modulo the indent character. The test table drives 24 address shapes through each copy (`cafe`/`beef`/`ace`/`ff`/`a` drop; `caddy`/`localhost` keep dropping; `0.0.0.0`, `255.255.255.255`, `::1`, `2001:db8::1`, `::ffff:1.2.3.4` keep) plus an end-to-end row for `cafe` asserting the count the dash shows excludes it. `99999999999999999999.1.1.1` is length-checked before the numeric compare so it drops rather than printing an arithmetic error.

Agreed this is pre-existing and inert on air — the reason to fix it here is exactly the one you gave: the count is the new thing, and a hint claiming three trusted proxies when one cannot match is worse than the silence.

**Marker is 644.** Single writer, controller only reads it.

**Word-splitting left alone, commented.** The candidate list has always been space-separated — it is how the DNS path returns several addresses for one name, and how `ICECAST_TRUSTED_PROXY_IPS="1.2.3.4 5.6.7.8"` has always been accepted. Splitting only on commas for the env path would break that and diverge the two sources. Each token shown is genuinely something that was not used, so the display is honest, just granular.

Gates: `controller` 1455 tests pass / 0 fail (30 in this file, +6), lint 0 errors; `web` lint 0 errors.

Issue #1539, opened 2026-09-05
Talk-slot rows can be held past their window by an unbounded pendingTalk clip


perminder-klair:
Follow-up from the review of #1505, which introduced the talk-slot table. Not a regression from that PR — it is a gap the new arbitration layer inherits, and it reintroduces the #1419 shape one level up.

## The gate

`controller/src/broadcast/talk-scheduler.ts:440`

```js
if (row.minGapMs > 0 && p.pendingTalk) {
  return waitPlan(row, slot, slotKey, { held: 'pending', pendingKind: p.pendingTalk.kind }, p);
}
```

A row with a minimum quiet gap steps aside whenever anything is boundary-deferred. That is the right instinct — a rendered segment queued for the next boundary is talk the listener has not heard yet, so `getLastTalkBreakAt()` cannot see it. But the check never asks **when** that held clip will actually air.

## Why the hold is unbounded

`_pendingVoice` can legitimately live up to `PENDING_VOICE_MAX_AGE_MS` (20 minutes, `controller/src/broadcast/queue/kinds.ts:22`), and the staleness drop only runs inside `airPendingVoice` — that is, only at a boundary not already claimed by a track-tied link. `queue.ts` names the conditions itself: a long mix, a stream stall, or a run of link-carrying boundaries.

So the blocker is talk that may air *after* the blocked row's own window has closed.

## Failure scenario

Chatty station, every boundary carrying its own link.

1. An ident fires at :45, defers to a boundary, and is held across several of them.
2. At :00 the session roll runs (correctly — it is outside the table).
3. The hourly time check (`minGapMs` 3 min) sees `pendingTalk`, waits :00–:08, and logs `missed` at :09.
4. The hourly clock gets **one chance an hour**. It previously fired at :00 unconditionally with no gap check at all, so this is a behaviour narrowing, not just a delay.

Banter's :50 window and the next ident slot freeze the same way, for the same reason.

## Severity

Moderate. The ten-minute windows absorb an ordinary 3–5 minute track, and every hold is logged, so it is visible when it happens. But `postpone, never cancel` is the stated contract of the table (`CLAUDE.md`: *"a row that is held postpones, never cancels"*), and an unbounded hold quietly turns a postpone into a cancel.

## Suggested direction

[...]

PR #1624, opened 2026-09-07
feat(broadcast): let the talk planner see the jingle rotate (#1619)

### LINKED ISSUE #1619: feat(broadcast): move the jingle rotate into the controller so the talk planner can see it
> From the Discord thread [Jingles vs The DJ](https://discord.com/channels/1514647956333002812/1532308571126235246) (VYRUS, 2026-07-30), carved out of #1568.

Liquidsoap decides on its own when a rotate jingle plays, every N tracks by `jingle_ratio`. The controller only learns about it after the fact through `jingle-playing.json`, and the collision guard (#997, #1258, #1468) is a hold: spoken segments wait until the jingle's window passes. That stops the station ID talking over the DJ, but the planner cannot plan around a jingle it never scheduled. A link written for the next boundary can land right after a stinger, and a scheduled ident can sit behind a jingle that took the same seam.

**Proposal.** Make the automatic rotate a controller decision.

- The controller counts track boundaries (it already sees every one through `now-playing.json`) and, every N tracks, hands a jingle to the mixer through the existing manual path, `queue.playJingle()` writing `jingle-now.txt`. That path is the one writer of that file today and already handles pending-press bookkeeping.
- The jingle becomes a row in the talk table in `broadcast/talk-scheduler.ts`, so the one-talker-per-minute planner knows a jingle is taking the seam and the segment director stands down for it the way it does for an ident. A jingle is not speech, but it consumes the listener's ear the same way, and the table is where that decision is made for everything else.
- The mixer's own rotate is disabled by writing `jingle_ratio` as 0, which is already the documented way to switch it off. The `jingle-playing.json` guard stays as it is, as the safety net for manual presses and for a controller running behind its broadcast image.

**Upgrade rule.** Absent the new setting, behaviour must be byte-identical. That means the mixer keeps rotating on its own until the operator opts in, or the controller-driven rotate ships as the only mode with the mixer ratio pinned to 0 on the same release. The second is simpler and there is no reason to keep both, but either way an upgraded station must not lose its jingles or get them twice.

**Out of scope.** Advance notice to the DJ script (a link that says "after the station ID") can come later once the planner owns the placement.

Related: #997, #1258, #1468, the talk table rules in `docs/internals/broadcast.md`.

perminder-klair:
Closes #1619.

## The problem

Liquidsoap decided on its own when a rotate jingle played — `rotate(weights=[1, jingle_ratio()])`, one stinger every N tracks, on the raw pre-cross music source. The controller only learned about it afterwards, through `jingle-playing.json`, and the collision guard built on that marker (#997, #1258, #1468) is a **hold**: spoken segments wait for the jingle's window to pass.

That stops the station ID talking over the DJ, but the planner cannot plan around a jingle it never scheduled. A link written for the next boundary lands right behind a stinger, and a scheduled ident sits behind a jingle that took the same seam.

## The fix

With `settings.jingleRotate: 'controller'` the count moves into the controller, which already sees every track boundary through `now-playing.json`. The stinger is handed over through the **existing single writer**: `queue.playRotateJingle()` draws the clip, `queue.playJingle()` writes `jingle-now.txt`. `settings/liquidsoap.ts` writes the mixer's `jingle_ratio` 0 — already #997's documented "jingles off" value — so exactly one side is ever counting, and **`radio.liq` is untouched by this PR**.

New pure policy module `controller/src/broadcast/jingle-rotate.ts` owns the four decisions (who counts, the exact bytes of the ratio file, whether a rotate is due, which clip to draw) plus the `/debug` snapshot. The mixer handoff writer and the talk tick resolve through the same functions, so "the mixer is rotating" and "the controller is rotating" cannot disagree — they are two readings of one value.

### The row is a `fill` row, not a slot

This is the same argument the segment director's role is, rather than a weaker version of it. A fill row is one with *no scheduled chance to lose*, and that is exactly what a rotate is: its due-ness is a COUNT (`queue.rotateJingleTracksSince()`) that keeps counting while the row waits, so a minute given away costs a minute and nothing else. A slot row here would be wrong twice — it would suppress the director on every minute the jingle merely waited, and every hold would log `missed` about a chance that was never lost, since an `opens: 'any'` row's window is one minute wide and `canRetry` is therefore false on all of them.

What the issue asked for ("the segment director stands down for it the way it does for an ident") is what one talker per minute already does between two fill rows: the higher-priority plan takes the minute and the other waits. Priority is **last**, by the table's own principle — among rows that can retry, fewer remaining chances outranks more, and the director is offered twelve minutes an hour against this row's sixty.

Three more distinctions, each deliberate:

- **Gap** — three minutes, the short segments' figure. That is the half of #310 this row can answer (a stinger must not land on the back of a break that just finished). The other half — talk landing on the back of the *stinger* — is not the planner's and stays in the `jingle-playing.json` hold and `radio.liq`'s `voice_until` gate, which is why that guard survives untouched. The gap is one-directional for a second reason too: a jingle is not talk, so it never appears in `getLastTalkBreakAt()`.
- **Air** — `'immediate'`, because that describes the HANDOFF. The write happens now and Liquidsoap's track-sensitive `jingle_now_queue` places the clip at the next safe boundary; claiming `'next-track'` would say the controller is deferring something it is not.
- **Pending-clip hold** — unbounded for this row, where #1539 bounds it for a row with a finite window. That bound exists only to stop a hold turning a postpone into a cancel, there is nothing here to cancel, and the bounded form would read as "never held at all". The generalisation in `pendingHolds` is guarded on `minGapMs > 0`, so the segment director (which has opted out of quiet questions entirely) is unmoved.

Eligibility carries **none** of the DJ gates: a stinger costs no tokens and no TTS, and neither the mixer's rotate nor an operator press has ever been listener-gated or muted by `tts.enabled` (`voice-policy.ts` names `jingleRatio: 0` as the way to silence jingles precisely because the voice switch does not). `runTalkSlot` also skips its heavy-TTS warm for this row — the clip is already on disk, and warming for it would reload a model nothing is about to use.

A rotate that comes due and cannot be drawn (empty library, the same clip still pending) **spends** the offer rather than banking it. That is what `radio.liq`'s `source.available` gate did — "skipping a jingle is the cheaper miss", in that file's own words — and banking it would leave the row due every subsequent minute, holding the seam against the director until the library was filled or a press aged out.

## Upgrade / behaviour note

**Opt-in, default `'mixer'` — absent the key nothing is written differently and nothing new fires.**

The issue offered the alternative of shipping the controller rotate as the only mode with the mixer ratio pinned to 0 on the same release, and called it simpler. It cannot satisfy its own rule (*an upgraded station must not lose its jingles or get them twice*), because the controller and broadcast images upgrade independently and `liquidsoap_jingle_ratio.txt` is **read once at mixer startup**:

- Pinning the file to 0 for everyone leaves the mixer rotating on the operator's old ratio until its next start, which on a `up -d --build` races the controller's own boot write. Every station that lost that race would hear its jingles twice, by default, on the release.
- Pinning it inside `radio.liq` instead loses every jingle on a station whose *controller* is still the old image — the other skew direction.

Opt-in has neither moment, and because `radio.liq` is unchanged both skews stay correct: a new controller switches an old mixer's rotate off through a file that mixer already honours, and an old controller drives a new broadcast image exactly as today.

The residual cost is the read-once lifecycle itself — a station that flips the toggle and does not restart the mixer runs both rotates until it does. That is the existing contract for this exact file (the jingle-ratio control has always carried "restart required" / "Save · needs restart"), it is paid on an operator action rather than involuntarily on an upgrade, and the new toggle's copy says it in as many words.

The operator's `jingleRatio` is **not** zeroed in `settings.json` — it stays the "1 jingle every N tracks" figure, which is exactly what the controller's counter now reads. `jingleRatio: 0` still means jingles off whichever side is counting. The mixer branch of the ratio-file writer is byte-identical to the `String(s.jingleRatio)` it replaces, including for values `load()` would never produce: coercing a malformed ratio to a number would write `0` and silence the jingles, where the unparseable string `radio.liq` has always been handed makes it keep its own default.

## Surfaces

- Admin → Imaging → Jingles: a **counted by · Mixer / Controller** segmented control in the existing "how often" panel, with the restart warning stated on it.
- `/debug` gains a `jingleRotate` row (`owner`, `ratio`, the verbatim `mixerRatioFile`, and `tracksSinceJingle` while the controller owns it) beside `voice` / `clock` / `talkAir`. The one state it cannot see is a mixer that has not restarted since the switch — which is why the control says "needs restart".
- Doctor's broadcast summary line names the owner.

## Verified

- `npm test` in `controller/` — **1488/1488 pass**, including a new `scripts/jingle-rotate.test.ts` (16 tests): the cold-load round trip both ways (the three-edits rule `controller/CLAUDE.md` documents), the byte-identical mixer branch including malformed values, the restart flag and its no-op re-save, the count and its overshoot, the draw's no-repeat, and the row's arbitration — director yields, slot row wins the seam, quiet gap, unbounded pending hold, `djTalkOnlyBetweenTracks`.
- `npm run lint` in `controller/` and `web/` — 0 errors.
- `npm run gen:schemas` re-run; the mirror is in sync (CI diffs it).
- Four harness updates in the existing `talk-scheduler.test.ts`, each commented inline. Chiefly: its default `eligible` now mirrors a station that has not opted in, and a structural loop over `row.opens` skipped `'external'` but not `'any'` (it was iterating the characters of the string, which only became visible with a second `'any'` row).
- Docs updated where the rules live: root `CLAUDE.md` (IPC table, one-writer bullet, talk-table bullet), `controller/CLAUDE.md`, `liquidsoap/CLAUDE.md` (why that file must keep honouring the 0), `docs/internals/broadcast.md` (full rationale) and `docs/internals/schemas.md` (fourth key added to the patch registry rather than converted into it).

## Deliberately left out

- **Advance notice to the DJ script** ("a link that says *after the station ID*") — the issue scopes it out, and it can come now that the planner owns the placement.
- **A detector for "has the mixer picked up the 0 yet"** — there is no signal that distinguishes an unrestarted mixer from a restarted one (`jingle-playing.json` carries the same shape from both writers), so the honest answer is the restart affordance that control already has.
- **No `radio.liq` change at all**, on purpose — see the upgrade note.
- The docker stack was not started, so nothing here is verified on air.


perminder-klair:
Tested this on air from a worktree running the branch (dev stack, `jingleRatio` lowered to 2 so rotates are observable, `maxTrackSeconds` 120).

**Default path is unchanged.** With `jingleRotate` absent, the ratio file keeps its real value and the mixer rotates on its own, one jingle per two tracks across four boundaries. `git diff develop..HEAD -- liquidsoap/radio.liq` is empty, so the claim that radio.liq is untouched holds.

**Opt-in works end to end.** `POST /settings {"jingleRotate":"controller"}` validated through the patch registry and flipped `liquidsoap_jingle_ratio.txt` from 2 to 0 immediately. After a mixer restart, with the mixer no longer counting:

```
10:22:53  TRACK   Breathe
10:23:00  jingle-now.txt  annotate:subwave_kind="jingle":/var/sub-wave/jingles/station_ident_default.wav
10:24:35  TRACK   Motivational Techno
10:24:37  jingle-playing.json  station_ident_default.wav
10:26:37  TRACK   Ground Cayenne
10:28:19  TRACK   SaReGaMa — Oh No, Babe
10:28:19  jingle-playing.json  station_ident_default.wav
```

One jingle every two tracks, queued by the controller and aired at the next safe boundary. No doubles, no lost stingers.

**The read-once premise is real, and I hit it by accident.** `broadcast` starts before `controller` in compose, so the mixer booted on a stale ratio and ignored the controller's later write until it was restarted. That is exactly the race the commit message cites for rejecting "pin the ratio to 0 for everyone", so the opt-in choice looks right for the reason stated.

**Two things I could not confirm.**

The skew cost may be overstated. Running with `jingleRotate: 'controller'` and the mixer still holding its old ratio, I saw 2 jingles against 2 controller fires over three tracks — 1:1, not the doubling the commit message predicts. radio.liq shares its live safety gates between on-demand and rotate jingles, which would explain the mixer standing down while a pressed jingle feeds, but three tracks is too thin to call it a property. Worth a look, since it would make the residual cost of forgetting the restart smaller than documented.

Jingle variety went untested: the state dir I tested against catalogues a single ident, so `pickRotateJingle`'s don't-repeat-the-last-one path never ran. Rig limitation, not a gap in the change.

**Two notes on the test edits.**

`talk-scheduler.test.ts`, the #310 loop: skipping `opens: 'any'` is a fix, not a loosening. `'any'` already existed on develop for the `segment` row, so the old `for (const m of row.opens)` was iterating the characters `'a'`, `'n'`, `'y'` and registering them as minutes — vacuous for that row, harmless only because they never collide with integers.

The #1539 relaxation is the one place a guard on pre-existing behaviour got wider: exempting `opens: 'any'` from "the hold never takes the row's last chance" covers the `segment` row too, not just the new `jingle` row. The reasoning is sound, but it is a change to what that assertion pins for a row that already existed.

Nit: at `:12` the assertion moved from `[]` to `['jingle']` but kept its old message, `'a closed window must not reach a policy module'`, which now reads as the opposite of what it asserts.

perminder-klair:
Holding this for a human read rather than for anything I found broken — the on-air testing above came out clean, and the feature does what it says.

**Two things I can't settle myself.**

The skew cost may not be real. The commit message argues for opt-in over "only mode" partly because flipping the toggle without restarting the mixer leaves both sides counting and gives you jingles twice. When I actually ran that state — `jingleRotate: 'controller'` with the mixer still holding its old ratio — I got 2 jingles against 2 controller fires over three tracks. 1:1, not 2:1. radio.liq shares its live safety gates between on-demand and rotate jingles, which would explain the mixer standing down while a pressed jingle feeds, but three tracks is nowhere near enough to call it a property and the two rotates may simply not have collided yet. If the doubling doesn't happen, the reasoning in the commit message is describing a cost that isn't there, and that reasoning is load-bearing for the design choice. Worth either confirming or correcting rather than shipping as written.

The #1539 relaxation widens a guard on pre-existing behaviour. Exempting `opens: 'any'` from "the hold never takes the row's last chance" is sound for the new `jingle` row, but `segment` was already an `'any'` row on develop, so that assertion now pins less for a row that predates this change. Deliberate and defensible, but it's the one edit here that loosens something that was already guarding something else.

(The other #310 loop change is the opposite — skipping `'any'` is a fix. The old `for (const m of row.opens)` was iterating the characters `'a'`, `'n'`, `'y'` for the segment row and registering them as minutes, so that assertion was vacuous there and only harmless by luck.)

**Also needs a refresh before merge.** #1627, #1623 and #1625 have landed on develop and this branch is three behind. It has the heaviest overlap of the six — `controller/src/schemas/settings.ts`, `settings.ts`, `settings/defaults.ts`, `web/components/admin/settings/shared.tsx`, both CLAUDE.md files and `web/lib/schemas.generated.ts`, all touched by #1623 as well. Merge develop in and re-run `npm run gen:schemas` before merging: the mirror is a flat concatenation and CI diffs it, so a textual auto-merge of that file can pass git and still fail the drift check.

perminder-klair:
Pushed `25032be` — review fixes, plus the `develop` merge that was blocking. Nothing here changes the design; the fill-row argument, the opt-in shape and the untouched `radio.liq` all stand.

**Merge.** `develop` merged in (`7bca366`), so the three commits that had landed are on the branch. The overlap with #1623 auto-merged without conflicts and, on the point that mattered, `npm run gen:schemas` produces byte-identical output to the merged mirror — so the drift check was never going to fire. Worth knowing the risk didn't materialise rather than assuming it couldn't.

**`/debug` was reporting a figure it computed.** `mixerRatioFile` was a pure function of settings, labelled "verbatim" — it agreed with settings by construction and could never disagree with the mixer, which is the one thing the row exists to show. It now reports `mixerRatioIntended` beside `mixerRatioOnDisk` (the actual bytes of `liquidsoap_jingle_ratio.txt`) and `mixerRatioMatches`. The state this makes visible: `ensureLiquidsoapSettingsFile()` only writes a handoff file that is *missing*, so hand-editing `jingleRotate: 'controller'` into `settings.json` and restarting the controller leaves the mixer booting on the old ratio, both sides counting, indefinitely. The read is at the call site so `jingle-rotate.ts` stays I/O-free, and an unreadable file reports `null` rather than a guess — "couldn't check" and "checked and it disagrees" are different instructions.

**The boundary count now survives a restart.** It was in-memory only. Unlike `_trackStarts`, which is only ever read as a difference against a stamp taken in the same process, this one is absolute — every `--build controller` cost up to a full `jingleRatio` of tracks, roughly two hours at the default 30. It rides `persist()` now and is repaired on recovery.

**The rotate no longer spends the operator's press budget.** It was going through `playJingle` and so drawing on `PENDING_JINGLE_MAX`. Three stuck rotates inside the 30-minute TTL would have answered `queue-full` to a press — the wedged button `PENDING_JINGLE_TTL_MS` exists to prevent, reintroduced from the other side. Separate budgets rather than a reserved slot, since reserving still shrinks the operator's headroom for a caller that isn't a runaway risk. The rotate's own cap is **one**: it's one-at-a-time by construction (the counter zeroes at handoff), so a second pending rotate can only mean the first never aired, and stacking on it is what the FIFO has no remove path to undo. Still one map, so de-duplication stays a single question asked across both callers.

**Flipping the toggle starts a clean cycle.** The counter runs on every boundary regardless of owner — `onTrackStarted` has no business branching on a setting — so a station up for hours was already well past the ratio and fired a stinger on the first tick after the switch, on top of a mixer that hadn't restarted. Zeroed on the `mixer → controller` transition via an owner subscriber shaped exactly like `time.ts`'s timezone listeners, and for the same two reasons: `update()` isn't the only writer (a backup restore reaches it directly), and `settings.ts` can't import the queue without closing a cycle.

**Two documentation gaps.** `liquidsoap/CLAUDE.md` now records that the controller-owned path drops **two** of the rotate's four gates. `not jingle_now_on_air()` is moot. `not music_starved()` is not — it exists because of a measured failure (#1300 bug 7). It doesn't come back, and the reason is structural rather than lucky: the count is fed by `now-playing.json`, written from `music_meta` — the pre-rotate handle, the same chain the starve guard samples — so a starved chain stops the counter and at most one already-due stinger airs into a starve. That argument is now written down, along with the condition that would kill it. Also noted in `pendingHolds` that for the jingle row the hold is courtesy, not `_pendingVoice` protection, since that row never writes the voice slot.

**On the skew magnitude — still unmeasured over a long run, and stated as such.** The "hear its jingles twice" claim is softened to "may hear jingles more often than configured" in both `docs/internals/broadcast.md` and the module header, with the 1:1-over-three-tracks observation and its likely explanation recorded: the mixer's own rotate carries `not jingle_now_on_air()` and `time() > voice_until()`, so it stands down while a controller-handed stinger is feeding. Three tracks settles nothing and I have not run a longer check — the number is genuinely open. It doesn't move the conclusion, since the objection to the only-mode shape is that its skew is *involuntary* and arrives on an upgrade, not that it's exactly double.

**Smaller items.** `doctor/types.ts` and the web settings form take `JingleRotateOwner` from the schema instead of respelling the union. The #1539 test exemption narrows from `opens === 'any'` to `opens === 'any' && minGapMs > 0`, so it stops covering the pre-existing `segment` row — inert today (its `minGapMs: 0` returns before the branch), but the assertion is the guard and shouldn't be the half that goes slack first. The `:12` assertion message now says what it asserts. And `jingle-play.test.ts` covers the queue half that had no tests: the handoff, the boundary count and its dedupe, the restart round trip, the new budget separation, the transition reset, and the rule the design leans on hardest — a fired rotate that cannot draw **spends** the offer.

**Left alone:** the Jingles panel's on-click ownership toggle. It's defensible as argued — a segmented control has no half-typed state, and `saveSettings` already raises "saved, restart the mixer to apply" on top of the static copy.

**Gate, run against the merged base:**

- `controller` — `npm run lint`: 0 errors (692 pre-existing `no-explicit-any` warnings). `npm test`: **1527/1527 pass**, 0 fail (was 1488; +39 new).
- `npm run gen:schemas`: mirror unchanged, `git status --porcelain` clean afterwards.
- `web` — `npm run lint`: 0 errors (5 pre-existing warnings).

Not verified on air: the docker stack wasn't started for these changes.

PR #1625, opened 2026-09-07
feat(skills): make feed: work on any skill, not just news (#1616)

### LINKED ISSUE #1616: feed:/feedMaxItems on a custom skill is never injected into the prompt — the fetch tool only exists for the built-in news skill
### Environment

Reproduced on **both 1.10.0 and 1.13.0** (same host, sequential test - pulled/recreated the controller image between runs, nothing else changed), OpenRouter (`qwen/qwen3.7-plus-20260602`), a custom skill (`giveaway`) declaring `feed:`/`feedMaxItems` in its SKILL.md front-matter, identically to how the built-in `news` skill declares its own.

### Summary

A custom skill's `feed:` front-matter field does nothing. The skill's own instruction text can say "the feed above is the actual, current state" - but no feed content is ever placed in the prompt, "above" or anywhere else. Verified two independent ways, on two different controller versions, with a smoking-gun reproduction below.
The field is accepted by schema/UI/readback paths but has no runtime effect for custom skills.

### Reproduction

1. Stood up a small external state service exposing an RSS `feed:` endpoint (a generic contest/state sidecar - irrelevant to the bug, any URL returning valid RSS behaves the same).
2. Added one contestant with a deliberately unmistakable name/entry, confirmed present in the endpoint's own output at the moment of the test:
   ```
   ACTIVE CONTESTANTS: Zorblatt Q. Fennelwhistle (a submarine)
   ```
3. Manually fired the skill (`POST /dj/skill {"name":"giveaway"}`).
4. The model's spoken output made **zero reference** to Zorblatt, the submarine, or the dish - it continued an unrelated storyline it already had from the on-air recap.
5. Repeated after pulling and recreating the controller on **1.13.0** - identical result.

### The smoking gun (1.13.0, via the new `/debug/llm-calls/export?format=ndjson` route)

The skill's own SKILL.md instructs the model as if a feed block precedes this text:

```
system: "This is the WitWisdom Signal Weekly Cook-Off. The feed above is the
actual, current state of the contest - treat it as ground truth, not a
suggestion. If it lists active contestants: report on their ongoing
progress instead of picking someone new. ..."
```

The **actual exported prompt** for that exact call contains no feed block anywhere - not before that instruction, not after it, not truncated, not empty-with-a-marker. Just absent. The full `messages` array for the call is:

```
user: "The current moment:\nDay: Sunday, 6 September ...\nOn now: the show
\"...\" - ...\nOn air now: \"Slow Fall\" by The Plan\n\nWhat you have already
said on air recently (do NOT repeat these topics or phrasing):\n- 1m ago
[link]: \"...\""
```

That's the entire user turn: wall-clock/weather/show context, plus the recap of recently-aired lines. No feed content, no `ACTIVE CONTESTANTS`, no `Zorblatt` - nothing resembling what the skill's own fetched data should have contributed. The `tools` array for this call also carries nothing resembling a `skill_giveaway` fetch tool - matching what grepping raw request logs showed on 1.10.0 and (independently, months earlier) on 1.11.0: `skill_news` appears once per fire, `skill_<customName>` never appears at all, on any custom skill that declares `feed:`.

### The mechanism (unchanged from earlier investigation, now confirmed on newer versions too)

`feed:`/`feedMaxItems` is not a generic front-matter-driven mechanism, despite the field name and despite `news` visibly using it. It appears hardcoded to the `news` skill specifically - the same way `web-search`/`library-deep-cut`/`now-playing-dig` each have their own dedicated built-in tool - rather than something any skill gains by declaring the field. Nothing in the SKILL.md schema, the admin UI, or the logs indicates this: the field validates, saves, and is echoed back by every read path, and then simply has no effect.

### What I am not claiming

I'm not claiming this is a version regression - it reproduces identically across three separate controller versions (1.10.0, 1.11.0, 1.13.0) I've had occasion to test against, so it looks like it has always been `news`-specific. I'm also not claiming there's no other way to give a custom skill access to external state - only that `feed:`/`feedMaxItems`, despite being general-purpose in name and schema, is not it.

### Thoughts, not prescriptions

Smallest fix: generalize what `news` already does - any skill declaring `feed:` gets an auto-generated `skill_<name>` fetch tool over that feed, with the same dedupe/truncation behavior `feedMaxItems` already implies. That makes the field do what its name says.

Cheaper if the current scope is intentional: fail loudly. A skill declaring `feed:` that isn't wired up should log a warning (or refuse to load) rather than silently behave as if the field were absent - the failure mode today is invisible short of diffing exported LLM prompts against the SKILL.md's own claims, which is a heavy way to discover a config field does nothing.

Either would matter for any skill that needs to track state with continuity across many separate LLM calls (a running contest, a scoreboard, anything an external service is the source of truth for) without hand-rolling per-skill wiring upstream each time.

Happy to share the full exported call (`/debug/llm-calls/export`) behind the smoking-gun section above.

perminder-klair:
Closes #1616.

## Problem

A custom skill could declare `feed:` / `feedMaxItems:` in its `SKILL.md`, watch them validate, save, and read back through every route and the admin UI — and still reach the model as nothing at all.

Confirmed in the code before changing anything: `skills/loader.ts` only ever set `cap.toolFn` / `cap.toolName` inside `if (toolMod?.fn)`. A skill got a `skill_<name>` tool **only** if it shipped a sibling `tool.mjs`. `feed:` / `feedMaxItems:` were parsed into `cap.config` and handed to that tool as its 4th argument — so they were config for exactly one hand-written file, `builtins/news/tool.mjs`. A skill with the same frontmatter and no `tool.mjs` produced no tool, made no request, and put no feed content in the prompt. The only way to notice was to diff an exported LLM call against the SKILL.md's own claims, which is how the reporter found it.

#1550 had already documented this as intended, including a manual callout ("FRONTMATTER IS CONFIG, NOT A FETCH") and a test pinning it. This PR takes the other option the issue offered: make the field do what its name says.

## The fix

The declaration is the mechanism now.

- **`skills/news.ts` → `skills/feed.ts`.** Nothing in that module was ever news-specific — `parseFeed`, `fetchHeadlines`, `hashHeadline` are the generic feed layer. The filename was a large part of why the mechanism read as news-only. It now also owns the declaration policy: `resolveFeedConfig` (frontmatter → a usable feed), `makeFeedTool` (fetch → burn-on-read dedupe → truncation), plus `FEED_CONFIG_FIELDS` and `FEED_ITEMS_PER_FIRE`.
- **`skills/loader.ts`** attaches that generated tool in the `else` of the `tool.mjs` branch — any skill with a usable `feed:` and no tool of its own. It reuses the existing `parseConfigFields` sanitiser (one declaration format, not a second one) and a shared `toolNameFor()` so both paths spell the tool name once.
- **`builtins/news/tool.mjs` is deleted.** News takes the generic path like everything else rather than sitting beside it as a second copy of the same fetch — the arrangement that produced the bug.

Three asymmetries are deliberate, and are written down in `controller/CLAUDE.md`:

- **The URL is strict, the count is lenient.** A non-`http(s)` `feed:` logs a warning naming the skill and yields no tool — a `feed:` line must never quietly do nothing, which is the whole complaint. A broken `feedMaxItems` has an obvious right answer, so it falls back to the station default and says so.
- **The feed knobs are offered to every tool-less skill**, whether or not a value is set. The edit sheet is where an operator types the first feed; a field that appears only once a value exists is a form nobody can use to create one.
- **Dedup memory is keyed per skill** (`segmentState.feedSeen`), so two skills reading two feeds can't suppress each other's items on a hash collision.

## Upgrade / behaviour note

**An upgrade is byte-identical for news.** The per-fire cap (6 fresh items), the read cap (`feedMaxItems`, default `NEWS_MAX_ITEMS`) and the seen-memory trim (120 → 60) are the numbers news' `tool.mjs` used, kept for every feed skill. The returned shape is unchanged (`{ headlines: [{ title, detail }] }`).

**Existing stations are untouched at runtime.** `state/skills` is seeded once and never re-seeded, so an install booted before this keeps its own `state/skills/news/tool.mjs` and keeps running it — same behaviour, its own copy. **Reset to default** on the News skill is the migration: `resetBuiltinSkill` now removes a `tool.mjs` that the shipped template no longer carries, instead of restoring everything except the one stale file.

The parser has handled Atom and RDF/RSS-1.0 since the `fast-xml-parser` change; the docs still said RSS 2.0 only, and that is corrected here.

## Verified

- `controller`: `npm run lint` — 0 errors. `npm test` — **1483/1483 pass**.
- `web`: `npm run lint` — 0 errors.
- `npm run gen:schemas` produces no diff (no `src/schemas/` changes).
- New `controller/scripts/skill-feed-tool.test.ts` (11 tests) drives fetch/parse/dedupe end to end against a loopback HTTP server rather than a stub: tool generation from a `feed:` line alone, the per-fire cap and the second fire returning the remainder, per-skill dedup isolation, `feedMaxItems` truncation, Atom, a `tool.mjs` winning over the generic path, the loud warn on an unusable URL, stand-down on a failing fetch, `resolveFeedConfig`'s strict/lenient split, and a guard that the news template ships no `tool.mjs`.
- `skill-tool-loading.test.ts`: its "a `feed:` line without `tool.mjs` stays prompt-only" case *was* the reported bug, so it now pins the boundary that still holds — no feed **and** no `tool.mjs` means no tool.
- `scripts/news-feed.test.ts` → `scripts/feed-parse.test.ts`, following the module.

Docs updated: `docs/custom-skills.md` and the in-product manual both gained a "Feeds without code" section and lost the now-wrong callout and the hand-rolled RSS example; `controller/CLAUDE.md` records the mechanism and the asymmetries.

## Deliberately left out

- **An empty fetch still returns `{ headlines: [] }`, not `{ available: false }`.** So a feed skill whose items have all been seen speaks from its brief rather than standing down. That is news' existing behaviour and tightening it would change what the station does on air — its own PR, stated as such, if it's wanted.
- **No docker stack was started**, so nothing is verified on air. The fetch path is covered by the loopback-server test instead.
- Existing installs are not rewritten on disk. Silently deleting an operator-editable `tool.mjs` at boot is worse than leaving a working copy in place; Reset to default is the opt-in route.


perminder-klair:
Reproduced the issue's own scenario against a running dev stack on this branch: a custom skill `giveaway` declaring `feed:` and `feedMaxItems: 5`, pointed at a small RSS endpoint carrying the same unmistakable marker the report used.

Fired it the same way the issue did, `POST /dj/skill {"name":"giveaway"}`:

```json
{"ok":true,"name":"giveaway","aired":true,
 "spoken":"WitWisdom Signal Weekly Cook-Off update: Zorblatt Q. Fennelwhistle,
  currently boiling a submarine in round three, leads with 91 points—forty ahead
  of second place. The afternoon has deteriorated exactly as scheduled.",
 "reason":"The feed lists an active contestant and current standings for the Cook-Off."}
```

Every fact in that line came from the feed. The report's whole complaint was that the model made zero reference to Zorblatt because no feed content ever reached the prompt.

The exported call via `/debug/llm-calls/export?format=ndjson` — the same route the reporter used — shows the mechanism rather than just the outcome:

```json
"toolCalls": [{"name": "skill_giveaway", "args": {},
  "result": {"headlines": [
    {"title": "ACTIVE CONTESTANTS: Zorblatt Q. Fennelwhistle (a submarine)",
     "detail": "Zorblatt Q. Fennelwhistle is currently ..."}]}}]
```

`skill_giveaway` is generated and called, and its result carries the feed items. That is precisely the tool the issue said "never appears at all, on any custom skill that declares `feed:`".

Also confirmed the skill loads as custom rather than being special-cased: `[skills] 1 custom skill(s): giveaway`, alongside the 7 built-ins.

One thing I did not do: re-verify the negative on `develop`. The issue documents it across 1.10.0, 1.11.0 and 1.13.0, so I took that as given rather than rebuilding a second stack to watch it fail.

Issue #1105, opened 2026-07-19
Raise the 1000-character limit on persona souls


jordykoppen:
The soul field on personas is capped at 1000 characters — SOUL_MAX in web/components/admin/personas/constants.ts, with the same 1-1000 check in the controller's settings validation, so it can't be worked around via the API either.

For a basic persona that's enough, but for a more complex character it runs out fast. Once you add a backstory, speech quirks, recurring bits, and a few do's and don'ts for how they talk about music, you're way past 1000 and end up trimming exactly the details that make the persona distinctive rather than a generic DJ voice.

Could the limit be raised to something like 4000-5000, or made configurable? The soul lands in the DJ system prompt, and current models handle prompts that size fine — token cost seems like the only trade-off, which could be left to the operator.

If the limit is tight for another reason (prompt budget across the roster?), curious to hear it — but from the operator side, more room here would go a long way for building richer personas.

perminder-klair:
Your read is right, token cost is the only reason it was bounded. Nothing structural depends on the number.

Gone with 2000 rather than 4000 in #1113. That's around 300 words, so backstory plus quirks plus do's and don'ts may still be tight for what you're describing. It's two numbers to move now, so try it and tell me if it's still cramped.

Your issue also surfaced something I'd missed: the community catalog parser had the same cap, and it dropped an over-cap persona from the listing rather than erroring. So the catalog couldn't ship richer characters either.

Issue #722, opened 2026-07-01
Expand show topic/direction field beyond 400 chars for acoustic-aware DJ prompting


Wayne-Harbart:
Now that acoustic analysis (BPM, key, intro_ms, time signature feel, pace curve) is in the mix, the show topic field is the natural place to give the DJ real musical direction — but 400 characters isn't enough to express it meaningfully.

**What I want to write:**

> Odd time signatures only — 5/4, 7/8, 11/8, 13/8. Use BPM + pace_curve to find tracks that feel metrically unstable even when tagged ambiguously. Blend across math rock, prog, jazz fusion and contemporary classical where the pulse is slippery. Prioritize tracks with long intros and modulating key_ranges. DJ should name the time signature when introducing a track and explain why it sits oddly.

That's already over 400 chars and it's barely scratching the surface of what acoustic data now makes possible.

**Ask:**
- Lift or remove the backend cap on `topic` (a textarea with no hard DB limit seems fine)
- If there's a UI character counter, update it to reflect the new limit
- Consider a dedicated "acoustic direction" sub-field on shows for BPM range, key filter, intro_ms hints — separate from the narrative topic

The acoustic analysis data is genuinely exciting. The bottleneck is now vocabulary, not data.

perminder-klair:
Thanks for the detailed writeup, the acoustic angle is a good push.

One thing worth clearing up first: the show `topic` field is already capped at 1000, not 400 (it went 500 to 1000 a while back), so your example actually fits with room to spare there.

The real 400 cap is on the persona **soul** field, which is the DJ's voice and personality sketch. That's the one you were running into. I've raised it to 1000 in #728, matching the topic cap, so there's a lot more room to write real direction now. I kept it bounded rather than unlimited because soul gets fed into every spoken segment the DJ generates, so length carries a running cost.

The dedicated acoustic direction sub-field is a nice idea. It's a bigger piece of work, so I'll track it separately rather than hold this change up.

PR #1626, opened 2026-09-07
feat(web): expand a clamped show topic, and open the DJ behind the name (#1621)

### LINKED ISSUE #1621: feat(web): expand the truncated show topic and DJ bio in the listener UI
> From the Discord thread [More info available for djs and show descriptions on front end gui](https://discord.com/channels/1514647956333002812/1525633155581739099) (Mrain1, 2026-07-11), carved out of #1568.

The limits on show topics (#722) and persona souls (#1105) were raised, and the public roster read exposes descriptions, but the listener UI still cuts them off with no way to read the rest. The schedule drawer in the classic skin clamps the show topic to three lines (`web/components/skins/classic/drawers/ScheduleDrawer.tsx`) and the persona line shows the name only.

**Proposal.**

- Tapping a clamped topic expands it in place. Tapping again collapses it. Show the affordance only when the text overflows, so short topics do not grow a control they do not need.
- The persona name in the same card opens the roster entry for that DJ, with whatever description the public read exposes. Souls stay behind `publishPersonaSouls`; this must not widen the public shape.
- Same treatment for the upcoming-show rows in that drawer.

Skin-scoped to classic unless the drift skin shows the same text, which it does not today.

Related: #722, #1105, `web/CLAUDE.md` skin contract.

perminder-klair:
Closes #1621.

## Problem

In the classic skin's schedule drawer the show topic was clamped to three lines with no way to read the rest, and the persona line showed only a name — the raised limits on show topics (#722) and persona souls (#1105) never reached the listener. The upcoming rows carried no topic at all.

## Fix

One file: `web/components/skins/classic/drawers/ScheduleDrawer.tsx`.

- **`useClampOverflow(text, expanded)`** measures `scrollHeight − clientHeight` (1px of slack, below one line at any of these type sizes) and re-measures through a `ResizeObserver` — the drawer is a sheet that changes width — and whenever the text changes. It measures **only while collapsed**: expanding removes the clamp, so `scrollHeight === clientHeight` there and a live measurement would report "fits" and pull the control out from under the finger that just used it. The last collapsed reading stands for the whole expanded pass.
- **`ExpandableText`** renders a plain block with no control when the text fits. When it overflows, the whole paragraph becomes the control — a real `<button type="button">` with `aria-expanded`, keyboard reachable, carrying `v3-focus` and a small `More`/`Less` line. Clamping is visual only, so the full topic is in the DOM either way and a screen reader never loses the tail.
- The **on-now card**'s topic moves from a hard `line-clamp-3` to that component; the **upcoming rows** get the same treatment and now carry the topic at all, clamped to two lines. Rows go `items-center` → `items-start` with a 3px nudge on the time column, since they are no longer single-height.
- **`PersonaName`** turns the persona name — on the card and on every upcoming row — into a disclosure button (`aria-expanded` + `aria-controls`) that opens that DJ's roster entry underneath it. A persona with nothing to say stays plain text rather than growing a control that opens an empty panel. Guest credits stay outside the control. The blurbs run through the same overflow-only clamp, because a published soul is a system prompt and can run long.

**The public shape is untouched.** The disclosure reads the persona index `/schedule` already sends: `tagline` always, `soul` only when `privacy.publishPersonaSouls` is on — `publicPersonaShape` already makes that call, absent rather than empty when off. No new fetch, no controller change, no new field.

Skin-scoped to classic, per the `web/CLAUDE.md` skin contract: co-located styles, core contexts only, nothing in `globals.css`.

## Verified

- `npm run lint` in `web/` (`eslint . && tsc --noEmit`) — 0 errors. The 5 warnings are pre-existing, in `components/admin/playlist-builder/generate.ts` and `components/ai-elements/file-tree.tsx`.
- `npx next build` — full production build succeeds.
- The `devOptional` lockfile churn `npm install` left behind was reverted, so the commit is that one file.

## Left out

- **No live render check.** Exercising the drawer needs the stack up for `/schedule`, which was out of scope here — the expand/collapse gesture and the overflow measurement are unverified against a real browser render.
- **No controller change and no controller test**: nothing on that side moved. `web/` has no test suite.
- **Drift and the other skins untouched** — they don't render this text today.
- The disclosure stays inline rather than becoming a nested dialog inside the drawer sheet: no second focus trap, and it matches the expand-in-place gesture the issue asks for.


perminder-klair:
Holding this one purely on verification, not on anything wrong with the change.

It is the only one of the six with no automated coverage at all, and that is structural rather than an omission: `web/` has no test suite, so there is nowhere for a test to go. What it has is `eslint` + `tsc --noEmit` clean and a successful `next build`, which between them prove it compiles and type-checks — not that the disclosure works.

The two things that most want a real render are the parts a build cannot check: the overflow measurement that decides whether the affordance appears at all (a short topic must not grow a control it does not need, which is the whole point of measuring rather than always showing it), and the expand/collapse gesture on the upcoming rows after the `items-center` → `items-start` alignment change.

I ran the other five on a live station but deliberately did not start the stack for this one, so none of that has been looked at on screen. Small change and low blast radius, so merging and checking it on the station is a perfectly reasonable call — it just should be somebody's call rather than something that slides through on a green tick that never tested the behaviour.

No controller changes here, and `publicPersonaShape` already gates `soul` behind `publishPersonaSouls`, so the public shape is not widened.

**Minor:** #1627, #1623 and #1625 have landed and this branch is three behind develop. It touches one file none of them went near, so it should merge without incident, but the current CI run is against the old base.

perminder-klair:
Pushed `216facf9` (plus a merge of `develop`, so CI now runs against the current base — none of #1625/#1623/#1627 touches `web/`).

**Fixed**

- **The overflow measurement was attached to a dead node.** Finding overflow swaps `ExpandableText`'s root from a `<div>` to a `<button>`, so React remounts the measured span — but the effect's deps (`[text, expanded]`) hadn't changed, so it never re-ran and the `ResizeObserver` kept watching a node that had left the DOM. Rotating a phone portrait→landscape left `More` sitting on a topic that now fitted, and only an expand/collapse round trip cleared it. The measured node is now a callback ref held in state, so it is itself a dependency and the observer follows the live span. This is the one that made the doc-comment's "re-measures on resize" untrue in practice.
- **Late web fonts could silently swallow the affordance.** A font swap that pushes a topic from 3 lines to 4 doesn't resize the box — at `line-clamp-3` the height is already pinned — so the observer can't see it, and the control never appears on exactly the text that needs it. One extra measure on `document.fonts.ready`.
- **`aria-controls` named an id that wasn't in the DOM** while the persona panel was closed. Now omitted until it's rendered.
- **On-now disclosures now key on the show id**, so an hour rolling over doesn't hand the incoming show an already-expanded topic under the outgoing DJ's open panel.
- **Restored the `truncate` dropped from the schedule row's persona name** — the show name above it still truncates, so the two lines of one row had started disagreeing about overflow. It rides the name element rather than the wrapping line, because `overflow: hidden` on an ancestor would clip the button's `v3-focus` ring (2px outside its box) and take the keyboard focus indicator with it.
- **`whitespace-pre-line` on the blurbs**, so an operator's paragraphing in a published soul survives instead of collapsing into one blob.
- **`cursor-pointer` on both new controls** — discoverability is the feature, and neither had a hover affordance.
- **A comment on the `cn('block', …)` line.** Tailwind emits `.block{display:block}` *after* `.line-clamp-N{display:-webkit-box}`, so on raw cascade the clamp would lose and the whole feature would be inert. It works only because `twMerge` resolves the pair, and it resolves it one-directionally — swapping those two arguments disables everything with no lint error and no type error. Worth a note in the file rather than a rediscovery later.

**Deliberately left alone**

- **The paragraph-as-button shape.** Making the whole topic the control means its accessible name is the entire paragraph, and since the full text is in the DOM either way a screen-reader user gains nothing from the disclosure. A `<p id>` plus a small sibling `<button aria-controls>` would read better. But the whole-paragraph tap is the gesture #1621 asks for, the current form *is* keyboard-operable with `aria-expanded`, and the trade is a judgement call rather than a defect — leaving it as authored.
- Two side effects of that shape are worth knowing rather than fixing: the topic text is no longer drag-selectable (browsers don't select inside a `<button>`), and a stray drag across the paragraph collapses it on mouseup.

**Still owed**

The on-screen verification called out above hasn't happened. `web/` has no test suite, so `eslint` + `tsc --noEmit` + a green build prove it compiles, not that the disclosure works — and the two things most wanting a real render are unchanged: whether the affordance appears only when the text actually overflows, and the expand/collapse gesture on the upcoming rows after the `items-center` → `items-start` change. The remount bug above is precisely the class of thing a build can't see and one rotation on a phone would have.

`cd web && npm run lint` → exit 0. Five warnings, all pre-existing and in files this PR doesn't touch (`playlist-builder/generate.ts`, `ai-elements/file-tree.tsx`); nothing new, nothing in `ScheduleDrawer.tsx`.

Issue #213, opened 2026-05-30
Pocket-TTS no reference audio folder support for cloning


WhatKateDoes:
Pocket-TTS as implemented in the TTS-heavy profile does not have a folder like chatterbox-voices in state for housing reference audio files to be selectable voices under personas.

WhatKateDoes:
Recommend using a shared folder for chatterbox and PocketTTS voices

WhatKateDoes:
Note as well that the voice selector needs scroll function where there are many voices in the folder.  Workaround is to zoom screen down to be able to select when list is long.

perminder-klair:
Thanks for flagging both of these, fix is up in #217.

Shared voice folder. Both Chatterbox and PocketTTS now read reference WAVs from a single state/voices/ directory. Drop a ~5-second clean-speech WAV in there and it'll show up under both engines' voice dropdowns on next reload. The old state/chatterbox-voices/ location is still scanned, so you don't have to move anything, but new uploads can go to the shared path.

PocketTTS cloning. While I was in there, PocketTTS now also supports zero-shot voice cloning from those same WAVs (it always could under the hood, the wrapper just didn't expose it). On the Personas page you'll see two groups under PocketTTS. Built-in (alba, anna, charles, …) and Custom (cloned) listing your WAVs.

Voice selector scroll. That was a real bug in the dropdown component, the scrollable area was pinned to the height of one row, so any list longer than that needed a browser zoom-out to reach. Removed the offending CSS rule; the dropdown now scrolls properly with mouse wheel + the up/down chevrons, no zoom needed.

Should land on develop shortly and ship in the next release.

PR #1628, opened 2026-09-07
feat(personas): export and import a persona as a single zip bundle (#1620)

### LINKED ISSUE #1620: feat(personas): export and import a persona as a single bundle
> From the Discord thread [Sharing DJs/Stations/etc](https://discord.com/channels/1514647956333002812/1532786180947710083) (PwrBlackout, 2026-07-31), carved out of #1568. Widens FR 11 on the #1485 tracker.

Sharing a DJ today means copying the persona JSON by hand and separately moving whatever audio it depends on. A persona with a cloned voice references a `.wav` sample by filename (Chatterbox and PocketTTS, #213), and that file does not travel with the JSON. FR 11 asks for a programmatic import of the JSON alone, which leaves the voice behind.

**Proposal.** One zip per persona.

- `GET /personas/:id/export` returns a zip containing the persona JSON, its voice sample if the engine uses one, and any jingles that name this persona. The manual backup export in `routes/backup.ts` already builds a zip of state files and is the precedent for the packaging.
- `POST /personas/import` accepts the zip, validates the JSON through the existing persona schema, writes the audio files into the state dirs they are read from, and creates the persona. The community install route (`POST /personas/community/:slug/install`) already does the create half from a fetched definition, so the import should share that code rather than add a second writer.
- Name collisions on audio files get a suffix, never an overwrite. An import must not replace a sample another persona is using.
- A "Share" button on the persona editor downloads the zip, and an "Import" button on the personas list uploads one.

**Not in this issue.** Whole-station export. The backup export already covers that, and a station bundle is mostly a backup with the secrets stripped, which is a different problem.

Related: #1485 (FR 11), #213, `routes/backup.ts`, `routes/personas.ts`.

perminder-klair:
Closes #1620.

## Problem

Sharing a DJ meant pasting the persona JSON by hand and separately remembering to move whatever audio it depended on. A cloned voice is the case that breaks: chatterbox and pocket-tts read `tts.voice` as a **filename** in the shared voice folder (#213), so the JSON alone arrives pointing at a WAV that is not there and the persona fails its synth on every line.

## Fix

One zip per persona.

- **`GET /personas/:id/export`** — the persona JSON, the reference WAV its own slot names, and the operator's own jingles whose text names this DJ. Packaging follows `backup/zip.ts`: an `AdmZip` the caller turns into a response body, a manifest with a format + version checked before anything is touched, member paths mirroring the state-dir layout, and `contents` listed before the manifest is added so it never names itself. `id` and `avatar` are stripped — they are station-local.
- **`POST /personas/import`** — validates the manifest, then the JSON through the existing `personaSchema`, then every member's shape, then the roster's room for one more name. Only then is anything written. Route-scoped `express.raw`, the same shape as `POST /backup/import`.

### The rules that carry it

- **`personas/install.ts` is the ONE create path.** The community install and the bundle import differ only in where the persona object came from; the roster cap, the duplicate-name 409, the strict validation and the id minting are one decision made once. `POST /personas/community/:slug/install` was moved onto it rather than a second writer being added next to it.
- **A clashing audio filename is suffixed, never overwritten — and the incoming persona is re-pointed at the name it actually got.** `tts.voice` is a bare basename in the shared voice folder, so an overwrite silently swaps the voice of whoever already pointed there (which is why `voice-library.importVoice` refuses a clash outright; an import cannot refuse, because the operator has no way to rename a member inside a zip they were handed). A suffix nobody re-points at is just an orphan. The comparison is case-**insensitive**: on a case-folding filesystem a case-sensitive check hands back a "free" name that overwrites on write. `voice-library.adoptVoice` and `jingles.adopt` do the writing, each still the sole owner of its folder; the jingle check unions the sidecar with the directory listing, so a file with no sidecar entry still counts as taken.
- **Only a persona's OWN slot exports a sample.** `personaCloneVoice` reads the raw slot, not the one resolved against the station default — an `inherit` persona on a chatterbox station is speaking with the *station's* wav, which is not this persona's to give away.
- **"A jingle naming this persona" is matched on a word boundary.** Nothing on disk links the two (the sidecar carries only spoken text), so the name has to be matched in that text — and a plain substring test hands every stinger to a DJ called "Al". Scanned rather than regexed, since a persona name is free operator text and may hold metacharacters. Builtins never travel: the default ident is station furniture every install already has and cannot delete.
- **Everything that can refuse is asked before a byte is written**, so a rejected import leaves no litter in the state dirs the operator cannot see. Audio is then written *before* the persona, because `tts.voice` has to name the file it actually got and the suffixed name is only knowable after the write.

Nothing near the backup module's temp-file or `STATE_DIR` sweep rules was touched — the bundle is never written to `STATE_DIR`, so retention's anchored `subwave-auto-backup-…` grammar and `GET /backup/restorable`'s `*.zip` listing are unaffected.

### Admin UI

**Export bundle** on the persona editor, next to (and distinct from) "Share to community" — that one opens a GitHub form carrying only the portable knobs, this one hands you a file with the audio in it. Disabled on a new or dirty row, since the export is built from what the station has stored. **Import** on the personas roster, beside Community.

## Verified

- `npm run lint` clean in `controller/` and `web/` — 0 errors in both, warning counts unchanged.
- `npm test` in `controller/`: **1483 pass, 0 fail**.
- `npm run gen:schemas` produces no mirror drift.
- `web/scripts/audit-admin-query.mjs` reports only a pre-existing `BrainSection.tsx:162` violation — confirmed present on the unmodified branch by stashing this work and re-running. Its self-tests: 9/9.

New: `controller/scripts/persona-bundle.test.ts`, 11 cases — the pure rules (suffixing, the word boundary, which engines name a sample, member paths), the export's exact contents, the duplicate-name 409 leaving the voice folder untouched, and a full round trip proving the suffix lands *and* that the persona points at it.

The traversal guard is pinned at the pure layer rather than through a crafted archive, because `adm-zip` normalises `..` out of an entry name as it writes — a traversing bundle can be received but not built with it.

## Deliberately left out

- **The avatar.** The issue scopes the payload to the JSON, the voice sample and the jingles; an avatar file is keyed on the persona id the bundle drops. Imports arrive with `avatar: ''`.
- **Whole-station export**, explicitly out of scope — `GET /backup/export` already is that.
- **A duplicate name is a 409, not an update-in-place.** That is the community install's existing rule, inherited by sharing its writer: the alternative leaves the operator two indistinguishable DJs in every picker that names one. Importing over an existing persona is not a gesture this issue asked for.
- The editor button says **"Export bundle"** rather than the issue's "Share" — "Share to community" already sits in that footer and means something else. Happy to rename if the original wording is preferred.


perminder-klair:
Exercised this against a running dev stack with a real `state/` (six personas), rather than temp dirs.

**Export.** `GET /personas/:id/export` returns a 943-byte zip: `manifest.json` + `persona.json`, `voice: null` and `jingles: []` for a cloud-engine persona with neither. `id` and `avatar` are dropped as documented.

**Auth.** Both routes are `requireAdmin` and return 401 unauthenticated. Worth stating explicitly since the bundle carries `soul` in full — an unauthenticated export would have been a privacy-lock leak, and it is not one.

**Duplicate name.** Re-importing a persona's own bundle returns 409 `a persona named "Heer" is already in the roster`, roster still 6, no litter in the state dirs.

**Malformed persona.** A bundle whose `tts` block fails the schema is refused 400 with the specific issue (`tts.cloudProvider must be one of: ...`) and writes nothing — the "collect audio up front so a malformed one refuses the whole import" comment holds in practice.

**Collision rule — the one I most wanted to see, and it is correct.** Imported two bundles naming the same reference WAV with different bytes:

```
import 1 → state/voices/sample-voice.wav        (md5 6e5cd6c0…)
import 2 → state/voices/sample-voice-2.wav      persona re-pointed at sample-voice-2.wav
           sample-voice.wav still md5 6e5cd6c0…  (untouched)
```

Suffix, not overwrite, and the incoming persona's `tts.voice` follows the stored name. That is the acceptance criterion, met on a real state dir.

**One edge worth a look: a clone-voice persona can import with no voice at all.**

`applyPersonaBundle` re-points `tts.voice` only `if (voiceMember)`. When the bundle carries no `voices/` member but the persona's engine reads `tts.voice` as a reference WAV, the import returns 200 and installs a persona pointing at a filename that need not exist locally — the synth then fails on every line, which is the exact failure the module header opens by describing.

I hit this by accident (my hand-built zip used `voice/` rather than `voices/`, and an unrecognised member is ignored rather than fatal, which is reasonable on its own). But there is a non-synthetic path to the same place: `buildPersonaBundle` only adds the WAV `if (found)`, so exporting a persona whose sample has been deleted from the source station produces exactly this bundle — `manifest.voice: null` with a persona JSON still naming the file.

Cheapest fix is probably at export: refuse, or warn, when `personaCloneVoice(persona.tts)` is set and `resolve()` misses, rather than shipping a bundle that cannot work. Alternatively the import could say so. Either way it seems better than a silent 200.

Not a blocker on the collision behaviour, which is the part the issue actually turns on.

perminder-klair:
Holding this one on the edge from the testing above, rather than on anything about the collision rule — that part is correct and I verified it against a real state dir.

**What I'd want changed first.** `applyPersonaBundle` re-points `tts.voice` only `if (voiceMember)`, so a bundle whose persona names a clone voice but carries no `voices/` member imports 200 and installs a DJ pointing at a file that need not exist. The synth then fails on every line, which is the failure the module header opens by describing.

It is reachable without hand-editing a zip: `buildPersonaBundle` adds the WAV only `if (found)`, so exporting a persona whose sample has been deleted from the source station produces exactly that bundle — `manifest.voice: null` beside a persona JSON still naming the file. Sharing a persona whose sample went missing is a plausible thing for an operator to do, and the bundle it produces is silently broken at the far end.

Cheapest fix looks like the export side: refuse, or warn, when `personaCloneVoice(persona.tts)` is set and `voiceLibrary.resolve()` misses, rather than shipping a bundle that cannot work. The import side saying so instead would also do. Either beats a silent 200.

**Also needs a refresh before merge.** #1627, #1623 and #1625 have since landed on develop and this branch is three behind. It still auto-merges, but it shares `controller/CLAUDE.md` with two of them — worth merging develop in and re-running the suite rather than merging on the strength of a CI run against the old base.

perminder-klair:
Pushed `f44f0219` — the four blocking findings from the review, plus the two cheap correctness notes. Branch is also caught up with develop (#1627, #1623, #1625 merged in, no conflicts) and the whole gate re-ran against that base.

**The one that mattered most: line injection into `jingles.m3u`.** `jingles.adopt` validated an adopted filename with `isAcceptedAudio`, which only looks at the extension. `rewritePlaylist` joins the folder and the name with `\n`, and Liquidsoap reads that playlist with `reload_mode="watch"` — so a bundle member named `one.wav\n<anything>\ntwo.wav` wrote its own extra line straight into the station's jingle rotation, no restart needed. Reproduced before the fix, and the test fails without the guard. Adopted names are now pinned to a character class: `JINGLE_FILENAME_RE` for a stinger, and the persona schema's own `TTS_CHATTERBOX_VOICE_RE` for a reference WAV — so a name the voice library accepts is a name the save accepts.

**"Everything that can refuse is asked before a byte is written" is now true rather than aspirational.** Three refusals used to fire after audio had landed — the voice filename (`adoptVoice` took any `*.wav`, the suffixed name then went onto `tts.voice` and `validatePersonasStrict` threw), a malformed member behind a good one, and the strict save itself. A refused import left a WAV in `state/voices/` and a stinger in `jingles.json` **and** `jingles.m3u`, which is audio on air for a DJ that was never created. The import now checks every member's name and size against the predicates the writers own, **reserves** the filenames (`voiceLibrary.reserveVoiceName` / `jingles.reserveNames`), puts the reserved voice on the persona, and dry-runs `validatePersonasStrict` over the resulting roster — only then does anything touch disk.

I took reserve-then-write over rollback-on-failure as suggested, and it is the right shape here for a reason worth stating: `jingles.adopt`'s write is three files (the WAV, the sidecar, the playlist), so unwinding a partial import means reversing another module's multi-file write from the outside — exactly the second writer this feature is otherwise careful not to add. Reserving is only sound one import at a time, so `applyPersonaBundle` now serialises; imports are rare and operator-driven, so the queue costs nothing.

**No more silent 200 on a mute DJ** (the one you held this on). Export refuses **409** when `personaCloneVoice` is set and `voiceLibrary.resolve()` misses — that bundle is well-formed, `manifest.voice: null` beside a persona JSON still naming the file, and installs a DJ that fails every line. Import refuses the mirror case too, so a hand-built zip is covered — **unless** the receiving station already has that sample, in which case the persona works exactly as its JSON asks and refusing would block the ordinary "we both already have this voice" share.

Also: **24-jingle cap** per bundle (each member is a file, a sidecar row and a playlist line, and the admin UI deletes them one at a time); `entry.header.size` checked **before** `getData()`, so a member that expands past the 25 MB cap is refused rather than materialised to measure it; and `manifest.jingles[].text` bounded, since it is the one string copied to disk without the persona schema ever seeing it.

`controller/CLAUDE.md` and the `bundle.ts` / `install.ts` headers now describe what the code actually does — the old wording asserted the invariant the code did not have.

**Tests.** 8 new cases. The important change is what they assert: a refused import is now checked against the jingle dir, `jingles.json` **and** `jingles.m3u`, not just `state/voices/`. Checking only the voice folder is precisely why the half-write slipped through the first time. Verified the three guard tests fail with the guards removed.

**Deferred, deliberately** — all from the non-blocking list, none of them correctness:
- `uniqueFilename` / `JINGLE_NAME_MAX` living in `personas/bundle-pure.ts` while `audio/` and `broadcast/` import them. Real layering smell, but moving it is a rename touching three modules and reads better as its own change.
- The loudnorm asymmetry between `adopt` (verbatim) and `importAudio` (levelled) — documentation, not behaviour.
- The slug grammar restated in `PersonasPanel.tsx` alongside `bundleFilename`.
- The "Export bundle" button title not mentioning that the zip carries `soul` in full. Worth doing; it is copy, and the export is `requireAdmin`, so no privacy-lock rule is at stake.

**Gate, re-run against the merged base** (`develop` @ `dd46fcc1`):
- `controller && npm run lint` — pass, 0 errors (706 warnings, all pre-existing `no-explicit-any`)
- `controller && npm test` — **1522 pass / 0 fail**, 222 files, 101s
- `controller && npm run gen:schemas` — clean `git status` after, no mirror drift
- `web && npm run lint` — pass, 0 errors (5 pre-existing warnings)
- `web && node --test scripts/audit-admin-query.test.mjs` — 9/9

Issue #1622, opened 2026-09-07
Follow-up tracker: pending roadmap and support items from #1485


perminder-klair:
Successor to #1485, which succeeded #1331. This tracker carries only the items that were still unchecked when #1485 closed. Completed delivery history stays in #1485 and #1331; the original validation notes, file/line references and reporter context live in #1300.

> **Carry-forward audit — 2026-09-07.** Every entry below was re-checked against `develop` at b479005 before being copied, not just copied. What that check changed is recorded per item: **FR 11 came off the list entirely** (superseded by #1620), and **FR 4 got more urgent, not less** (FR 3 shipped the album cooldown that now blocks the thing FR 4 wants to queue).

> **Delivery update — 2026-09-07.** FR 4, FR 5a and FR 14 are done and merged to `develop`. Their entries are kept below with their PRs rather than deleted, so the reasoning stays readable next to the items still open. One item that was **not** on this list shipped with them: see "Found on the way" at the end.

## Feature requests

- [x] **FR 4** — album/block queueing. **Done — #1632.** `POST /dj/queue-block` (album or artist run) as one operator press, `DELETE /dj/queue/block/:blockId` to cancel the unaired remainder. The framing in the original entry turned out to be wrong in a useful way: it is **not** a bypass feature. `picker.albumHours` and the artist guard are PICK paths an operator push never reaches, and `queue.recentAlbumKeys` already walks `upcoming`, so a queued block stops the picker adding more of the same record behind it — there was nothing to opt out of. The exemptions a block needs are the ones `POST /dj/queue-track` already carried. The blocklist is *not* bypassed: a blocked track drops from the block and is named in the response. Not yet exercised against a live station — the Subsonic resolution, the telnet pull-back in the block cancel and the admin UI are typechecked but unrun.
- [x] **FR 5a** — vocal-aware timing for `announce()` segments. **Done — #1629.** The lever is timing, not a trim: a scheduled segment is rendered before the boundary it will take is known, so a clip that would outlast the runway keeps its slot and takes the next boundary, bounded by the existing `PENDING_VOICE_MAX_AGE_MS`. New `broadcast/vocal-runway.ts` owns the composition of `rampBudgetMs` with `shiftOnsetMs` that previously sat inline in `maybePushBed`, so the bed and the link read one spelling rather than three. Worth watching in the booth log: this adds a second reason a segment can be held at a boundary, which on a catalogue with consistently early vocal entries raises the odds of hitting that staleness drop.
[...]

Issue #1485, opened 2026-08-26
Follow-up tracker: pending roadmap and support items from #1331


perminder-klair:
Successor to #1331 after its shipped and follow-up work was completed. This tracker carries only the items that were still unchecked when #1331 closed on 2026-08-26.

Completed delivery history remains in #1331. The original validation notes, file/line references, reporter context and rationale live in #1300.

> **Carry-forward audit — 2026-08-26.** These entries were copied from the live #1331 checklist after Bug 8 merged in #1484. “Pending” includes partially documented work whose requested operator-facing outcome is still missing.

## Feature requests

Roughly in support-traffic order. Each item still needs a design decision before code.

- [x] **FR 3** (#1563) — album-level cooldown in the picker; `music/recency.ts` has no album key, and it has to live at the artist-guard choice point, not in the discovery tools (#618)
- [ ] **FR 4** — album/block queueing; needs an explicit “operator block” flag that bypasses pair-drain and the artist guard, since a queued album is exactly what those prevent
- [ ] **FR 5a** — vocal-aware timing for `announce()` segments (station IDs, weather, hourly, skills — only links and request intros get the vocal hint today)
- [x] **FR 5b** (#1562) — “talk only between tracks” scheduling constraint on `dj-gate.shouldFire` (segments currently fire on wall-clock minutes)
- [x] **FR 5c** — bed tail length (#1499): `beds.tailSec`, default 3s. Not the tuning ask it read as — the bed's own exit cross was missing from `bedLengthFor`'s sum, so the quiet was a residual (`tail - cross`) and ran to **minus 4s** at the defaults, i.e. the bed was talking over the song it exists to protect
- [x] **FR 6** (#1560) — duck depths (0.22 / 0.30 literals in `radio.liq`) as settings via the existing `liquidsoap_*.txt` pattern; a true voice-only interlude (music paused, not ducked) is a separate, larger change
- [ ] **FR 8** — non-English DJ discoverability: connect the persona language field, a matching-language voice and house rules for orthography in one operator-facing explanation
- [ ] **FR 9** — per-show phased briefs; extending the programme producer plan to carry per-hour music steering looks much cheaper than a new show schema
- [ ] **FR 11** — programmatic persona import (file drop / `POST /personas`); the community-install route is the closest precedent
[...]

PR #1629, opened 2026-09-07
feat(broadcast): vocal-aware timing for scheduled segments (#1622 FR 5a)

perminder-klair:
Closes FR 5a of #1622.

## The gap

`enforceIntroBudget` — the never-talk-over-a-singer rule — was applied in exactly one place: `broadcast/dj-agent/enqueue.ts` on a pick's link. Every other spoken segment (station idents, the hourly check, weather, skill segments) reached the air with no budget at all, and since #1485 FR 5b `djTalkOnlyBetweenTracks` reads every talk row as `'next-track'`, those segments land on the head of a track — the same runway the link path has always been trimmed against.

## Why the lever is timing, not a trim

The link path can trim because the song is known before the TTS runs. A scheduled segment cannot: `announce()` renders the WAV and only then reads the talk-air scope, `announceAtNextTrack` renders early by design to keep TTS latency off the air path, and `airPendingVoice` may hold the result across several boundaries. By the time the incoming track is a fact there is no text left to trim.

So a segment that would outlast the runway does what a segment already does when a boundary is busy (`boundaryCarriesTrackVoice`): it keeps its slot and takes the next boundary. Postpone, never cancel; nothing is regenerated, and the existing `PENDING_VOICE_MAX_AGE_MS` staleness check is what bounds the wait rather than a new timeout.

## What changed

New `controller/src/broadcast/vocal-runway.ts`, a policy module in the sense the root CLAUDE.md means it:

- `vocalRunwayMs(track)` — the one composition of `bed-policy.rampBudgetMs` (three-state: measured onset / `Infinity` instrumental / `null` un-analysed) read through `silenceTrim.shiftOnsetMs`. That composition previously sat inline in `queue.maybePushBed`; it now reads it from the module, so this is a **second reader of one spelling rather than a third opinion** — which is the failure the comment at that call site already recorded (the prompt told the runway was 2s while the bed decision still thought it was 8s).
- `segmentFitsRunway(voiceMs, runwayMs)` plus `VOCAL_RUNWAY_FLOOR_MS` / `VOCAL_RUNWAY_CEILING_MS`, which `intro-budget.ts` now imports instead of restating 2500/18000. `llm/internal/prompts` → `broadcast` policy is the existing import direction (`clock-policy`, `announce-line`).

`airPendingVoice` gains one gate, sited right after the existing busy-boundary hold.

## Three deliberate asymmetries

- **A clip ≥ 18s airs, and that test is asked first.** No runway in the binding band can hold a news read, so refusing boundaries for it would only ride the pending slot to the staleness drop and lose it unaired — strictly worse than the talk-over it already did. Long segments are byte-identical.
- **A measured onset < 2.5s refuses every clip.** That floor's leniency belongs to the energy heuristic `intro_ms`, not to a measured vocal.
- **The immediate placement gets no budget, deliberately.** It airs mid-track through `say.txt` under the heavy duck (0.22), where the vocal onset is behind, not ahead — there is no runway to land inside, and the heavy duck is already the answer chosen for a placement with no runway. Relocating such a segment to a vocal-free window is exactly what `djTalkOnlyBetweenTracks` is for.

## Upgrade safety

`null` (un-analysed) fails open at every gate, so a station with no Demucs pass behaves exactly as before, end to end.

## Reviewer's note

This adds a second reason a scheduled segment can be held at a boundary. On a station whose catalogue has consistently early vocal entries, that raises the chance a segment reaches the existing 20-minute staleness drop rather than airing late. The drop is logged with its reason, and the bound is the pre-existing one — but it is a real behaviour change and worth watching in the booth log rather than discovering later.

## Verification

- `npm run lint` in `controller` — 0 errors.
- `npm test` — 1513 pass / 0 fail, including 11 new tests in `scripts/vocal-runway.test.ts`. One of them pins `vocalRunwayMs` equal to `intro-budget.firstVocalMsFor` on the same track through a real library DB, so the two readers cannot drift apart again.


perminder-klair:
Rebased onto `develop` after #1624, #1626, #1628, #1630 and #1631 landed.

Only `CLAUDE.md` conflicted — `queue.ts` and `docs/internals/broadcast.md` auto-merged. The conflict was #1624 extending the talk-scheduler bullet while this branch appended a new one after it; resolved by keeping develop's extended bullet and appending the FR 5a bullet after it.

**On the #1624 interaction**, since both changes land in `queue.ts`: there isn't one. `airPendingVoice` now carries three gates in order — staleness, the existing `boundaryCarriesTrackVoice` hold, and this PR's runway hold — and none of #1624's rotate logic appears in that method. The rotate acts a layer below, through `jingle-playing.json` holding `airVoice`, and its talk-table row is a *fill* row that stands down whenever a slot row wants the minute, so it cannot take a scheduled segment's minute away. The two compose rather than compete.

Re-verified on the rebased branch: `npm run lint` 0 errors, `npm test` 1567 pass / 0 fail.

Issue #1568, opened 2026-09-05
Discord triage: 20 suggestion threads where related work landed but the ask did not


perminder-klair:
> Sourced from the community #suggestions forum on Discord. Full cross-reference (all 65 open threads against every GitHub issue) lives in [`docs/discord-open-suggestions.md`](../blob/develop/docs/discord-open-suggestions.md).

For each of these, related work has landed but the specific ask has not. They are the most likely to be mistaken for done: someone sees the shipped half, closes the thread, and the requested outcome never arrives.

Each entry needs one decision — carve the remainder into its own issue, or reply saying the shipped half is the answer and close it.

- [x] **[Full album show and special artist show](https://discord.com/channels/1514647956333002812/1545685340487024681)** — elysium, 2026-09-05
  Hello I'd like to create two shows: a show that plays an entire album chosen by the DJ a show that plays tracks by a single artist chosen by the… **Closed — single-artist works via playlist-anchored shows (#701); album-as-a-block stays FR 4 on #1485.**
  Single-artist shows already work through playlist-anchored shows (#701, tightened by #1533). Album-as-a-block is FR 4 on the tracker and still unstarted.
- [x] **[More control on spoken segments](https://discord.com/channels/1514647956333002812/1544677009131573248)** — shinjuku70, 2026-09-02
  Hello! It would be great to have more control on which spoken segments the DJs can use on their shows. Sometimes I would prefer to disable most of…
  Several dials shipped piecemeal — per-skill quips optional (#471), clock switch (#1286, #1380), talk-only-between-tracks (#1562). A per-segment-kind enable/disable for a show is not tracked; vocal-aware timing for announce() is FR 5a, still open. **Closed — tracked by #1249**, which asks for exactly this (per-show enabled skills, speech frequency, DJ mode) and is better specified than the thread.
- [x] **[Genre Leans within shows](https://discord.com/channels/1514647956333002812/1538883985097097246)** — KeithyMcC, 2026-08-17
  I'm finding when i'm selecting genres within the genre leans section, i'll type a genre for example "trance" when i click one, it reverts back to…
[...]

Issue #619, opened 2026-06-26
Bug: Duplicate track queued when concurrent listener requests resolve to the same song


geekylakshya:
## Bug

When two listener requests are submitted in quick succession and both resolve to the same song, the track is queued twice and plays back-to-back.

## Background

Issue #538 (fixing #537) intentionally scoped the `queue.push()` deduplication guard to `aiPicked === true` only, with the reasoning:

> "Human requests pass through unconditionally (no false 'coming up')."

The goal was to avoid acknowledging a listener request ("coming up!") and then silently dropping it because the song was already queued. That change correctly addressed the case where the DJ agent autonomously re-selected the currently playing track.

This issue exposes a different race condition that was not covered by that change.

## Root cause

The request pipeline can take **25 to 45 seconds** when `identifyRequestedTrack` performs a web search, identifies the song, and matches it against the library.

During that window, two requests can independently conclude that the song is not yet queued:

1. Request A starts and reads `queuedIds()`. Song X is not queued.
2. Request B starts shortly after and also reads `queuedIds()`. Song X is still not queued because A has not reached `queue.push()`.
3. Both requests resolve to Song X.
4. Both call `queue.push()`.
5. The deduplication guard is skipped because `aiPicked === false`.
6. Song X is queued twice.

**File:** `controller/src/broadcast/queue.ts` (around line 288)

## Reproduction

1. Submit the same song request or the same lyrics request twice from `/listen` within a few seconds.
2. Wait for both agent runs to resolve.
3. Observe that the same track appears twice in the queue and plays back-to-back.

The race window is largest when `identifyRequestedTrack` is involved (approximately 30 to 45 seconds), but the issue can occur with any sufficiently slow request.

## Possible fix

The challenge is preserving the existing acknowledgement guarantee. Listeners should never receive a "coming up!" response if their request is silently discarded.

Possible approaches include:

* **In-flight deduplication**

  * Track song IDs currently being resolved by request agents.
  * If another request resolves to the same song, return an acknowledgement such as "Someone already requested that song. It's on the way."

* **Deduplicate at `queue.push()`**

  * Extend deduplication to listener requests.
  * Return a distinct status so the caller can send an honest acknowledgement instead of silently dropping the request.

* **Queue-aware request handling**

  * If the resolved song is already queued, avoid pushing it again.
  * Inform the listener that the song has already been requested and is waiting to play.

## Input needed

@perminder-klair

[...]

PR #1632, opened 2026-09-07
feat(dj): queue a whole album, or an artist block, in one press (#1622 FR 4)

perminder-klair:
Closes FR 4 of #1622. Discord demand for it is filed in #1568.

> **Stacked on #1631** — that fix is this PR's parent and should merge first. Review this diff as `fix/1622-studio-pushes-block-request-line..HEAD`.

## What this is, and what it is not

One press instead of twelve, in the right order, with an honest report of what the never-play list refused. **It is not a bypass feature**, and that is worth saying because the tracker framed it as one.

Every guard the tracker named was checked against the code. `picker.albumHours`, the artist guard — those live in `filterPickerCandidates` and `dj-agent/*-guard.ts`, which are **pick** paths, and an operator push reaches none of them. What the album cooldown does to a queued block is already right: `queue.recentAlbumKeys` walks `upcoming`, so the block's presence stops the picker adding more of the same record behind it. There was nothing there to opt out of.

The exemptions a block does need are the ones `POST /dj/queue-track` already carries, carried exactly as before: `allowDuplicate: true` past the #619 dedup guard, and `requestedBy: 'studio'` past the length cap, the show-boundary cut and the bed's request reason. **No new bypass, and the queue gains no new gate.**

## The one thing that is not bypassed

The never-play blocklist. It is absolute, requests included, and a block of thirty is no different. `hitOf` is read in the pure planner so the response can *name* what was skipped, but the refusal still happens where it always has, in `queue.push()` — a reporting read of the existing chokepoint, not a second rule-filter. The route pushes every planned track and records a `-2` return as a skip whether or not the plan predicted it. One blocked track drops from the block and is named; the rest queue. A 409 for the whole album because track 7 is blocked would be the wrong answer.

## API

`POST /dj/queue-block` (admin), modelled on `POST /library/blocklist`'s resolve-from-a-track-row precedent:

```
{ kind: 'album',  trackId | albumId,  order?: 'natural' }
{ kind: 'artist', trackId | artistId | artist, limit?: 10, order?: 'natural' | 'shuffle' }
```

- Album order is `(discNumber, track)` with the returned order as a stable tiebreak. Navidrome's order is conventional, not contractual, and this is the one queue where getting it wrong is the whole failure.
- `order: 'shuffle'` and `limit` on an album are **refused, not ignored** — silently discarding a field the caller sent is how two fields come to disagree about what was asked for.
- Artist blocks resolve through `getTopSongs`, falling back to an album walk when it returns nothing. That fallback is load-bearing: `getTopSongs` is empty for obscure artists, which is exactly the niche catalogue #618 was measured on.
- `QUEUE_BLOCK_MAX_TRACKS = 30`; a longer record is truncated and the truncation is **reported**. Refusing a double album outright is worse; silently dropping six tracks is worse still.

`DELETE /dj/queue/block/:blockId` cancels the unaired remainder as one action. An operator who can queue thirty tracks with one press must be able to undo it with one.

## Where the block lives

Plain `upcoming` entries, appended in order, one `queue.push()` per track — no parallel structure, no second writer, no new drain path. FIFO is the ordering guarantee already. `push()` is declared `async` but has no `await` in its body, so a loop of pushes is atomic against the request path and nothing can interleave into a block.

The item gains `block: { id, index, size, label }`, **identity only, never read by any gate** — it buys the grouped admin row, the block cancel, and one booth-log line instead of twelve.

## Three consequences, each answered rather than papered over

- **A listener request lands mid-album.** It queues and appends, and the acknowledgement now names the wait. FIFO is not negotiable — `drain-policy.ts` states that a request *is* the successor arriving and that FIFO is never inverted; letting a request jump the block would break that and split the record, which is the one thing this feature exists to prevent. The listener is told instead, in a number they can act on.
- **A show boundary lands mid-album.** Warn, don't act. The route returns `runsPastShowChange` so the operator decides *before* committing — cutting the block contradicts "an explicit operator action always fires", and a mic-pass between two tracks of one album is worse than a missed one. `stopAtShowChange` is the obvious follow-up if wanted later.
- **`drainAction` is untouched.** A twelve-track block costs one drain pass N × (loudness resolve + handoff poll), and special-casing the drain for it would be a second copy of a decision at a pick path in different clothing. The honest lever is the cap.

## Forecast counts the bed

`airForecastSec` walks `upcoming`, and a bed is never an `upcoming` entry — `maybePushBed` writes straight to `next.txt`. It reuses the existing #1574 helper `bedDelayBeforeItemAirs` rather than re-walking. Missing it understates both callers in the direction that matters: the listener notice says a request is closer than it is, and `runsPastShowChange` reads "this fits" when it does not. Three tests pin it, including one asserting an unsent item ahead contributes **zero**, so nobody later "fixes" the sent gate and double-counts a bed that has not been pushed.

## Not exercised against a live station

The Subsonic resolution, the telnet pull-back in the block cancel, and the admin UI are typechecked and lint-clean but have not been run against a real Navidrome or mixer. The telnet path in particular is not reachable by unit tests. Worth a dev-stack pass before merge.

## Verification

- `npm run lint` — 0 errors in `controller`, `web` and `mcp-subwave`.
- `npm test` — 1553 pass / 0 fail.
- `npm run gen:schemas` run; mirror committed and re-verified in sync.
- `node web/scripts/audit-admin-query.mjs` byte-identical to the develop baseline (its one finding, `BrainSection.tsx:162`, is pre-existing); `audit-admin-query.test.mjs` 9/9.


perminder-klair:
Rebased onto `develop` and retargeted from the stacked base now that #1631 has merged.

#1631 was **squash**-merged, so `develop` carries a new commit rather than this branch's original parent. Rebased with `--onto origin/develop fix/1622-studio-pushes-block-request-line` so that parent is dropped rather than replayed — verified afterwards that the diff against `develop` no longer contains any of #1631's changes (`routes/request.ts` shows none), while `operator: true` and `pendingListenerRequests()` are still in place from develop.

Only `docs/internals/broadcast.md` conflicted; resolved by keeping develop's queue/playback bullet and appending the FR 4 one. `queue.ts`, both CLAUDE.md files and the schema mirror auto-merged.

Re-verified on the rebased branch:
- `npm run lint` — 0 errors in controller, web and mcp-subwave
- `npm test` — 1599 pass / 0 fail
- `npm run gen:schemas` leaves the mirror byte-identical

Still outstanding before merge: none of this has been exercised against a live station. The Subsonic resolution, the telnet pull-back in the block cancel, and the admin UI are typechecked but have not been run, and the telnet path is not reachable by unit tests.

PR #1634, opened 2026-09-07
feat(picker): native track shortlisting and Musical Leanings

Jaz666:
# Why

Next-track selection had been asking one model-led tool loop to do two jobs: discover eligible music and make the editorial decision. That made routine picks slower, more token-heavy, and harder to explain after the fact.

This change moves candidate discovery into the controller, where show rules, recency, blocklists, playlists, journey context and de-duplication already belong. The DJ model remains responsible for the part that benefits from judgement: choosing and explaining the best next record from a bounded eligible shortlist.

# What changes

- **Native Track Shortlisting.** The controller plans candidate sources across Context, Continuity and Exploration, then makes one structured editorial selection from the resulting shortlist.
- **A deliberate breadth control.** Three passes are the default: **Context** grounds the show or journey, **Continuity** follows the track on air, and **Exploration** reaches beyond the familiar. Two passes omit Exploration; one concentrates solely on Context. Four and five repeat Context then Continuity, widening the candidate set and final editorial prompt. This affects native selection only; Agentic Segment tools remain separate.
- **Operational clarity.** Shortlist picks retain their candidate-source provenance and a private rationale in the Booth Log, without exposing implementation machinery in listener-facing language.

Track Shortlist passes

# Musical Leanings

Musical Leanings had effectively been a hidden soft feature: a persona could carry a private musical perspective, but its limited role was not visible enough to an operator. This PR makes it an explicit, editable part of Persona Identity while preserving its deliberately narrow influence.

Musical Leanings editor

A persona may hold up to 500 characters of Musical Leanings. For the host, it is a private soft tie-breaker only after the shortlist has already satisfied the show brief, station safety, rotation, recency and musical-flow constraints. An eligible guest can contribute an even lighter nudge on a small share of picks, and only from their configured Musical Leanings—not from their Soul.

It cannot create candidates, override a show lock, bypass safety or variety rules, or enter listener-facing speech. If the artist-variety guard causes a re-pick, the same bounded shortlist rules still apply. This gives a DJ or guest a recognisable point of view without turning personality text into unbounded playlist control.

Musical Leanings host setting

Musical Leanings guest setting

# Evidence so far

**24-hour soak test**

Overnight shortlist call evidence

The overnight station window recorded 152/152 successful `djShortlistPick` calls (14.8s average; 1.06M tokens total) and 62/62 successful corrective `djShortlistRepick` calls (5.7s; 296.2k tokens). The source mix is balanced over the three intended lanes: Context 305 calls, Continuity 305 and Exploration 305.

The higher repick rate during the overnight show is expected: its strict filter limits the eligible library to 1,114 tracks. Repicks remain bounded editorial selections over already eligible alternatives; they do not rediscover music.

### Agentic vs Shortlist Comparisions
[View Document](https://github.com/Jaz666/subwave/blob/feat/track-cpu-shortlisting/docs/internals/track-shortlisting-comparisons.md)

# Verification

- `controller npm test -- shortlist-runner`
- Production web build
- Live station health check: `on-air`

# Review status

Reconciled with current `develop`; GitHub reports this branch mergeable. The live station is running the candidate successfully.


Jaz666:
I'm preparing some user facing guides to the new or changed Subwave features I've developed that you can link to in the release notes,
[this is the page for Track Shortlisting](https://github.com/Jaz666/subwave-docs/blob/main/updates/smarter-track-selection.md)

If you need a demo before merging this, send me a DM on Discord.

Jaz666:
**Important Consideration:**

This PR does not cover the listener-request agent, that remains with the Agentic Picker for the time being.
Additionally _djAgentSegment_ also remains on the existing path.

I intend on further developing to cover both these features, but for now the Agentic Picker code needs to remain in place.

EDIT 11th Sept: Segment/Skill now removed from the Agentic Path in PR#1663

Jaz666:
Follow-up integration fix (3567bcf3): native shortlisting now remains selection-only, then passes the chosen track to the verified generateLink stage. This preserves Prompt Safety and Sleeve Notes link handling and cadence. The reconciliation also restores v1.15’s lifecycle-safe boundary handoff path: captured incoming show/persona, final-track timing, and duplicate protection. The Booth’s in-progress picker status no longer exposes raw prompt wording.

Jaz666:
Follow-up (611c1487): restores the public telemetry label to `generateLink`. It also keeps cadence-selected links audible when the model ignores a short intro runway: unsafe/overlong generated wording is replaced with a compact, controller-grounded track identification and then budgeted again. Tracks with a measured vocal entry under 2.5 seconds still stay silent by design.

Jaz666:
All internal documentation cleaned up from this PR, and the two related ones #1663 and #1668

perminder-klair:
Thanks for the substantial work here. The latency and token reductions are compelling, but I don’t want native shortlisting to become the only next-track path. The model still chooses the final track, but it no longer chooses how to explore the library or reacts to thin and surprising results by trying a different source. That adaptive discovery is part of what makes the station feel like a DJ rather than a playlist engine. The paired examples and overnight run prove speed and reliability, but not equal musical judgement.

Could we keep agent-led discovery when Agentic Picker is enabled, use native shortlisting when it is off, and fall back to the shortlist or candidate pool on timeout, breaker, model failure, or budget pressure? Musical Leanings, shortlist provenance, source passes, and the diagnostic work are all worth keeping. The current branch also leaves the UI describing `pickerAgent` as controlling next-track selection even though `runTrackEvent` no longer checks it, which would leave operators with a misleading control.

Jaz666:
Hey Perminder, thanks for finding the time for looking into my 3 PRs related to offering an alternative to the Agentic path, and I'm absolutely happy for them to live behind a switch, as an alternative. All I'd ask is when it's in place, you run your station for a few days without the Agentic tools, and see if you can really tell the difference in track selection.

The whole idea is to make Subwave cheaper to run, and to offer the ability to run the DJs using a model that doesn't have to support tools.

I'll put the 3 PRs back into draft to re-work them behind a option.

I may have to split Musical Leanings out to it's own PR, so the Agentic path can benefit from it - I'm already working on my "Extended Sleeve Notes" feature which relies on it being in place.

PR #1661, opened 2026-09-11
fix(settings): let shows ride in a settings patch

uhhuhyeah:
## Why

`POST /api/settings` validates `shows` with an empty persona roster. `[].includes(id)` is false for every host and guest id, so every settings patch containing a show is rejected before `settings.update()` can assemble the effective configuration.

## What

- Use the existing `null` sentinel for an unavailable persona roster in the route’s shape-only schema pass.
- Skip host and guest membership checks only when that sentinel is present.
- Keep `settings.update()` as the strict chokepoint: after applying any same-patch persona changes, it validates shows against the resolved roster before saving.
- Add regression coverage for a show with both a host and guest persona, and regenerate the browser schema mirror.

This regression has affected whole-document settings clients since v1.7.0.

## Verification

- `cd controller && npm test -- settings-patch-schema`
- `cd controller && npm run lint` (0 errors; existing warnings only)
- `cd controller && npm test` (stops at unrelated `aio-log-link.test.ts` filesystem assertion in this environment)

### DIFF
--- a/controller/scripts/settings-patch-schema.test.ts
+++ b/controller/scripts/settings-patch-schema.test.ts
@@ -731,6 +731,23 @@ test('scheduleOverride validates SHAPE at the route, roster membership in update
   );
 });
 
+test('shows validate SHAPE at the route, roster membership in update()', () => {
+  // The route cannot know whether this patch also replaces personas. It must
+  // accept real-looking host and guest ids here; update() checks the assembled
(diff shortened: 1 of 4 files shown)


perminder-klair:
Thanks for the fix, David! Saving personas and shows together now works without weakening the final checks. I added a fuller regression test and updated the docs, and everything is green.

PR #1663, opened 2026-09-11
refactor(segments): make direct execution the only runtime

Jaz666:
# Why

Skills were taking the same agentic route as music picking: the model chose whether to call a data provider, supplied its arguments, then continued through a tool loop before it could write a short on-air segment. That was unnecessary machinery for a bounded station feature, and made the outcome less predictable for operators.

A Skill already knows what it is for. Its schedule, cooldown, eligibility, frontmatter and provider define when it can run and what information it needs. The model is still valuable for the editorial job: deciding what to say and how to say it in the station voice.

# What changes

- **Deterministic Skill runtime.** The controller selects an eligible Skill, calls its provider directly once, validates the returned data, then makes one structured generation call.
- **No agentic Segment tool loop.** Automatic, manual, cron-driven and co-hosted Skills no longer enter `djAgentSegment` or let the model choose and invoke a provider.
- **Grounded silence happens before the model.** When a data-backed Skill has no usable source material, it stands down without spending an LLM call.
- **Advanced providers remain supported.** Existing `tool.mjs` files still receive the station context, persistent state, supported services and their `SKILL.md` configuration. The former model-filled `inputs` object is now always `{}`; a visible review warning identifies Skills that declared one.

# Benefits for station operators

Every Skill firing now has a bounded, understandable cost: **zero LLM calls** when grounded data is unavailable, or **one structured LLM call** when it is ready to write. There are no exploratory tool calls or multi-step agent runs hidden behind a short segment.

That reduces token consumption, makes response time more consistent, and gives the operator a clearer answer to “why did this Skill air or remain silent?”. It also keeps the power-user extension point: code-savvy operators can still use `tool.mjs` for sources that cannot be expressed as a simple feed, without putting routine provider decisions back in the model.

# Evidence so far

The live v1.15 station has run this branch successfully for several hours through both automatic scheduling and manual firing. All installed Skills have been manually exercised, including custom providers and the restored Now-playing Dig provider.

The Debug screen shows a single `generateSegment` call for a ready Skill, with no Segment tool trail. A grounded Skill with unavailable data logs its stand-down and makes no LLM call.

# Verification

- Focused Skill runtime tests: eligibility/stand-down, `tool.mjs` loading, feeds, co-hosted Skills and discovery checks
- Controller and web lint/type-checks
- Live automatic scheduler soak plus manual Skill firing on the v1.15 station

# Scope

This is the compatibility and cost-control step from the Deterministic Segment Runtime plan. It deliberately does not redesign the Skills editor, migrate installed Skills, or introduce new Segment formats. Those changes can build on this simpler runtime in a later PR.

### DIFF
--- a/controller/scripts/cohosted-skill.test.ts
+++ b/controller/scripts/cohosted-skill.test.ts
@@ -57,14 +57,14 @@ test('a successful run maps immutable speaker ids back to the full persona objec
   const result = await runCohostedCapability({
     capability: { kind: 'case-discussion', desc: 'Discuss one case.' },
     host, guests, context: {}, situation: 'The current moment.', segmentState: {}, forced: true,
-    runAgent: async () => ({ object: full, steps: 0, toolCalls: [] }),
+    runObject: async () => full,
   });
   assert.equal(result.aired, true);
   assert.deepEqual(result.lines?.map((line: any) => line.persona), cast);
   assert.deepEqual(result.lines?.map((line: any) => line.text), full.lines.map(line => line.text));
 });
 
-test('a grounded skill must observe usable data before it may return dialogue', async () => {
+test('a grounded skill fetches usable data before it may return dialogue', async () => {
   const capability = {
     kind: 'case-discussion', desc: 'Discuss one case.', toolName: 'skill_case_discussion',
     toolDesc: 'Search for one historical case.', toolInputs: { query: 'case to find' },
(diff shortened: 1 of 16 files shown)


perminder-klair:
Thanks for the work here. I don’t think direct execution should be the only Segment runtime. The deterministic path is cheaper and more predictable, but it removes the creative and exploratory part of Skills: the model can no longer decide how to use a provider, choose useful inputs, or adapt its data gathering to the current moment. Legacy provider inputs being forced to `{}` is a concrete loss for richer custom Skills.

Could we keep agentic Segment execution when Agentic Picker is enabled, and use the new direct path when it is disabled or the agent times out, fails, hits the breaker, or is over budget? The direct provider fetch, early grounded stand-down, and compatibility work are useful fallback improvements. I only object to making direct execution the only runtime.

PR #1665, opened 2026-09-11
fix: offer timely boundary handoffs

### LINKED ISSUE #1664: [bug] Between-tracks show handoffs can sit queued for as long as the on-air track runs, actually airing several minutes after the new show started — no upper bound on the wait
Hello there!

After the latest update (v1.15.0) I've noticed that the handoffs were always late, happening several minutes after the new show started, and after the first track of the show had already aired.

I asked Claude Code to look at the issue, checking the actual codebase and recent merges, and he gave this analysis.

Hope this will be useful to solve the issue, and please let me know if you need anything else on my part.

Thanks!

–––

## Summary

Since v1.15.0 (`fix(dj): make boundary handoffs lifecycle-safe`, #1635), a show-boundary persona handoff rendered under **"Talk only between tracks"** placement waits for "the first real track seam at or after the scheduled boundary" before it airs. In practice this wait has **no effective upper bound**: it is exactly as long as whatever track happens to be on the air when the boundary lands, however long that is. Reproduced twice live, with delays of ~7:30 and ~4:00.

This isn't caused by a misconfigured `pauseTalk`/programme setting — see Environment below.

## Environment

- Subwave v1.15.0 (self-hosted, docker compose)
- `djTalkOnlyBetweenTracks: true`
- All shows involved: `programme: false`, `pauseTalk: false` (ruling out the new pause-and-talk feature and the programme outro `handover.offsetMinutes` timing as factors)
- `llm.pauseWhenEmpty: true`

## Reproduction 1 — handoff generated on time, airs 7:30 late

Station clock reaches the top of the hour while "Midnight Cruising" (DJ Keiko)'s last track ("Domino Line" by Casiopea) is still playing. The boundary-fade feature (#1574/#1583) correctly cues it out early so the transition lands close to the hour. The very next track, the first of the new show, ("The Rainbow" by Talk Talk) starts right at the boundary — a clean, on-time musical transition:

```
T+0:00   "Domino Line" cued out early (boundary fade) — clean transition to the new show
T+0:00   "The Rainbow" starts, session rolls essentially on the hour
T+0:00–7:32  "The Rainbow" runs the whole time
T+7:32   "Restless Water" starts (the second track of the show)
         → boundary handoff (sign-off + greeting) airs here
T+7:44   handoff pair complete
```

Booth log (`docker logs subwave-controller`):

```
[handoff] DJ Keiko: I'm handing over to Anna for Isotopes.
[handoff] Anna: Thanks for handing over DJ Keiko, let's get lost in Isotopes with some post-rock that prefers tape studios and jazz groups over choruses.
```

Nothing here is wrong per the current design: "The Rainbow" is genuinely the first track to start at or after the boundary. The problem is that the design puts no ceiling on how long that wait may be — it is entirely a function of whichever track happens to span the hour mark.

## Reproduction 2 — same behaviour, opposite djMode direction, cleanest possible transition

Confirmed this is not specific to one persona/show. Here the outgoing persona has `djMode: true` (does look-ahead pair-drain), the incoming one does not — opposite of Reproduction 1 — and the musical transition is about as clean as it gets:

```
13:56:45  "There's Nothing To Be Scared Of" (Tom Skinner) starts — "The Current"'s last track
13:56:54  Handoff pair GENERATED (sign-off + greeting, rendered atomically)
          → [scheduler] Holding handoff for the next track boundary
          → [scheduler] Holding handoff — the show boundary has not arrived
13:59:25  "Sketch of Melba" (Eric Dolphy) starts — 35s BEFORE the hour;
          this is already the incoming show's own opening pick
14:00:02  session rolls exactly on schedule (session.end / session.start)
          → [hourly] stood down — a handoff is rendered and waiting for the
            next track boundary
14:03:56  "Lines In My Face" (Lonnie Johnson) starts — "Sketch of Melba" finally ends
          → [handoff] Anna: Lester, here with The Prestige Sessions.
          → [handoff] Lester: Anna, thanks for passing the mic over before I
            drop this Dolphy session onto the air.
```

The handoff script was ready at 13:56:54. It aired at ~14:04 — **not** because anything failed, but because "Sketch of Melba" (the incoming show's own first track, which simply happened to start 35 seconds before the hour) had to run to completion first. The boundary-fade feature (#1574) only trims an *outgoing*-show track that would overshoot into the next show; it has no equivalent for "the incoming show's own opener already started, so the ceremonial handoff is stuck behind it."

## Root cause

No upper bound on the between-tracks wait. Per `docs/internals/show-boundary-handoffs.md`: *"With Talk only between tracks, render it during the final track but hold it for the first real track seam at or after the scheduled boundary."* The only safety net is `HANDOFF_MAX_AGE_MS` (`controller/src/broadcast/dj-agent/breaker.ts:17`, 20 minutes) — a generic "don't rot forever" fallback, not a timeliness guarantee. Both reproductions above are well inside that window and are treated as fully normal.

## Why this regressed in v1.15.0

Nothing above is new logic — it's a gap between two features developed independently and apparently never tested together:

- `fadeAtShowEnd` (#1574/#1583, merged 2026-09-06) — trims an overshooting *outgoing* track at the boundary.
- The boundary-handoff rewrite (#1635, merged 2026-09-09) — "no early handoff" fix, holds the pair for "the first real seam at or after the boundary."

`docs/internals/show-boundary-handoffs.md` and both of #1635's dedicated test files (`handoff-boundary.test.ts`, `handoff-memory-wiring.test.ts`) never mention fading/cue-out at all. The fade feature makes the *music* transition land right on the hour far more often than before — which means the handoff now almost always has to wait out a full, fresh, just-started track rather than catching a transition that was already near its end. The old "sometimes-early" bug was traded for a "frequently several minutes late" one.

## Suggested directions (not prescribing the fix)

- Give the between-tracks handoff wait an actual ceiling short of `HANDOFF_MAX_AGE_MS`, after which it falls back to ducked/immediate delivery instead of continuing to hold for an arbitrary track length.
- Alternatively, split the pair back into two moments instead of holding one atomic exchange for a full extra track: outgoing sign-off at the seam *before* the final outgoing track, incoming greeting right *after* it — landing the greeting essentially at the real show change. This is not a new idea for this codebase: it's exactly the "one closing track plays between the two voices" ordering already built and configurable (`handover.offsetMinutes`) for **programme** show handovers in #1576/#1581 — just not extended to the ordinary persona-to-persona boundary handoff this issue is about.

Jaz666:
Fixes #1664.

## What this changes

With **Talk only between tracks** enabled, a show-boundary handoff currently waits for the first real track seam at or after the scheduled show change. That can leave the outgoing sign-off and incoming greeting waiting through the entire first track of the new show.

This PR makes the delivery rule bounded for every station, with no new settings:

1. The prepared handoff pair continues to wait for a real track boundary after the scheduled show change.
2. If no seam arrives within two minutes, the same already-rendered pair is released through the normal ducked voice path over the track currently on air.

This keeps the cleaner between-tracks delivery whenever it is available, while ensuring a completed handoff does not turn into a detached greeting several minutes into the next show. It also avoids discarding a pair that has already used LLM and TTS work.

## Safety and lifecycle

The handoff remains one atomic outgoing-sign-off/incoming-greeting exchange. It keeps the existing captured outgoing and incoming identities, stale-delivery protection, listener gate, and final stream-edge acknowledgement.

The two-minute fallback does not roll the session early or permit the incoming presenter to speak before the scheduled boundary. It applies only after that boundary has passed and only while the exact pending handoff is still waiting.

## Checks

- Controller type-check
- Talk-placement regression test
- Boundary-handoff lifecycle regression test

Web changes were removed; this is now a focused queue-only fix.


trickbag:
Thanks for the quick follow-up on this!

But honestly I'm not sure this is the right approach though...

I asked Claude Code to put together a more precise and detailed answer based on my observations on the proposed fix.

–––

A couple of concerns with the approach here, before it lands:

**The default doesn't change.** `djHandoffTiming` defaults to `'next-track'`, so every station that upgrades keeps the exact behaviour reported in #1664 — an operator has to know this setting exists and go flip it to get a timely handoff. Given #1635's whole point was to make handoff timing *more* correct (no early handoff), it doesn't seem like there's a real use case for *wanting* a handoff to land several minutes into the next show by default. If `next-track`'s unbounded wait is genuinely undesirable in the general case, it reads more like a regression to fix outright than a preference to expose.

**Three modes plus a new numeric setting feels like more surface than the problem needs.** `next-track` / `on-time` / `skip`, plus `djHandoffMaxWaitMinutes` for the third — that's two new settings for what's really one question: how long should the wait be allowed to run before something else happens. A single bounded wait (short, sane default) with a fallback would cover `on-time` and the intent behind `skip` without asking operators to pick a strategy.

**`skip` reintroduces the "rendered and discarded" problem.** If nothing arrives within the wait window, it drops the whole pair outright (`dropPendingVoice` + `markHandoffAired`, no announce at all) — the sign-off and greeting were already generated (LLM + TTS paid for) and just vanish, with the boundary going by in silence. That's a worse outcome than airing them ducked over whatever's on air, which is presumably why `on-time` exists as a separate mode in the first place.

Would it be simpler to fold this into one behaviour: hold for a real seam up to a short bound (a couple of minutes, not configurable-to-10), and if none shows up, air the pair ducked immediately rather than either waiting indefinitely (today) or discarding it (`skip`)? That gets every station the fix by default, with no new settings surface and no path that throws away rendered content.

Happy to help test against the scenarios in #1664 if that's useful.

Jaz666:
Fair points, I'll defer to whatever Perminder decides on this one, as I'm getting a bit low on coding credits this week!
EDIT: OK, it's was a small change, hopefully it now aligns with your suggestions.

Jaz666:
Thanks — agreed. I have reworked the PR to remove the settings/UI and use one default rule: hold for a real post-boundary seam for two minutes, then release the same rendered atomic pair ducked over the current track. The net PR diff is now queue-only.

trickbag:
@Jaz666 thank you so much, the new PR looks great!

perminder-klair:
Thanks @Jaz666 for putting this together, and @trickbag for the thoughtful feedback. The final approach keeps handoffs natural when possible without letting them arrive several minutes late or wasting the rendered audio. Really appreciate the work on this.

PR #1668, opened 2026-09-12
refactor(requests): make direct resolution the only runtime

Jaz666:
# Why

Listener requests were the last remaining agentic path. The model could choose library/search actions through a tool loop before the controller resolved a request. That made a bounded station feature depend on tool-capable LLMs and introduced variable cost and latency.

Request handling already has a deterministic controller path: accept and persist the request, choose a locally available track, queue it, generate the listener-facing acknowledgement, expose the result to API/MCP callers, and notify webhooks. The model is needed to understand natural language, not to operate the station library.

# What changes

- **Plain JSON request matcher.** The controller makes a text-only structured request to interpret the listener's intent; no output tool or agent tool loop is required.
- **Controller-native resolution.** Track, artist, album and contextual matching remain in the existing request route and Subsonic library search path.
- **No tool fallback.** A matcher failure falls back to a direct local-library query, rather than silently reintroducing agentic tools.
- **Exact title + artist priority.** Explicit song-and-artist requests are checked for an exact local match before broader artist/catalogue selection.
- **Robust sort handling.** Text-model values such as `sort: "none"` are treated as no sort preference, while `latest`, `oldest` and `popular` retain their existing meaning.
- **Integration contract retained.** The REST request lifecycle, MCP wrapper path and webhook ordering remain unchanged.

# Benefits for station operators

Requests now work with a wider range of LLMs, including text-only models which cannot call tools. Resolution is bounded and understandable: one intent-matching generation call, then deterministic library work by the controller. The existing listener experience and Connect integrations remain in place.

# Evidence so far

The branch was tested on the live v1.15 station, built from current upstream plus the Debug Code, Track Shortlisting, and Deterministic Segment Runtime development branches.

Successful listener-facing checks included:

- chat-only request (`who are you?`) produced a suitable reply and queued no track
- direct track/artist request, older artist request (Dreadzone), genre request (Grebo), French-language request, mood request, and `more like this`
- title/artist ambiguity (`Slave to the Rhythm` by Grace Jones) initially exposed a real regression; trace logs showed the correct title and artist but a model-provided `sort: "none"` took the artist-only branch and queued `Do or Die`
- the resulting fix was deployed live and confirmed: the exact Grace Jones track then queued correctly
- requests for unavailable or contextual artists continued through the established vanilla fallback behaviour, intentionally left unchanged

Connect integration checks:

- Admin/API `POST /request` returned `202`; `GET /request/:id` returned the completed outcome
- an actual API request delivered the configured webhook event; the Webhooks admin test endpoint also returned the expected `202` after saving its configuration
- MCP shares these API endpoints; the wrapper handler and request payload lifecycle are covered by a public contract test

# Verification

- Focused request exact-match and request-limit regression tests
- REST/webhook/MCP public request-contract test
- Full controller suite: **1,676 passing tests**
- Live listener, API and webhook testing on the v1.15 station
- Merge compatibility checked in both directions with the Track Shortlisting work: this branch does not reference `djAgentPick` or its successor `djShortlistPick`, so either merge order is safe

# Scope

This removes the agentic request-resolution loop only. It deliberately does not redesign request wording, change established vanilla fallback behaviour, alter Connect endpoint shapes, or add a tool-capable fallback. Those boundaries preserve compatibility while allowing more model providers to serve requests.

### DIFF
--- a/controller/scripts/request-exact-match.test.ts
+++ b/controller/scripts/request-exact-match.test.ts
@@ -1,0 +1,35 @@
+// Exact title + artist is a listener promise, not an editorial pool. This
+// pins the narrow deterministic match that runs before broad request search.
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import { exactTitleByArtist } from '../src/music/request-match.js';
+
+const candidates = [
+  { id: 'other-grace', title: 'Private Life', artist: 'Grace Jones' },
+  { id: 'wanted', title: 'Slave to the Rhythm', artist: 'Grace Jones' },
+  { id: 'cover', title: 'Slave to the Rhythm', artist: 'A Different Artist' },
+];
+
+assert.equal(
+  exactTitleByArtist(candidates, { titles: ['Slave to the Rhythm'], artist: 'Grace Jones' })?.id,
+  'wanted',
+);
+assert.equal(
+  exactTitleByArtist(candidates, { titles: ['SLAVE TO THE RHYTHM!'], artist: 'grace-jones' })?.id,
+  'wanted',
+  'case, punctuation, and accents should not turn an exact listener request into a broad pick',
+);
+assert.equal(
+  exactTitleByArtist(candidates, { titles: ['Slave to the Rhythm'], artist: 'Unknown Artist' }),
+  null,
+  'a title-only hit must not override an explicitly named different artist',
+);
+assert.equal(exactTitleByArtist(candidates, { titles: [], artist: 'Grace Jones' }), null);
+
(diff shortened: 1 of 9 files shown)


perminder-klair:
Thanks for putting this together. After tracing a compound request such as “I want a rainy song with heavy bass,” I don’t think we should make direct resolution the only runtime. The direct matcher can keep “rainy,” but it has no energy or sound field, so it drops the “heavy bass” part. The agent can handle both through the existing mood, energy, and sound-search tools.

I’d prefer to keep agentic request handling when Agentic Picker is enabled, with the direct path as the fallback when the agent is disabled, times out, or is over budget. That preserves support for simpler models without reducing what listener requests can express.

PR #1678, opened 2026-09-14
feat(personas): Musical Leanings: explicit, auditable editorial control

Jaz666:
# Musical Leanings: explicit, bounded editorial control

## Why this exists

SUB/WAVE already includes the DJ’s Soul in the normal Agentic Picker prompt. Musical preference language written into that Soul is therefore visible to the model and may implicitly affect discovery or selection.

That is not, however, a dependable operator control. Editing the Soul also changes the DJ’s character and presentation style, and SUB/WAVE cannot determine whether its musical wording affected any particular pick.

Musical Leanings adds a separate, optional and private selection field. It provides an explicit soft editorial preference among tracks that already satisfy the current flow, show rules, rotation, requests, safety requirements and library guards. It does not create candidates, alter listener-facing speech or override existing protections.

This lets an operator express a preference such as “warm voices, melodic post-punk and overlooked discoveries” without rewriting the DJ’s wider persona.

## What the investigation established

The investigation used frozen candidate fixtures, real-library snapshots, paired control/Soul runs and A/A null calibration.

Vanilla Agentic Picking does receive the DJ’s Soul as part of its system prompt. Consequently, musical preference language in the Soul is available to the model and could influence its decisions.

The calibrated results did not, however, demonstrate a measurable Soul-specific selection effect above ordinary sampling variation:

| Route | A/A null changes | Soul A/B changes | Result |
| --- | ---: | ---: | --- |
| Track Shortlist | 9/24 | 9/24 | No detectable Soul effect above sampling variation |
| Agentic Tools | 14/24 | 12/24 | No detectable Soul effect above sampling variation |

These results do not prove that Soul can never affect selection. They show that its independent musical effect was not distinguishable from normal model variation in the evaluated model and fixtures.

SUB/WAVE therefore must not claim that:

- Soul changed a particular pick;
- vanilla Agentic Picking reliably follows musical preferences contained in Soul; or
- leaving Musical Leanings blank makes Agentic selection independent of the Soul.

The accurate statement is that Soul is present in the vanilla Agentic prompt and may have an implicit effect, but no reliable causal effect was established.

## Why Agentic uses a separate editorial pass

Early versions asked the Agentic Picker to report whether Musical Leanings had decided a close choice. Testing showed that this provenance was not dependable.

Observed failures included:

- repeating example wording from the response schema;
- relabelling ordinary flow considerations such as energy, mood or pace as Leanings evidence;
- paraphrasing the supplied preference without establishing that it affected the choice;
[...]

PR #1687, opened 2026-09-15
feat(selection): introduce Track Shortlist alongside Agentic Tools

Jaz666:
## Dependency
This PR is sequenced after #1678 and assumes its shared Musical Leanings policy. Keep both PRs independently reviewable; merge #1678 first, then refresh this branch against develop.

## Why

SUB/WAVE's retired Candidate Pool offered a simpler route for stations that did not want a model-led tool loop on every pick. This PR turns that idea into a clear, intelligent and supported choice rather than a legacy fallback.

**Track Shortlist** keeps library exploration in the Controller, where show rules, rotation, recency, playlists, requests, journey context and de-duplication already belong. The LLM still performs the editorial job it is good at: selecting the next record from a bounded eligible shortlist and giving the operator a private reason.

Stations can retain **Agentic Tools** when model-led exploration is the better fit. The two routes share the same library exploration capabilities and the same selection safeguards; they differ only in who runs the exploration and when the LLM makes its choice.

image

## What changes

- **Music Selection is a dedicated Station page.** It sits below Music Source and separates track-choice controls from the DJ's spoken behaviour.
- **Two explicit selection routes.** Agentic Tools lets the LLM explore the library and choose the final track. Track Shortlist has the Controller explore the same library sources, build an eligible shortlist, then ask the LLM for one bounded final choice.
- **Useful shortlist breadth controls.** One to five passes progress from Context through Continuity and Exploration, with the prior operator guidance restored in the UI.
- **Safe Candidate Pool upgrade.** A pre-existing `pickerAgent: false` installation with no explicit modern selection setting is migrated to Track Shortlist. Explicit current settings always take precedence.
- **Shared music policy.** Both routes honour the same no-repeat window, artist spacing, album cooldown and minimum track length.
- **Auditable Musical Leanings.** Both routes use Leanings only as an optional, private close-call tie-break. A LEANINGS diagnostic appears only when the model supplies a concrete matching tie-break; it is never listener-facing.
- **Independent optional paths.** Listener request matching and Segments & Skills can each run directly or agentically, without forcing the music-selection route.
- **Operator visibility.** The Booth and dashboard distinguish shortlist picks and corrective `djShortlistRepick` calls, preserve a valid final selection reason and source provenance, and show observed shortlist timing.
- **Maintainer comparison bench.** The hidden Discovery Bench can compare the configured Agentic and Shortlist paths without queuing music. It is disabled by default and real comparisons consume model tokens.

image

## Behaviour and compatibility

Track Shortlist is an intelligent alternative to Candidate Pool, not a removal of Agentic Tools. Existing Agentic stations continue to use Agentic Tools by default. Existing Candidate Pool stations move to Track Shortlist on upgrade, so they receive the supported controller-led route rather than silently falling back to retired behaviour.

For ready Skills, Direct execution performs a Controller fetch followed by one structured writing call. A data-backed Skill with no usable source material stands down before the LLM is called. Direct listener-request matching uses one text-only structured interpretation followed by Controller-native library resolution; existing REST, MCP and webhook contracts remain intact.

## Scope

This consolidates and supersedes [#1634](https://github.com/perminder-klair/subwave/pull/1634), [#1663](https://github.com/perminder-klair/subwave/pull/1663) and [#1668](https://github.com/perminder-klair/subwave/pull/1668) after live integration testing.

## Verification

- Controller typecheck and web typecheck
- Focused shortlist runner, latency, Segment/Skill abstention, request, artist-guard and album-guard coverage
- Candidate Pool cold-load migration test
- Shortlist Booth reason/provenance and telemetry regression checks
- Live v1.16 test-station testing, including direct and Agentic request/segment paths

## Testing notes

The Discovery Bench is available at `/admin/discovery` only when `SUBWAVE_DISCOVERY_BENCH=true` is set. It makes real model calls and is intended for maintainer comparison work, not routine production use.

## Note for Future Developments

Changes to shared selection policy—eligibility, safeguards, diagnostics, and operator-facing meaning—must be evaluated for both routes. Discovery mechanics may remain route-specific.


Jaz666:
Development documentation and internal debugging code removed.

Jaz666:
Ping @perminder-klair thanks for your patience, this feature along with Musical Leanings (which need to land first) are now considered ready for your evaluation.

Cheers,
Jamie.

